PR #2, opened 2026-04-15
feat: @amplify/ui component library + Storybook + SDUI docs

apaksh-gupta:
## Summary
- **@amplify/ui**: Shared component library with 5 core components (Badge, Button, Card, EmptyState, Skeleton) using token-based styling, TypeScript strict types, variant pattern
- **Storybook 8**: Dark-themed with Amplify branding, auto-docs, 25 stories covering all components and variants
- **SDUI Documentation**: 19 snippet types documented with JSON contracts in Storybook (SnippetDoc helper)
- Updated CI to verify UI + Storybook builds
- Chromatic visual regression workflow on UI/token changes

## Test plan
- [ ] Verify `npm run build` succeeds for all packages including UI
- [ ] Verify Storybook builds and renders all stories
- [ ] Verify SDUI snippet stories render correct JSON contracts
- [ ] Verify Chromatic CI triggers on PR changes to ui/storybook/tokens

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -43,7 +43,7 @@ jobs:
           workingDir: packages/storybook
           buildScriptName: build
           exitZeroOnChanges: false
-          exitOnceUploaded: true
+          exitOnceUploaded: false
           onlyChanged: true
           externals: |
             packages/ui/src/**
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,0 +1,50 @@
+name: Chromatic Visual Regression
+
+on:
+  pull_request:
+    paths:
+      - 'packages/ui/**'
+      - 'packages/storybook/**'
+      - 'packages/tokens-*/**'
+
+jobs:
+  chromatic:
+    name: Visual Regression
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: packages/storybook
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: .
+
+      - name: Build UI package
+        run: npm run build:ui
+        working-directory: .
+
+      - name: Build all token packages
+        run: npm run build
+        working-directory: .
+
+      - name: Run Chromatic
+        uses: chromaui/action@latest
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          workingDir: packages/storybook
+          buildScriptName: build
+          exitZeroOnChanges: false
+          exitOnceUploaded: true
+          onlyChanged: true
+          externals: |
+            packages/ui/src/**
+            packages/tokens-*/tokens/**
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,9 @@ jobs:
           fi
           echo "All packages built successfully"
 
+      - name: Build @amplify/ui for Storybook
+        run: npm run build -w packages/ui
+
       - name: Verify Storybook build
         run: npm run build -w packages/storybook
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,13 +31,12 @@ jobs:
       - name: Verify build outputs
         run: |
           MISSING=""
-          for pkg in tokens-foundation tokens-brand tokens-creator tokens-atmosphere; do
+          for pkg in tokens-foundation tokens-brand tokens-creator tokens-atmosphere ui; do
             DIR="packages/$pkg/dist"
             if [ ! -d "$DIR" ]; then
               MISSING="$MISSING\n  - packages/$pkg/dist/ (directory missing)"
               continue
             fi
-            # Check for at least one output file
             FILE_COUNT=$(ls "$DIR" 2>/dev/null | wc -l)
             if [ "$FILE_COUNT" -eq 0 ]; then
               MISSING="$MISSING\n  - packages/$pkg/dist/ (empty)"
@@ -49,6 +48,9 @@ jobs:
             exit 1
           fi
           echo "All packages built successfully"
+
+      - name: Verify Storybook build
+        run: npm run build -w packages/storybook
 
   secret-scan:
     name: Secret Scan
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -1,0 +1,37 @@
+{
+  "name": "@amplify/ui",
+  "version": "0.1.0",
+  "description": "Shared UI components for Amplify products — Brand, Creator, and Atmosphere",
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    },
+    "./styles.css": "./dist/styles.css"
+  },
+  "files": ["dist"],
+  "scripts": {
+    "build": "tsup src/index.ts --format esm --dts --external react --external react-dom",
+    "dev": "tsup src/index.ts --format esm --dts --external react --external react-dom --watch",
+    "lint": "eslint src/"
+  },
+  "peerDependencies": {
+    "react": ">=18.0.0",
+    "react-dom": ">=18.0.0"
+  },
+  "devDependencies": {
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0",
+    "tsup": "^8.0.0",
+    "typescript": "^5.5.0",
+    "clsx": "^2.1.0",
+    "tailwind-merge": "^2.3.0"
+  },
+  "dependencies": {
+    "clsx": "^2.1.0",
+    "tailwind-merge": "^2.3.0"
+  }
+}
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -26,9 +26,7 @@
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "tsup": "^8.0.0",
-    "typescript": "^5.5.0",
-    "clsx": "^2.1.0",
-    "tailwind-merge": "^2.3.0"
+    "typescript": "^5.5.0"
   },
   "dependencies": {
     "clsx": "^2.1.0",
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -9,24 +9,28 @@
     ".": {
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
-    },
-    "./styles.css": "./dist/styles.css"
+    }
   },
   "files": ["dist"],
   "scripts": {
     "build": "tsup src/index.ts --format esm --dts --external react --external react-dom",
     "dev": "tsup src/index.ts --format esm --dts --external react --external react-dom --watch",
-    "lint": "eslint src/"
+    "lint": "eslint src/",
+    "test": "vitest run"
   },
   "peerDependencies": {
     "react": ">=18.0.0",
     "react-dom": ">=18.0.0"
   },
   "devDependencies": {
+    "@testing-library/jest-dom": "^6.6.0",
+    "@testing-library/react": "^16.1.0",
+    "jsdom": "^25.0.0",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "tsup": "^8.0.0",
-    "typescript": "^5.5.0"
+    "typescript": "^5.5.0",
+    "vitest": "^2.1.0"
   },
   "dependencies": {
     "clsx": "^2.1.0",
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -47,6 +47,7 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       <button
         ref={ref}
         disabled={disabled || loading}
+        aria-busy={loading}
         className={cn(
           'inline-flex items-center justify-center font-medium transition-all duration-150',
           'focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2',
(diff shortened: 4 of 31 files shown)


packages/ui/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "peerDependencies": {
+    "react": ">=18.0.0",
+    "react-dom": ">=18.0.0"
+  },
+  "devDependencies": {
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0",
+    "tsup": "^8.0.0",
+    "typescript": "^5.5.0",
+    "clsx": "^2.1.0",
```
zenith-oi: **FLAG** — [dependency-placement] `clsx` and `tailwind-merge` appear in both `devDependencies` (L30-31) and `dependencies` (L34-35). Since these are runtime imports used inside component source files, they belong only in `dependencies`. Having them in `devDependencies` too is harmless in a monorepo (hoisted), but when consumers install `@amplify/ui` as an external package they rely on `dependencies` being resolved — the duplication will also confuse automated dependency audits. Remove the entries from `devDependencies`.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,80 @@
...
+      iconPosition = 'left',
+      children,
+      className,
+      disabled,
+      ...props
+    },
+    ref
+  ) => {
+    return (
+      <button
```
zenith-oi, line 47: **FLAG** — [missing-aria-label-on-loading-button] When `loading=true` the button text is still visually present but a spinner replaces the icon, and the button is disabled. There is no `aria-busy` attribute and no `aria-label` override to communicate the loading state to screen readers — `disabled` alone does not convey 'currently processing'. Add `aria-busy={loading}` to the `<button>` element so assistive technology announces the loading state correctly.

zenith-oi (commented):
## Zenith Code Review

**Bootstraps the `@amplify/ui` component library (Button, Badge, Card, EmptyState, Skeleton) with a Storybook package for documentation and a Chromatic workflow for visual regression. The code is well-structured and clean for a greenfield design system.**

Found 5 FLAGs. See inline comments for details.

**Positive:**
- Consistent use of `clsx` + `tailwind-merge` via the shared `cn()` utility across all components — correct approach for a Tailwind design system.
- Button uses `React.forwardRef` with `displayName` set — required for composability and correct Storybook docs generation.
- Card's interactive mode handles both `role="button"`, `tabIndex`, and `onKeyDown` — shows accessibility-first thinking even if Space key needs adding.
- Chromatic workflow uses `onlyChanged: true` and path filters so visual regression only runs when UI/token files change — avoids unnecessary Chromatic snapshot consumption.
- SDUI.stories.tsx is a smart approach: documents the server-driven UI JSON contract as living Storybook stories without needing a React Native renderer.

---

packages/ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,58 @@
...
+        variantClasses[variant],
+        paddingClasses[padding],
+        onClick && 'cursor-pointer hover:shadow-md active:scale-[0.99]',
+        className
+      )}
+      onClick={onClick}
+      aria-label={ariaLabel}
+      role={onClick ? 'button' : undefined}
+      tabIndex={onClick ? 0 : undefined}
+      onKeyDown={onClick ? (e) => e.key === 'Enter' && onClick() : undefined}
```
zenith-oi: **FLAG** — [keyboard-accessibility-card] The `onKeyDown` handler only triggers `onClick` on `Enter` (L53), but interactive elements activated via keyboard should also respond to `Space` (per WCAG 2.1 SC 2.1.1). Since `role="button"` is set, Space is expected by assistive technology. Change the condition to `(e.key === 'Enter' || e.key === ' ') && onClick()` and add `e.preventDefault()` on Space to prevent page scroll.

packages/ui/src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,34 @@
...
+    label: string;
+    onClick: () => void;
+  };
+  icon?: React.ReactNode;
+  className?: string;
+}
+
+export const EmptyState: React.FC<EmptyStateProps> = ({ title, description, action, icon, className }) => {
+  return (
+    <div className={`flex flex-col items-center justify-center py-16 px-6 text-center ${className || ''}`}>
```
zenith-oi: **FLAG** — [no-console-log-in-production] `className` is interpolated directly into the class string with `${className || ''}` rather than using the `cn()` utility already available in the package. This bypasses Tailwind Merge, meaning consumer-passed classes that conflict with base classes won't be resolved correctly (e.g. passing a custom `py-4` won't override `py-16`). Replace with `cn('flex flex-col items-center justify-center py-16 px-6 text-center', className)` — identical pattern used by every other component in this PR.

.github/workflows/ci.yml
```diff
@@ -50,6 +49,9 @@ jobs:
           fi
           echo "All packages built successfully"
 
+      - name: Verify Storybook build
+        run: npm run build -w packages/storybook
```
zenith-oi, line 56: **FLAG** — [ci-storybook-build-ordering] The 'Verify Storybook build' step (`npm run build -w packages/storybook`) runs after the token/UI build verification but there is no explicit step that builds `@amplify/ui` first in this job. The Chromatic workflow has a dedicated `build:ui` step before running Chromatic, but the CI job that verifies the Storybook build does not. If `packages/ui/dist` is not present (e.g. on a clean runner), the Storybook build will fail with a missing module. Add `npm run build:ui` before the Storybook build step in `ci.yml`.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,81 @@
...
+    'border border-border bg-transparent text-neutral-900 hover:bg-surface-overlay focus-visible:ring-neutral-200',
+};
+
+const sizeClasses: Record<ButtonSize, string> = {
+  sm: 'h-8 px-3 text-sm gap-1.5 rounded-md',
+  md: 'h-10 px-4 text-base gap-2 rounded-md',
+  lg: 'h-12 px-6 text-base gap-2 rounded-lg',
+};
+
+export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
```
zenith-oi, line 31: **FLAG** — [missing-tests] Zero unit tests exist for any component in `@amplify/ui`. Storybook stories are documentation, not tests — they don't assert behaviour or catch regressions in CI without Storybook test runner or Chromatic interaction tests. The Button's loading state, disabled+loading combination, keyboard handler on Card, and icon rendering paths are all untested. Add at minimum a `Button.test.tsx` using Vitest + Testing Library covering: default render, loading disables button and shows spinner, disabled prop, icon left/right positions, and forwarded ref. Same for Card's `onKeyDown` handler (Enter/Space triggers onClick) which is the most likely source of future breakage.

zenith-oi (commented):
## Zenith Code Review

**This PR bootstraps the Amplify Design System with a `@amplify/ui` component package (Button, Badge, Card, EmptyState, Skeleton), a Storybook documentation site, SDUI snippet documentation stories, and Chromatic visual regression CI. The code is well-structured and production-ready for a v0.1 design system.**

Found 4 FLAGs. See inline comments for details.

**Positive:**
- The `cn` utility (clsx + tailwind-merge) is the right pattern for composable Tailwind classes — prevents class conflict bugs when consumers pass `className` overrides.
- Button correctly uses `React.forwardRef` and sets `displayName`, making it compatible with form libraries and debugging tooling.
- Card's keyboard handler (Enter/Space → onClick) is thoughtful and correct for custom interactive elements — most design systems miss this.
- Chromatic workflow correctly uses `fetch-depth: 0` for accurate baseline comparison and scopes triggers to only `packages/ui` and `packages/storybook` paths.
- SDUI stories as JSON contract documentation is an excellent idea — it makes the server-driven UI schema self-documenting and reviewable without needing a native renderer.

---

packages/ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,63 @@
...
+      className={cn(
+        'transition-all duration-150',
+        variantClasses[variant],
+        paddingClasses[padding],
+        onClick && 'cursor-pointer hover:shadow-md active:scale-[0.99]',
+        className
+      )}
+      onClick={onClick}
+      aria-label={ariaLabel}
+      role={onClick ? 'button' : undefined}
```
zenith-oi, line 60: **FLAG** — [accessibility-contract] A `div`/`article`/`section`/`li` with `role='button'` and `tabIndex={0}` is a custom interactive element. Screen readers announce it as a button, so it must have an accessible name. The `aria-label` prop is optional and callers frequently won't pass it when `children` provides visual context — but `role='button'` elements need an explicit label or `aria-labelledby`. Additionally, when `as='button'` is not used, the element won't receive native button behaviour (e.g. form submission, touch targets). Consider requiring `aria-label` when `onClick` is provided (TypeScript conditional type), or document clearly that callers must supply it. The a11y addon in Storybook will catch this at story time but not enforce it for library consumers.

packages/ui/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "description": "Shared UI components for Amplify products — Brand, Creator, and Atmosphere",
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    },
+    "./styles.css": "./dist/styles.css"
```
zenith-oi: **FLAG** — [missing-css-bundle] The package exports `./styles.css` at `./dist/styles.css`, but the build script (`tsup src/index.ts --format esm --dts`) does not emit any CSS. tsup does not process Tailwind CSS by default — it needs a PostCSS/Tailwind pipeline (e.g. `--inject-style` or a separate `tailwindcss` CLI step) to generate the stylesheet. Any consumer who does `import '@amplify/ui/styles.css'` will get a module-not-found error at runtime. Either add a CSS build step to `package.json` scripts or remove the export until it is implemented.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,50 @@
...
+        working-directory: .
+
+      - name: Run Chromatic
+        uses: chromaui/action@latest
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          workingDir: packages/storybook
+          buildScriptName: build
+          exitZeroOnChanges: false
+          exitOnceUploaded: true
```
zenith-oi: **FLAG** — [chromatic-exitOnceUploaded] `exitOnceUploaded: true` causes the Chromatic action to exit as soon as the build is uploaded, before any visual comparisons run. This means the CI step will always pass even if there are visual regressions — defeating the purpose of the workflow. Set `exitOnceUploaded: false` (or remove it, as false is the default) so the action waits for comparison results and fails the PR when visual changes are detected.

zenith-oi (commented):
## Zenith Code Review

**This PR bootstraps the `@amplify/ui` component library (Button, Badge, Card, EmptyState, Skeleton) and a Storybook package for documentation and visual regression testing via Chromatic. The code is well-structured with good accessibility handling and CI integration. A few issues worth addressing before merge.**


### FLAG
- **[missing-aria-label-enforcement]** packages/ui/src/components/Card/Card.tsx:60 — When `onClick` is provided, the Card renders with `role="button"` but `aria-label` is typed as optional (`aria-label?: string`). If the card's text content is not descriptive (e.g. it contains only an image or icon), screen readers will have no accessible name. The comment in the JSDoc acknowledges this but doesn't enforce it. Consider requiring `aria-label` when `onClick` is present via a conditional type or a runtime warning in dev: `if (process.env.NODE_ENV !== 'production' && onClick && !ariaLabel) console.warn('[Card] aria-label is required when onClick is provided')`. This is a real accessibility gap that won't be caught at compile time.
- **[no-console-log-in-production]** packages/ui/src/components/Card/Card.tsx:60 — Related to above — if a dev warning is added (recommended), use a proper logger or guard it with `process.env.NODE_ENV !== 'production'`. Per Amplify coding standards, `console.log` must not appear in production code paths. This is a preemptive note for when that warning is added.
- **[missing-tests-critical-paths]** packages/ui/src/components:1 — Tests exist only for Button. Badge, Card, EmptyState, and Skeleton have no test coverage. Card's interactive mode (keyboard navigation via Enter/Space, role=button, tabIndex) is a non-trivial accessibility contract that's particularly risky without tests. At minimum, Card's click/keyboard handling and EmptyState's conditional rendering paths should be covered. The Zenith standard is explicit: every feature needs unit tests.
- **[chromatic-exitZeroOnChanges]** .github/workflows/chromatic.yml:45 — `exitZeroOnChanges: false` means any visual change detected by Chromatic will fail the CI job and block the PR — including intentional visual changes in the same PR that introduced the components. For a greenfield design system where visual iteration is frequent, this will create constant friction. The conventional setup for a new design system is `exitZeroOnChanges: true` (CI passes, Chromatic review is async in the Chromatic UI). Switch to `false` once the baseline is stable and the team has a visual review workflow established.
- **[missing-tailwind-config]** packages/ui/package.json:1 — The UI package's `build` script (`tsup src/index.ts --format esm --dts`) does not process Tailwind CSS. The components use Tailwind utility classes (e.g. `bg-brand`, `text-neutral-900`, `animate-pulse`) that are custom tokens — consumers of `@amplify/ui` will need to include Tailwind with the design system's token config in their own Tailwind setup, or the classes will silently produce no styles. There is no documentation in the package about this requirement, and no `tailwind.config` in `packages/ui`. This needs either: (a) a CSS bundle output (process Tailwind in the build step), or (b) explicit documentation in a README that consumers must configure Tailwind to include `@amplify/ui` paths and the token preset.

**Positive:**
- The Card component's keyboard accessibility implementation is correct and complete — Enter/Space key handling, tabIndex, and role=button are all wired up properly with e.preventDefault() to prevent scroll on Space.
- The SDUI.stories.tsx approach of documenting the JSON contract alongside the type name and description is genuinely useful for mobile engineers consuming the SDUI system — this is a creative and practical use of Storybook.
- Chromatic workflow correctly uses `fetch-depth: 0` for full git history (required for Chromatic's TurboSnap diff detection) and `onlyChanged: true` to avoid re-testing unaffected stories.
- The `cn()` utility using clsx + tailwind-merge is the correct pattern for a Tailwind component library — prevents class conflicts when consumers pass `className` overrides.
- Button correctly sets both `disabled` and `aria-busy` in loading state, and hides the spinner from screen readers with `aria-hidden="true"`.

---

PR #5, opened 2026-04-18
feat: token drift detection script (Phase 6 CI)

apaksh-gupta:
## Summary
- Adds `scripts/check-token-drift.js` — zero-dependency Node.js script that scans codebases for hardcoded hex colors not in the design system token allowlist
- Supports `--generate-baseline` mode for grandfathering existing violations
- Outputs GitHub Actions annotations for inline PR feedback

## Context
Phase 6 of the Design System Unification plan. This script is consumed by token-drift CI workflows in one-dashboard-web, one_club_app, and odin-agent.

## Test plan
- [ ] Run `node scripts/check-token-drift.js --help` locally
- [ ] Run against a product repo to verify detection works
- [ ] Verify baseline generation mode captures existing violations

### DIFF
--- a/scripts/check-token-drift.js
+++ b/scripts/check-token-drift.js
@@ -1,0 +1,390 @@
+#!/usr/bin/env node
+
+// Token Drift Checker -- Phase 6 of Design System Unification
+//
+// Scans source files for hardcoded hex colors and reports any that are not
+// in the design system token allowlist. Designed to run in CI on PRs.
+//
+// Usage:
+//   node check-token-drift.js --scan <glob> --exclude <dir> --tokens <path>
+//       [--baseline <file>] [--generate-baseline]
+//
+// Arguments:
+//   --scan <glob>          Glob pattern for files to scan (repeat for multiple)
+//   --exclude <dir>        Directory prefix to exclude (repeat for multiple)
+//   --tokens <path>        Path to amplify-design-system root
+//   --baseline <file>      Path to baseline JSON (grandfathered violations)
+//   --generate-baseline    Generate baseline from current state and exit
+//
+// Modes:
+//   Default:             Compare found hex colors against tokens + baseline. Exit 1 if new violations.
+//   --generate-baseline: Output a baseline JSON of all currently-used hex colors.
+//
+// Zero dependencies -- uses only Node.js built-ins (fs, path).
+
+const fs = require('fs');
+const path = require('path');
+
+// ---------------------------------------------------------------------------
+// Argument parsing
+// ---------------------------------------------------------------------------
+
+const args = process.argv.slice(2);
+const scanPatterns = [];
+const excludePatterns = [];
+let tokensDir = null;
+let baselinePath = null;
+let generateBaseline = false;
+
+for (let i = 0; i < args.length; i++) {
+  switch (args[i]) {
+    case '--scan':
+      scanPatterns.push(args[++i]);
+      break;
+    case '--exclude':
+      excludePatterns.push(args[++i]);
+      break;
+    case '--tokens':
+      tokensDir = args[++i];
+      break;
+    case '--baseline':
+      baselinePath = args[++i];
+      break;
+    case '--generate-baseline':
+      generateBaseline = true;
+      break;
+    default:
+      console.error(`Unknown argument: ${args[i]}`);
+      process.exit(2);
+  }
+}
+
+if (scanPatterns.length === 0) {
+  console.error('Error: at least one --scan pattern is required');
+  process.exit(2);
+}
+
+// ---------------------------------------------------------------------------
+// File matching (simple, zero-dep)
+// ---------------------------------------------------------------------------
+
+// Parse a glob pattern like "src/**/*.scss" into { prefix, extensions }
+// Supports: dir/**/*.ext and dir/**/*.{ext1,ext2}
+function parsePattern(pattern) {
+  const normalized = pattern.replace(/\\/g, '/');
+
+  // Extract the directory prefix (everything before the first *)
+  const firstStar = normalized.indexOf('*');
+  const prefix = firstStar >= 0 ? normalized.substring(0, firstStar) : normalized;
+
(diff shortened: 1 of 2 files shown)


scripts/check-token-drift.js
```diff
@@ -0,0 +1,390 @@
...
+    const uniqueColors = [...new Set(violations.map(v => v.color))];
+    console.log(`::warning file=${file}::${uniqueColors.length} grandfathered hex colors: ${uniqueColors.join(', ')}`);
+  }
+}
+
+// New violations — errors
+if (newViolations.length > 0) {
+  console.log(`\n--- NEW violations (${newViolations.length}) ---`);
+  for (const v of newViolations) {
+    console.log(`::error file=${v.file},line=${v.line},col=${v.column}::Hardcoded hex color ${v.color} — use a design token instead. Context: ${v.context}`);
```
zenith-oi, line 383: **FLAG** — [no-console.log-in-production] The `context` field in the error output is derived directly from raw file content (line.trim().substring(0, 120)) and emitted via console.log to CI output. If any scanned file contains secrets, tokens, or credentials embedded in a line that also happens to contain a hex color (e.g. a hardcoded API key in a config file that also sets a color), those secrets will be surfaced in CI logs. Consider redacting the context field, or at least only including it in non-CI output modes. This is a low-probability but non-zero risk depending on what the scan glob covers.

scripts/check-token-drift.js
```diff
@@ -0,0 +1,390 @@
...
+  for (const pkg of packages) {
+    const tokDir = path.join(packagesDir, pkg, 'tokens');
+    if (!fs.existsSync(tokDir)) continue;
+    for (const file of fs.readdirSync(tokDir).filter(f => f.endsWith('.json'))) {
+      try {
+        const content = JSON.parse(fs.readFileSync(path.join(tokDir, file), 'utf8'));
+        for (const c of extractAllTokenColors(content)) {
+          colors.add(c);
+        }
+      } catch {
```
zenith-oi, line 196: **FLAG** — [silent-catch] The catch block at L196 silently skips malformed JSON token files with no output at all. A corrupted or accidentally empty token file would silently reduce the allowlist, potentially causing false positives (valid token colors flagged as violations) and breaking CI in a confusing way. At minimum, emit a warning: `console.warn('Warning: skipping malformed token file:', file, e.message)`.

scripts/check-token-drift.js
```diff
@@ -0,0 +1,390 @@
...
+        // skip malformed JSON
+      }
+    }
+  }
+
+  // Also add universal safe colors
+  const universals = [
+    '#000', '#000000', '#fff', '#ffffff', '#fff0', '#0000', '#00000000',
```
zenith-oi, line 204: **FLAG** — [correctness-hex-validation] `'#transparent'` is in the universals allowlist (L205) but `#transparent` is not a valid CSS hex color — it starts with `#` but contains non-hex characters, so `HEX_REGEX` at L217 (`#[0-9a-fA-F]{3,8}`) would never match it in source files. This entry is harmless but misleading; it will never be used and implies the script handles named colors, which it does not. Remove it to avoid confusion.

scripts/check-token-drift.js
```diff
@@ -0,0 +1,390 @@
...
+  if (tokenColors.has(finding.color)) {
+    tokenMatches.push(finding);
+    continue;
+  }
+
+  const key = `${finding.file}:${finding.color}`;
+  const fileKey = finding.file;
+
+  // Check if this specific file+color combo is in baseline
+  if (baseline[fileKey] && baseline[fileKey].includes(finding.color)) {
```
zenith-oi, line 316: **FLAG** — [correctness-baseline-shrinkage] The baseline check only prevents a file+color combo from being flagged as *new*, but there is no mechanism to detect when baseline entries are resolved (i.e., a previously grandfathered color is removed from a file). This means the baseline can only grow, never shrink — teams get no signal that they have successfully migrated a file off hardcoded colors. Consider adding a post-scan step that warns when baseline entries no longer appear in the scan results, so the baseline can be trimmed over time. Without this, the drift-reduction goal of Phase 6 has no enforcement path.

scripts/check-token-drift.js
```diff
@@ -0,0 +1,390 @@
...
+  const findings = scanFile(fullPath);
+  for (const f of findings) {
+    f.file = relPath;  // Use relative path for readability
+  }
+  allFindings.push(...findings);
+}
+
+// 3. Load token allowlist
+let tokenColors = new Set();
+if (tokensDir) {
```
zenith-oi, line 287: **FLAG** — [missing-tokens-dir-warning] If `--tokens` is not provided, `tokenColors` remains an empty Set and every hex color found will be classified as a violation (unless baselined). This is likely to produce massive false-positive noise if someone runs the script without `--tokens`. There is no warning emitted about running without a token allowlist. At L287, add: `if (!tokensDir) console.warn('Warning: --tokens not provided; all hex colors will be treated as violations.')` so the behavior is not silently surprising.

zenith-oi (commented):
## Zenith Code Review

**Adds a zero-dependency Node.js CI script that scans source files for hardcoded hex colors and compares them against a design token allowlist and a grandfathered baseline. The script is well-structured and thoughtfully built for its purpose.**

Found 5 FLAGs. See inline comments for details.

**Positive:**
- Zero-dependency implementation using only Node.js built-ins is exactly right for a CI utility — no lockfile churn, no supply-chain risk, runs anywhere Node is available.
- The baseline/grandfathering mechanism is well-designed: it allows incremental adoption without blocking existing work, which is the correct pattern for a Phase 6 brownfield migration.
- GitHub Actions annotation format (`::error file=...` and `::warning file=...`) is used correctly, so violations will appear as inline PR annotations — a good CI UX decision.
- The `generate-baseline` mode outputting a sorted, stable JSON is thoughtful — it makes baseline diffs in PRs readable and avoids spurious churn.

---

PR #8, opened 2026-04-20
feat: Phase 1 — W3C token architecture with Canvas design system

apaksh-gupta:
## Summary
- Complete token restructure: flat files → 3-layer hierarchy (primitives → semantic → product)
- W3C DTCG format (`$value`, `$type`, `$description`)
- Approved Canvas design: Rich Violet #7C3AED, Warm Charcoal #1C1917, 14px body, 16px radius
- Unified violet across ALL products (Atmosphere gold removed)
- 6 theme files (3 products × light/dark)
- Shared `build-tokens.js` generates 5-6 platform outputs per package
- Removed Figma integration (tokens.config.json + figma-sync.yml)

## What changed
- 9 primitive token files (colors, spacing, radii, shadows, typography, breakpoints, z-index, **motion**, **opacity**)
- 2 semantic token files (light + dark color mapping)
- 6 product theme files with resolved references
- New shared build script generating CSS, SCSS, JSON, JS, Tailwind v4, React Native
- All 21 dist artifacts verified building correctly

## Test plan
- [x] `node scripts/build-tokens.js foundation` — builds 5 artifacts
- [x] `node scripts/build-tokens.js brand` — builds 5 artifacts
- [x] `node scripts/build-tokens.js atmosphere` — builds 5 artifacts
- [x] `node scripts/build-tokens.js creator` — builds 6 artifacts (+ React Native)
- [ ] CI passes (Build & Validate, Secret Scan, SDUI Sync Check)


scripts/build-tokens.js
```diff
@@ -0,0 +1,268 @@
...
+ */
+
+import { readFileSync, writeFileSync, mkdirSync, existsSync, readdirSync, statSync } from 'fs';
+import { join, dirname } from 'path';
+import { fileURLToPath } from 'url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const ROOT = join(__dirname, '..');
+
+const pkg = process.argv[2];
```
zenith-oi, line 27: **BLOCK** — [shell-injection] The `pkg` variable comes directly from `process.argv[2]` and is interpolated into a shell command via `execSync` in the per-package build.js delegates (e.g. `execSync(`node ${join(root, 'scripts/build-tokens.js')} brand`)`). While the delegates hardcode the package name so the immediate call is safe, the main script itself accepts arbitrary user input at L26 and uses it to construct filesystem paths at L114-L135 (`join(ROOT, 'packages/tokens-${pkg}/tokens')`, `join(ROOT, 'packages/tokens-${pkg}/dist')`). A path like `../../etc` or `../tokens-foundation/../../../some-other-path` could read/write outside the repo. Add an allowlist check immediately after L26-L30: `const VALID_PACKAGES = new Set(['foundation', 'brand', 'atmosphere', 'creator']); if (!VALID_PACKAGES.has(pkg)) { console.error(...); process.exit(1); }` This is also the right guard if this script is ever called from CI with a dynamic input.

packages/tokens-foundation/tokens/primitives/z-index.json
```diff
@@ -0,0 +1,14 @@
+{
+  "z": {
```
zenith-oi: **BLOCK** — [breaking-token-rename] The old token key was `z-index` (e.g. `--amp-z-index-modal: 1050`); the new key is `z` (e.g. `--amp-z-modal: 1050`). Similarly, the old `tokens/radii.json` used key `radius` with a `2xl` value of `24px`; the new file has `2xl` as `20px` — a silent value change. These are breaking changes to the published CSS variable names and values that all three product repos depend on. The `z-index` → `z` rename will silently break any hardcoded `var(--amp-z-index-*)` references in `one-dashboard-web`, `odin-agent`, and `one_club_app`. The `radius-2xl` value change (24px → 20px) will cause visual regressions. Before merging, confirm that (a) Pixel's drift detection will catch this, (b) dependent repos have been updated or a migration guide is in place, and (c) the value change for `radius-2xl` is intentional.

scripts/build-tokens.js
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// ── Load tokens ──
+const foundationDir = join(ROOT, 'packages/tokens-foundation/tokens');
+const packageDir = join(ROOT, `packages/tokens-${pkg}/tokens`);
+
+let allTokens = {};
+
+// Foundation primitives + semantics always loaded
+allTokens = loadJsonFiles(foundationDir);
```
zenith-oi: **FLAG** — [theme-merging-correctness] The build script loads ALL JSON files recursively from `foundationDir`, which includes both `semantic/colors-light.json` and `semantic/colors-dark.json`. These two files define the same `semantic.*` keys with different values. `deepMerge` will apply them in filesystem readdir order (non-deterministic across OSes/Node versions), meaning the resolved `semantic.accent` could be either the light or dark value depending on which file is processed last. The CSS output at L139-L149 uses this merged result, so the generated `variables.css` will have randomly light or dark semantic tokens. The theme files need to be separated into distinct build passes (e.g. a `:root` block for light and a `[data-theme='dark']` block for dark), not merged together. This affects all packages since foundation semantics are always loaded.

scripts/build-tokens.js
```diff
@@ -0,0 +1,268 @@
...
+// ── Write outputs ──
+writeFileSync(join(distDir, 'variables.css'), buildCSS());
+writeFileSync(join(distDir, 'variables.scss'), buildSCSS());
+writeFileSync(join(distDir, 'tokens.json'), buildJSON());
+writeFileSync(join(distDir, 'tokens.js'), buildJS());
+writeFileSync(join(distDir, 'tailwind.css'), buildTailwindPreset());
+
+if (pkg === 'creator') {
+  writeFileSync(join(distDir, 'tokens.native.js'), buildReactNative());
+  console.log(`@amplify/tokens-${pkg}: built 6 artifacts (+ React Native)`);
```
zenith-oi, line 288: **FLAG** — [no-console-log-in-production] Minor: `console.log` is used for build output which is fine for a CLI build script. Not a production service concern — ignore this.

packages/tokens-creator/tokens/theme-light.json
```diff
@@ -0,0 +1,44 @@
...
+      "$type": "color",
+      "neutral-strong":  { "$value": "{color.stone.900}" },
+      "neutral-medium":  { "$value": "{color.stone.500}" },
+      "neutral-weak":    { "$value": "{color.stone.300}" },
+      "neutral-subtle":  { "$value": "{color.stone.200}" },
+      "neutral-inverse": { "$value": "{color.white}" },
+      "primary":         { "$value": "{color.violet.600}" },
+      "primary-weak":    { "$value": "{color.violet.100}" },
+      "positive":        { "$value": "{color.green.600}" },
+      "positive-weak":   { "$value": "{color.green.50}" },
```
zenith-oi: **FLAG** — [sdui-backward-compat] The old `tokens/colors.json` had `positive-weak: #E3F6EC` (a solid color). The new token maps this to `{color.green.50}` which resolves to `#ECFDF5` — a different hex value. Similarly `notice-weak` was `#FFF3E1`, now `{color.amber.50}` = `#FFFBEB`. The CLAUDE.md notes a `sdui-sync-check.yml` CI workflow that validates creator tokens against `api-gateway ColorType enum`. If that workflow still runs, it may catch this. But these are visual regressions for the Creator mobile app that should be explicitly called out in the PR description and verified against the SDUI spec.

zenith-oi (dismissed):
## Zenith Code Review

**Major refactor of the token build pipeline: replaces per-package Style Dictionary configs with a single shared build script, migrates all token files from legacy `value`/`type` format to W3C DTCG `$value`/`$type` format, restructures foundation tokens into primitives/semantic layers, adds light/dark theme token files, and removes the Figma sync workflow. The architecture is sound and the implementation is clean.**

Found 2 BLOCKs, 4 FLAGs. See inline comments for details.

**Positive:**
- Eliminating four near-identical Style Dictionary configs in favour of a single shared build script is the right call — the old per-package scripts had diverged (brand used CJS require(), creator/foundation/atmosphere used ESM import), and the new approach enforces consistency.
- Migrating to W3C DTCG format ($value/$type) is the correct direction for long-term tooling compatibility and the implementation of resolveValue() handles recursive reference chains correctly.
- The primitives/semantic split in tokens-foundation is well-structured — having stable primitive values that semantic tokens reference via {color.violet.600} makes theme changes surgical rather than scattered.
- The deepMerge correctly treats nodes with $value as leaf tokens (L52), preventing it from descending into token objects and corrupting their structure.

---

zenith-oi (commented):
## Zenith Code Review

**This PR migrates all token packages from per-package Style Dictionary configs to a single shared build script (scripts/build-tokens.js), while simultaneously restructuring the token files from legacy Style Dictionary v3 format to W3C DTCG format ($value/$type). The consolidation is architecturally sound, but there are two real issues worth fixing before merge.**

Found 4 FLAGs. See inline comments for details.

**Positive:**
- Excellent DRY improvement — replacing four divergent per-package build scripts (including the brand package's 219-line hand-rolled Style Dictionary reimplementation) with a single 268-line shared script is exactly the right call.
- The W3C DTCG token format migration ($value/$type) is well-executed and consistent across all new files.
- The primitive/semantic/theme layering in the new token structure is architecturally clean — primitives define raw values, semantic maps them to intent, and product themes specialize on top.
- Adding `"type": "module"` to the root package.json correctly unifies the ESM approach across all build scripts instead of mixing CJS require() and ESM import syntax.

---

scripts/build-tokens.js
```diff
@@ -0,0 +1,268 @@
+#!/usr/bin/env node
+/**
+ * Shared token build script for all @amplify/tokens-* packages.
+ *
+ * Reads W3C DTCG-format token JSON files, resolves {references},
+ * and generates platform-specific outputs:
+ *   - CSS custom properties (variables.css)
```
zenith-oi, line 7: **FLAG** — [shell-injection] The `pkg` variable from `process.argv[2]` is interpolated directly into a shell command string via execSync in all four package build.js files (e.g. `execSync('node ... atmosphere')`). In the package build.js files the argument is hardcoded so this is low-risk there, but `build-tokens.js` itself accepts arbitrary CLI input at L26 (`const pkg = process.argv[2]`) and that same value is used to construct file paths and output labels throughout. If anyone ever calls `node scripts/build-tokens.js` directly with a crafted argument containing path traversal (e.g. `../../etc/passwd`), the `join(ROOT, 'packages/tokens-${pkg}/tokens')` at L114 and distDir at L135 will happily read/write outside the packages directory. Add an allowlist check: `const VALID_PKGS = ['foundation', 'brand', 'atmosphere', 'creator']; if (!VALID_PKGS.includes(pkg)) { console.error(...); process.exit(1); }` immediately after L30.

packages/tokens-foundation/tokens/primitives/colors.json
```diff
@@ -0,0 +1,65 @@
+{
```
zenith-oi, line 1: **FLAG** — [broken-contract-pixel-integration] CLAUDE.md explicitly states that Pixel's token-sync.ts reads from `packages/tokens-foundation/tokens/*.json` (flat), `packages/tokens-brand/tokens/*.json`, etc. This PR restructures all files: foundation tokens are now under `tokens/primitives/` and `tokens/semantic/` subdirectories, and product tokens are now `theme-light.json`/`theme-dark.json` instead of `colors.json`/`typography.json`. The token schema also changed from `{"value": ..., "type": ...}` to `{"$value": ..., "$type": ...}`. Pixel's `token-sync.ts` will silently read zero tokens or misparse everything it finds — drift detection becomes blind the moment this merges. This needs a coordinated update to Pixel's file path patterns and DTCG $value parsing before or alongside this PR. At minimum, confirm with the Pixel team that token-sync.ts handles the new paths and W3C format, and document the Pixel change requirement in the PR description.

scripts/build-tokens.js
```diff
@@ -0,0 +1,268 @@
...
+  const tokens = {};
+  if (!existsSync(dir)) return tokens;
+
+  for (const entry of readdirSync(dir)) {
+    const full = join(dir, entry);
+    const stat = statSync(full);
+    if (stat.isDirectory()) {
+      Object.assign(tokens, loadJsonFiles(full));
+    } else if (entry.endsWith('.json')) {
+      const data = JSON.parse(readFileSync(full, 'utf8'));
```
zenith-oi, line 44: **FLAG** — [silent-catch-missing-error-handling] `JSON.parse(readFileSync(full, 'utf8'))` at L43 inside `loadJsonFiles` has no error handling. A malformed JSON file (easy to introduce when designers hand-edit tokens) will throw an uncaught SyntaxError that surfaces as a cryptic stack trace with no indication of which file caused the problem. Wrap in try/catch and re-throw with the filename: `try { ... } catch (e) { throw new Error('Failed to parse ' + full + ': ' + e.message); }`

scripts/build-tokens.js
```diff
@@ -0,0 +1,268 @@
...
+// ── Load all JSON token files recursively ──
+function loadJsonFiles(dir) {
+  const tokens = {};
+  if (!existsSync(dir)) return tokens;
+
+  for (const entry of readdirSync(dir)) {
+    const full = join(dir, entry);
+    const stat = statSync(full);
+    if (stat.isDirectory()) {
+      Object.assign(tokens, loadJsonFiles(full));
```
zenith-oi, line 42: **FLAG** — [token-collision-silent-overwrite] `loadJsonFiles` uses `deepMerge` with `Object.assign(tokens, loadJsonFiles(full))` for directories and `deepMerge(tokens, data)` for files. When two JSON files in the same directory define the same top-level key (e.g. both `semantic/colors-light.json` and `semantic/colors-dark.json` define a `semantic` root key), the second file silently wins for any overlapping leaf tokens. This is currently load-order dependent (filesystem readdir ordering). The two semantic files happen to define the same key names with different values — both light and dark `semantic.accent` will be loaded into `allTokens`, with whichever file sorts last in readdir winning. This means the CSS output will arbitrarily contain either light or dark values depending on OS/filesystem. The architecture needs to separate light/dark into distinct resolved passes, or merge them under distinct namespaces (`semantic-light` / `semantic-dark`). This is a correctness bug that will produce inconsistent builds across environments.

packages/tokens-atmosphere/build.js
```diff
@@ -1,32 +1,7 @@
...
-      buildPath: 'dist/',
-      files: [{ destination: 'tokens.js', format: 'javascript/es6' }],
-    },
-  },
-});
-
-await sd.buildAllPlatforms();
-console.log('tokens-atmosphere: build complete');
+const root = join(dirname(fileURLToPath(import.meta.url)), '../..');
+execSync(`node ${join(root, 'scripts/build-tokens.js')} atmosphere`, { stdio: 'inherit' });
```
zenith-oi, line 7: **BLOCK** — [shell-injection] The package name is interpolated directly into a shell command string: `execSync(`node ${join(root, 'scripts/build-tokens.js')} atmosphere`, ...)`. While the package names are hardcoded literals here (atmosphere/brand/creator/foundation) and VALID_PACKAGES validation exists in build-tokens.js, the pattern itself is unsafe — anyone who adds a new package and follows this pattern with a user-supplied name creates a shell injection vector. More concretely, `join()` on a path containing spaces or special characters on Windows could break the command. Use `execFileSync` with an args array instead: `execFileSync(process.execPath, [join(root, 'scripts/build-tokens.js'), 'atmosphere'], { stdio: 'inherit' })`. This also removes the implicit dependency on the system shell.

packages/tokens-foundation/tokens/primitives/colors.json
zenith-oi, line 1: **FLAG** — [breaking-contract / pixel-integration] CLAUDE.md explicitly states that Pixel Agent (pixel-agent repo) reads canonical token files from this repo via GitHub API — specifically `packages/tokens-foundation/tokens/*.json` and the per-product token paths. This PR completely replaces the token schema: old keys like `color.primitive.red-500`, `color.semantic.error`, `color.semantic.brand`, and the old flat `tokens/colors.json` paths are gone. The new primitives are under `tokens/primitives/colors.json` with entirely different key paths (`color.violet.600`, `color.stone.900`, etc.) and the semantic layer has moved to `tokens/semantic/colors-light.json`. Pixel's `token-sync.ts` will start reading empty or mismatched data from the old paths. This is a cross-service breaking change that needs explicit coordination with the Pixel Agent repo before or alongside merge — the path patterns Pixel uses to read these files must be updated atomically.

scripts/build-tokens.js
```diff
@@ -0,0 +1,287 @@
...
+    } else if (entry.endsWith('.json')) {
+      const data = JSON.parse(readFileSync(full, 'utf8'));
```
zenith-oi, line 44: **FLAG** — [missing-error-handling] `JSON.parse(readFileSync(full, 'utf8'))` has no error handling. A malformed JSON token file will throw an uncaught SyntaxError with no context about which file caused it. Wrap in a try/catch that re-throws with the file path: `try { ... } catch (e) { throw new Error(`Failed to parse token file: ${full}\n${e.message}`); }`. This matters because Figma/Tokens Studio syncs write these files automatically and a partial write could produce invalid JSON.

scripts/build-tokens.js
```diff
@@ -0,0 +1,287 @@
...
+  const match = val.match(/^\{(.+)\}$/);
+  if (!match) return val;
+
+  const path = match[1].split('.');
+  let current = root;
+  for (const key of path) {
+    if (current && typeof current === 'object' && key in current) {
+      current = current[key];
+    } else {
+      return val; // unresolved — keep as-is
```
zenith-oi, line 78: **FLAG** — [silent-unresolved-references] When a `{reference}` path can't be resolved, `resolveValue` silently returns the original `{reference.string}` and the build succeeds. This means a typo like `{color.violett.600}` in a token file produces a CSS variable with literal value `{color.violett.600}` — a broken design token that ships to consumers without any warning. Add a warning log (or optional strict-mode error) when a reference doesn't resolve, keyed to the path and original value. Token build failures should be loud.

zenith-oi (dismissed):
## Zenith Code Review

**This PR migrates the design token system from fragmented per-package Style Dictionary configs to a single shared build script (scripts/build-tokens.js), and simultaneously moves all token files from the old Tokens Studio format ({value}/{type}) to W3C DTCG format ($value/$type) with a richer, restructured primitive/semantic/theme hierarchy. The code is well-structured overall but has one security/correctness issue and one meaningful architectural concern worth fixing.**

Found 1 BLOCK, 3 FLAGs. See inline comments for details.

**Positive:**
- The consolidation of four divergent per-package build scripts into a single well-documented shared script is a genuine DRY win — the old tokens-brand/build.js alone was 219 lines of custom logic including a hand-rolled reference resolver, which is now replaced with a correct, shared implementation.
- The deterministic token loading order (primitives → light semantic → light theme) with explicit comments explaining why dark theme files are excluded from the default build is exactly the right approach — avoiding the non-deterministic deep-merge footgun is worth calling out.
- Migration to W3C DTCG format ($value/$type/$description) with clean primitive/semantic/theme separation is the right direction for long-term toolchain compatibility.
- Adding `"type": "module"` at the workspace root to unify ESM across all packages is cleaner than per-package configuration.

---

packages/tokens-atmosphere/build.js
zenith-oi, line 7: **BLOCK** — [shell-injection] The package name is inserted directly into a shell command via template literal: `execSync(`node ${join(root, 'scripts/build-tokens.js')} atmosphere`, ...)`. While the package name here is a string literal ('atmosphere'), the pattern is dangerous — if this template is ever generalized or if `root` contains spaces (e.g. a path like '/home/user/my projects/amplify-design-system'), the execSync call will break or be exploitable. The same pattern exists in tokens-brand/build.js L7, tokens-creator/build.js L7, and tokens-foundation/build.js L7. Fix: pass the package name as an argument array using execFileSync instead: `import { execFileSync } from 'child_process'; execFileSync('node', [join(root, 'scripts/build-tokens.js'), 'atmosphere'], { stdio: 'inherit' });` — this avoids shell interpretation entirely.

scripts/build-tokens.js
zenith-oi, line 78: **FLAG** — [silent-error-on-unresolved-reference] When a `{reference.path}` token cannot be resolved (the path doesn't exist in the token tree), `resolveValue` silently returns the raw reference string (e.g. `{color.violet.500}`) and it ends up in the compiled CSS output as a literal string value. This will silently produce broken CSS like `--amp-theme-color-accent: {color.violet.500};`. With the new theme-light/theme-dark JSON structure, any typo in a reference will ship broken tokens to consumers with no build error. Add a warning (at minimum) or error when a reference is unresolved, e.g.: `console.warn(`[build-tokens] Unresolved reference: ${val} in package ${pkg}`); ` so the build pipeline surfaces the breakage.

scripts/build-tokens.js
```diff
@@ -0,0 +1,287 @@
...
+    deepMerge(allTokens, JSON.parse(readFileSync(themeLightFile, 'utf8')));
+  }
+}
+
+// Resolve all references
+const resolved = resolveAll(allTokens, allTokens);
+const flat = flatten(resolved);
+
+// ── CSS prefix ──
+const PREFIX = pkg === 'foundation' ? 'amp' : `amp-${pkg}`;
```
zenith-oi, line 155: **FLAG** — [breaking-change-undocumented] The CSS variable prefix has changed from `--amplify-brand-*` / `--amplify-atmosphere-*` / `--amplify-creator-*` / `--amplify` to `--amp-brand-*` / `--amp-atmosphere-*` / `--amp-creator-*` / `--amp`. Any downstream consumer (one-dashboard-web, odin-platform, creator app) that references these CSS variables by name will silently break — the old variables will just be undefined. Pixel's `token-sync.ts` also reads these canonical files, and the token key names have changed format (e.g. `color.primitive.red-500` → `color.red.500`). This is a hard breaking change in the public contract. It needs a migration note, a semver major bump, or a compatibility shim that emits the old variable names as aliases.

scripts/build-tokens.js
```diff
@@ -0,0 +1,287 @@
...
+// ── Load all JSON token files recursively ──
+function loadJsonFiles(dir) {
+  const tokens = {};
+  if (!existsSync(dir)) return tokens;
+
+  for (const entry of readdirSync(dir)) {
+    const full = join(dir, entry);
+    const stat = statSync(full);
+    if (stat.isDirectory()) {
+      Object.assign(tokens, loadJsonFiles(full));
```
zenith-oi, line 42: **FLAG** — [non-deterministic-merge] The `loadJsonFiles` function uses `readdirSync` without sorting, so file merge order depends on the filesystem's directory enumeration order (which varies between Linux ext4, macOS APFS, and CI containers). If two files in the same directory define the same top-level key and one should win over the other, this is non-deterministic across environments. For the current primitives directory this is probably fine since files are namespaced (colors.json vs spacing.json), but it's a latent correctness hazard. Add `readdirSync(dir).sort()` to guarantee stable ordering.

packages/tokens-foundation/tokens/primitives/colors.json
zenith-oi, line 1: **FLAG** — [pixel-integration-broken] Per CLAUDE.md, Pixel's `token-sync.ts` reads token files from known paths: `packages/tokens-foundation/tokens/*.json`, `packages/tokens-brand/tokens/*.json`, etc. This PR moves foundation tokens into subdirectories (`primitives/` and `semantic/`) and renames all the files. The old flat files (colors.json, radii.json, spacing.json, shadows.json, z-index.json, typography.json, breakpoints.json) are deleted. Pixel's file path assumptions will now 404, breaking drift detection silently. This needs to be coordinated with a Pixel update, or at minimum tracked as a known breakage.

packages/tokens-brand/tokens/theme-light.json
```diff
@@ -0,0 +1,21 @@
...
+    "product": "brand",
+    "mode": "light",
+    "color": {
+      "$type": "color",
+      "accent":          { "$value": "{color.violet.600}" },
+      "accent-hover":    { "$value": "{color.violet.700}" },
+      "accent-light":    { "$value": "{color.violet.100}" },
+      "bg-primary":      { "$value": "{color.stone.50}" },
+      "bg-surface":      { "$value": "{color.white}" },
+      "bg-raised":       { "$value": "#F8F9FD" },
```
zenith-oi, line 13: **FLAG** — [brand-bg-raised-hardcoded-value] `bg-raised` uses a hardcoded value `#F8F9FD` instead of a foundation primitive reference like `{color.stone.50}`. Every other token in this file uses references. This hardcoded value won't be tracked by Pixel's drift detection and will diverge silently if the foundation palette changes. Check whether `{color.stone.50}` (#FAFAF9) is intentionally different from #F8F9FD (the old Brand surface.raised), and if so, add the value to the foundation palette and reference it.

packages/tokens-creator/tokens/theme-light.json
```diff
@@ -0,0 +1,44 @@
...
+      "notice":          { "$value": "{color.amber.600}" },
+      "notice-weak":     { "$value": "{color.amber.50}" },
+      "negative":        { "$value": "{color.red.600}" },
+      "negative-weak":   { "$value": "{color.red.50}" },
+      "offset-strong":   { "$value": "{color.stone.200}" },
+      "offset-medium":   { "$value": "#F1F6FE" },
```
zenith-oi, line 39: **FLAG** — [creator-sdui-hardcoded-values] `offset-medium` (#F1F6FE) and `offset-weak` (#F6FAFF) in the SDUI mapping use hardcoded hex values while every other token in the file uses `{color.*}` references. These were the old Creator legacy values. If they're intentionally kept for backward compatibility with the api-gateway ColorType enum (as the `$description` notes), that's fine — but they should be added as named primitives in the foundation colors palette so Pixel can track them, not buried as magic strings.

zenith-oi (dismissed):
## Zenith Code Review

**This PR is a significant token architecture overhaul: migrating from ad-hoc per-package Style Dictionary configs to a single shared build script, adopting W3C DTCG token format ($value/$type), reorganizing foundation tokens into primitives/semantic layers, and adding proper light/dark theme splits. The work is well-structured and purposeful.**

Found 1 BLOCK, 7 FLAGs. See inline comments for details.

**Positive:**
- Consolidating four divergent per-package build scripts into a single shared script eliminates significant duplication — the brand package alone dropped 218 lines of bespoke (and partially broken) token resolution logic.
- The migration to W3C DTCG format ($value/$type) with explicit primitives/semantic layering is the right architectural direction and sets up proper theme switching.
- The explicit light-semantic-only loading strategy (L122-L135) to avoid non-deterministic deep-merge of conflicting light+dark keys is a thoughtful design decision, well-commented.
- Replacing the old colors.json format (which mixed semantic aliases, product-specific values, and primitive definitions in a single flat file) with a properly layered primitives → semantic → product-theme hierarchy is a meaningful improvement in token hygiene.

---

zenith-oi (commented):
## Zenith Code Review

**Migrates all token packages from Style Dictionary v4 + Tokens Studio to a custom W3C DTCG-format build pipeline, consolidating four per-package build scripts into one shared script and restructuring token files into primitives/semantic/theme layers. The approach is sound and the execution is clean.**

### BLOCK
- **[broken-consumer-contract]** scripts/build-tokens.js:198 — The JS export camelCase conversion (key.replace(/-([a-z0-9])/g, ...)) operates on the flat key WITHOUT the PREFIX, meaning exported names like `colorViolet600` are unchanged from before — but the token key structure has fundamentally changed (e.g. old `colorBrandPrimary` → new `themeColorAccent`, old `colorSduiPrimary` → new `sduiColorPrimary`). Any consumer of `@amplify/tokens-creator/dist/tokens.js` or `@amplify/tokens-brand/dist/tokens.js` importing named exports by their old camelCase names will get undefined at runtime with no build-time error. CLAUDE.md notes Pixel's `token-sync.ts` reads from these packages. Check what named exports `one-dashboard-web`, `odin-platform`, and any React Native code import before merging — the old names are gone.
- **[broken-consumer-contract]** packages/tokens-creator/tokens/theme-light.json:33 — The SDUI `positive-weak` mapping changed from `#E3F6EC` (old `color.sdui.positive-weak`) to `{color.green.50}` which resolves to `#ECFDF5` — a noticeably different green tint. Similarly `negative-weak` changed from `#FFEBEF` to `{color.red.50}` = `#FEF2F2`. The CLAUDE.md notes `sdui-sync-check.yml` validates creator tokens vs api-gateway ColorType enum. That workflow was NOT updated in this PR (it's not in the changed files list). If the sync-check CI job still references old token paths/values it will either fail or silently pass on stale expectations. Verify the sdui-sync-check workflow handles the new structure, and confirm with the Creator app team that these color shifts are intentional.
- **[no-figma-sync-without-replacement]** .github/workflows/figma-sync.yml:1 — The Figma/Tokens Studio sync workflow is deleted with no replacement. CLAUDE.md explicitly documents this workflow as the integration point between Tokens Studio and the build pipeline. Removing it means Tokens Studio pushes to `tokens-studio/**` branches will no longer auto-build, validate, or create PRs. If this is intentional (migrating away from Tokens Studio), that decision should be documented in CLAUDE.md and `tokens.config.json` deletion confirmed as deliberate. If it's unintentional, the workflow needs updating to call the new build script. As-is, the Figma → code pipeline is silently broken.

### FLAG
- **[silent-reference-resolution]** scripts/build-tokens.js:74 — resolveValue() silently returns the raw `{reference}` string when a reference can't be resolved (line 74: `return val; // unresolved — keep as-is`). This means a token with a broken reference like `{color.violet.999}` will emit `--amp-theme-color-accent: {color.violet.999};` into CSS — valid syntax that silently produces no visible color. The validate-tokens.js script does catch broken refs, but only if run. Consider throwing or logging a warning here so build failures are loud, not silent.
- **[non-deterministic-file-load-order]** scripts/build-tokens.js:38 — loadJsonFiles() uses readdirSync() without sorting, and filesystem readdir order is not guaranteed to be stable across platforms/OS. For the primitives directory this is fine (disjoint keys), but if two JSON files in the same directory define overlapping keys, the merge result is OS-dependent. The current token structure appears safe, but this is a latent bug waiting for someone to add a file. Add `readdirSync(dir).sort()` to make the load order deterministic.
- **[react-native-numeric-parsing]** scripts/build-tokens.js:259 — buildReactNative() uses parseInt() to parse font-size and spacing values (e.g. `parseInt('14px', 10)` → 14). This is fine for px values, but if any token value is `rem`, `em`, or `%` (possible after the W3C migration), parseInt will silently return the wrong number (e.g. `parseInt('1.5rem', 10)` → 1). The old build script used a dedicated `numericIfPossible` that stripped 'px' explicitly. Consider checking the unit before stripping, or at minimum adding a warning when a non-px unit is encountered.

**Positive:**
- Consolidating four near-identical per-package build scripts into one shared script is exactly the right call — the old code had the same resolver logic copy-pasted four times with subtle divergences.
- The explicit light-theme-first loading strategy (lines 120-144) with a comment explaining WHY is good defensive engineering — non-deterministic deep-merge of light+dark was a real bug waiting to happen.
- The W3C DTCG migration (using $value/$type instead of value/type) is the right direction for toolchain compatibility.
- findBrokenRefs() in validate-tokens.js properly checks references against the merged foundation+product context rather than just the product tokens in isolation — this is the correct resolution scope.

---

zenith-oi (commented):
## Zenith Code Review

**Large refactor migrating all four token packages from Style Dictionary v4 to a bespoke W3C DTCG-format build pipeline, consolidating four near-identical per-package build scripts into one shared script, and restructuring token files into primitives/semantic/theme layers. The code is generally well-structured but has two issues worth flagging.**

Found 3 FLAGs. See inline comments for details.

**Positive:**
- Eliminating four nearly-identical per-package build scripts in favour of one shared script is the right call — the old code had meaningful divergence between packages (brand had a hardcoded Tailwind preset with brittle string interpolation, creator used Style Dictionary's API differently) that would have caused silent inconsistencies.
- The deterministic load order comment at L120-L126 is exactly the right thing to document — non-deterministic deep-merge of light+dark semantic files is a real trap and calling it out explicitly shows good defensive thinking.
- The reference integrity check in validate-tokens.js (findBrokenRefs, L64-L89) is a meaningful improvement over the old validator which only checked that `value` fields were non-empty.

---

packages/tokens-foundation/tokens/primitives/colors.json
zenith-oi, line 1: **FLAG** — [broken-contract / pixel-integration] CLAUDE.md explicitly documents that Pixel's token-sync.ts reads from specific canonical paths: `packages/tokens-foundation/tokens/*.json`, `packages/tokens-brand/tokens/*.json`, etc. This PR restructures foundation tokens into `tokens/primitives/` and `tokens/semantic/` subdirectories and renames every color scale (old: `color.primitive.purple-500`, new: `color.violet.600`). Pixel will stop resolving its expected paths silently — it will see zero tokens or raise false drift alerts against all three products simultaneously. Before merging, verify that Pixel's `token-sync.ts` path patterns cover recursive subdirectory scanning, and confirm the rename from the old purple/neutral/red/green/amber palette to the new violet/stone/red/green/amber palette won't break Pixel's drift comparisons against deployed product repos.

scripts/build-tokens.js
```diff
@@ -0,0 +1,288 @@
...
+  const match = val.match(/^\{(.+)\}$/);
+  if (!match) return val;
+
+  const path = match[1].split('.');
+  let current = root;
+  for (const key of path) {
+    if (current && typeof current === 'object' && key in current) {
+      current = current[key];
+    } else {
+      console.warn(`WARNING: unresolved reference ${val} — token will output raw reference string`);
```
zenith-oi: **FLAG** — [no-console-log-in-production] `console.warn` is used for unresolved token references at L74 and `console.error`/`console.log` throughout. For a build script this is fine, but the validate script at scripts/validate-tokens.js similarly uses only `console.log` for all error output (L110, L124, L159). Errors and warnings should go to stderr (`console.error`) so CI pipelines can distinguish failure output from informational output without relying solely on the exit code. Minor, but affects CI log parsing.

scripts/build-tokens.js
```diff
@@ -0,0 +1,288 @@
...
+const packageDir = join(ROOT, `packages/tokens-${pkg}/tokens`);
+
+/**
+ * Load foundation tokens deterministically:
+ * 1. Always load primitives (stable base values)
+ * 2. Load ONLY colors-light.json for the default theme (avoids non-deterministic
+ *    deep-merge of light+dark which define the same semantic keys)
+ * 3. Overlay product tokens on top
+ */
+let allTokens = {};
```
zenith-oi, line 130: **FLAG** — [missing-dark-theme-css-output] The build explicitly loads only the light semantic file and only `theme-light.json` for product packages (L133-L145), producing a single `:root {}` block. Dark theme tokens are loaded into the JSON/JS/SCSS outputs only if both theme files happen to be merged — but they aren't: dark files are never loaded at all during the build. This means `variables.css` has no `@media (prefers-color-scheme: dark)` block and no `.dark` selector block, even though dark theme token files exist for all four packages. Consumers will get no dark mode CSS from this pipeline. Either the dark theme tokens need to be loaded and emitted as a separate selector/media-query block, or it needs to be documented clearly that dark mode CSS is intentionally out of scope and consumers must implement it themselves.

zenith-oi (commented):
## Zenith Code Review

**This PR migrates all token packages from Style Dictionary v4 to a custom W3C DTCG-format build pipeline, consolidating four per-package build scripts into one shared script, restructuring token files under primitives/semantic/theme subdirectories, and removing the Figma/Tokens Studio sync workflow. The migration is substantial and well-structured, but has one breaking contract issue and one meaningful bug worth fixing.**

### BLOCK
- **[broken-contract]** CLAUDE.md:1 — The CLAUDE.md explicitly documents that Pixel's `token-sync.ts` reads from specific paths like `packages/tokens-foundation/tokens/*.json`, `packages/tokens-brand/tokens/*.json`, etc. This PR reorganizes those files into subdirectories (`primitives/`, `semantic/`, `theme-light.json`, `theme-dark.json`) and removes the top-level flat files that Pixel is reading via GitHub API. Pixel will now fail to find the files it polls for drift detection. Before merging, either update Pixel's `token-sync.ts` path configuration to match the new structure, or confirm with the Pixel team that the path globs are recursive (`**/*.json`) and will still resolve correctly.
- **[broken-contract]** .github/workflows/figma-sync.yml:1 — The Figma/Tokens Studio sync workflow is deleted with no replacement. CLAUDE.md states 'All token changes should go through Figma sync or direct PR (Pixel will cascade)' and the workflow was the automation for this. Deleting it breaks the design → token → build pipeline that product teams depend on. If this is intentional (moving to manual PRs only), that decision needs to be explicitly documented in CLAUDE.md and communicated to Pixel team, since Pixel's brand-cascade and token-sync flows may depend on that branch naming convention (`tokens-studio/**`).

### FLAG
- **[silent-failure]** scripts/build-tokens.js:42 — In `loadJsonFiles`, `deepMerge(tokens, data)` at L45 silently overwrites conflicting top-level keys when two JSON files define the same root key. For example, if `theme-light.json` and `theme-dark.json` are both present in the same directory and loaded together (which happens in `validate-tokens.js` at L144 where `loadJsonFilesRecursive` loads the entire tokens dir including both themes), the dark theme silently wins over light or vice versa depending on sort order. The `build-tokens.js` avoids this correctly by only loading `theme-light.json` explicitly, but `validate-tokens.js` calls `loadJsonFilesRecursive` on the full `tokens/` directory at L150 for reference integrity checking, meaning both `theme-light.json` and `theme-dark.json` get merged together — the reference check runs against a non-deterministic blend of light and dark values. This means broken dark-theme references could pass validation undetected. Consider running reference checks against each theme file independently.
- **[missing-error-handling]** packages/tokens-brand/build.js:7 — `execSync` will throw synchronously if the child process exits non-zero, but without a try/catch the raw Node stack trace is the only output — no context about which package failed. More importantly, if `scripts/build-tokens.js` is missing or the path resolution fails silently, the error message is unhelpful in CI. Consider wrapping with `try/catch` and re-throwing with the package name, or pass `{ stdio: 'inherit', env: process.env }` and let the child script's output surface directly (the `stdio: 'inherit'` is already there, so this is minor — but the lack of any catch is worth noting given this runs in CI for all four packages).
- **[console.log-in-production]** scripts/build-tokens.js:74 — `console.warn` at L74 for unresolved references is appropriate in a build script, but unresolved references silently produce raw `{reference.path}` strings in the output CSS/JS/JSON. This means a token file referencing `{color.violet.600}` that fails to resolve would ship as the literal string `{color.violet.600}` in `variables.css` without failing the build. Given the validate step exists separately, this may be intentional, but it's worth either: (a) collecting unresolved refs and exiting with code 1 at end of build, or (b) documenting that `validate-tokens.js` must always run before `build-tokens.js` in CI.

**Positive:**
- Consolidating four near-identical per-package build scripts into a single shared `scripts/build-tokens.js` is exactly the right call — the old code had the same resolver/flattener logic duplicated across all packages with subtle inconsistencies.
- The explicit light-only loading strategy in `build-tokens.js` (L122-L145) with a clear comment explaining why both themes aren't merged is good defensive design that prevents non-deterministic output.
- W3C DTCG format (`$value`, `$type`, `$description`) is the correct direction — the old Style Dictionary `value`/`type` format was already being handled inconsistently across packages.
- The new token taxonomy (primitives → semantic → theme) with proper subdirectory structure is clean and the separation of concerns is well thought out.

---

zenith-oi (commented):
## Zenith Code Review

**Major refactor migrating all token packages from legacy Style Dictionary format (value/type) to W3C DTCG format ($value/$type), replacing per-package build scripts with a single shared build-tokens.js, and restructuring token files into primitives/semantic/theme layers. The architecture is sound and the implementation is thorough, but there are a few real issues worth flagging.**

### BLOCK
- **[broken-contract: Pixel token path dependency]** packages/tokens-foundation/tokens/primitives/colors.json:1 — CLAUDE.md explicitly documents that Pixel's token-sync.ts reads from specific paths: `packages/tokens-foundation/tokens/*.json`, `packages/tokens-brand/tokens/*.json`, etc. This PR restructures foundation tokens into `primitives/` and `semantic/` subdirectories and renames nearly every token key (e.g. `color.primitive.purple-500` → `color.violet.600`, `color.semantic.error` → `semantic.status-error`, `typography.weight.bold` → `font.weight.bold`). Pixel's token drift detection compares these canonical JSON paths and key names. This will cause Pixel to flag every consuming product repo as 'drifted' from the canonical source on next sync, and may break automated cascade logic in `brand-cascade.ts`. Before merging, confirm with the Pixel team (or check `token-sync.ts` and `brand-cascade.ts`) exactly which paths and key names are accessed, and either: (a) update Pixel's path references, or (b) keep backward-compatible alias files. The color palette has also fundamentally changed (purple-500 #6531FF → violet-600 #7C3AED), which is a visual breaking change across all three products.
- **[broken-contract: sdui-sync-check.yml removed validation]** scripts/validate-tokens.js:1 — CLAUDE.md mentions `sdui-sync-check.yml` which validates creator tokens against the api-gateway ColorType enum. The old validate-tokens.js had an explicit check for this (Check 3 in the original file header). The new validation removes all cross-package color consistency checks and the sdui/api-gateway enum validation entirely. The new creator tokens also change SDUI color values — e.g. `positive` changes from #21C179 to `{color.green.600}` which resolves to #059669, and `negative` changes from #FD5154 to `{color.red.600}` which resolves to #DC2626. If the api-gateway ColorType enum or the Creator app's React Native code hard-references these hex values, this is a runtime breakage in the mobile app. Verify the sdui-sync-check workflow still runs and passes with the new token values before merging.
- **[shell-injection via execSync]** packages/tokens-brand/build.js:7 — All four package build.js files call `execSync(\`node ${join(root, 'scripts/build-tokens.js')} brand\`, { stdio: 'inherit' })` with a hardcoded string argument, which is safe. However, `scripts/build-tokens.js` takes `process.argv[2]` and validates it against VALID_PACKAGES before use, which is correct. The actual risk is that `join(root, 'scripts/build-tokens.js')` could resolve to an unexpected path if the package is installed in an unusual location (e.g. symlinked monorepo). More critically: the path constructed via `join(dirname(fileURLToPath(import.meta.url)), '../..')` from within `packages/tokens-brand/` goes up two levels — to the monorepo root. If this package were ever published and installed as a dependency in another project, `root` would point to that project's root and `scripts/build-tokens.js` would not exist, producing a confusing silent failure rather than a clear error. Add an `existsSync` check on the script path before exec, or use a workspace-relative path from package.json scripts instead.

### FLAG
- **[no-console.log in production build scripts]** scripts/build-tokens.js:77 — `console.warn` is used for unresolved references during build (L77), and `console.error` at L295-299. These are appropriate for a build script. However, `unresolvedRefs` is a module-level mutable array (L63) that accumulates across the entire build run. If `resolveValue` is ever called in a context where the same ref appears multiple times (e.g. multiple tokens referencing the same broken path), the same ref will be pushed multiple times. This inflates the error count reported at L295 and makes the output misleading. Deduplicate with a Set: `const unresolvedRefs = new Set()` and push with `.add()`, then iterate with `[...unresolvedRefs]`.
- **[missing-ci-workflow: figma-sync.yml deleted without replacement]** .github/workflows/figma-sync.yml:1 — The figma-sync.yml workflow (Tokens Studio → build → validate → auto-PR pipeline) has been deleted entirely with no replacement. CLAUDE.md documents this as part of the CI/CD pipeline. If Tokens Studio is no longer the source of truth, that's a legitimate architectural change, but it should be explicitly documented. If it's still intended to sync from Figma, the workflow needs to be updated for the new token format and file structure (e.g. the new primitives/ and semantic/ paths). The deletion without a replacement leaves the Figma sync pipeline broken.
- **[incomplete-output: dark theme not built]** scripts/build-tokens.js:130 — The build script only loads `colors-light.json` as the semantic layer and `theme-light.json` for product packages (L136-148). Dark theme tokens (`theme-dark.json`, `colors-dark.json`) are parsed and validated but never built into dist artifacts. There is no `variables-dark.css`, `tailwind-dark.css`, or equivalent output. The token files for dark themes exist and are validated, but consumers cannot actually use them — there's no CSS `@media (prefers-color-scheme: dark)` block and no separate dark dist file. This should either be intentional (with a comment explaining the dark theme delivery strategy) or dark theme outputs should be generated.
- **[path-traversal risk in loadJsonFiles]** scripts/build-tokens.js:34 — `loadJsonFiles` recurses into all subdirectories of the tokens dir (L41-42) using `Object.assign` for directories, then `deepMerge` for JSON files. There's an inconsistency: directory recursion uses `Object.assign(tokens, loadJsonFiles(full))` (L42) which does a shallow merge of the top-level keys returned, while JSON file loading uses `deepMerge(tokens, data)` (L45). For directory traversal, if two subdirectory loads return overlapping top-level keys (e.g. both `primitives/` and `semantic/` had a `color` key), `Object.assign` would clobber the first with the second rather than deep-merging. The same `loadJsonFilesRecursive` in validate-tokens.js correctly uses `deepMerge` for both cases (L30, L33). Fix L42 in build-tokens.js to use `deepMerge(tokens, loadJsonFiles(full))` for consistency.

**Positive:**
- The shared build-tokens.js consolidation is the right call — eliminates ~350 lines of duplicated per-package build logic and ensures all packages produce identical output formats.
- Deterministic theme loading (explicitly loading only colors-light.json rather than recursively merging light+dark) is well-reasoned and the comment at L124-128 explains it clearly.
- The W3C DTCG format migration is clean — $value/$type throughout, $-prefixed metadata keys skipped correctly in all traversal functions.
- Reference integrity validation in validate-tokens.js is genuinely improved — per-theme validation prevents false positives from non-deterministic merging, and findBrokenRefs correctly resolves multi-segment paths.
- Backward compat annotations on legacy SDUI values in creator/theme-light.json (positive-weak, notice-weak, negative-weak) show good awareness of downstream consumers.

---

PR #1, opened 2026-04-15
docs(zenith): update CLAUDE.md

zenith-oi:
## Summary
- **Trigger:** commit `4df26e1`
- **File updated:** `CLAUDE.md`
- **Confidence:** 🟢 0.87
- **Reason:** New @amplify/feature-flags package introduces a significant service integration with multiple public APIs (FeatureFlagClient, Provider, hooks, Gate component) that engineers in this repo need to understand for usage and maintenance.

This is an automated documentation update by Zenith. Auto-merge eligible.

---
Generated by [Zenith AI CTO](https://github.com/one-impression/zenith-coding-agent)

PR #9, opened 2026-04-20
feat: Phase 2 — 25-component library with Canvas CSS variables

apaksh-gupta:
## Summary
- 25 React components using `var(--amp-semantic-*)` CSS custom properties
- Zero hardcoded colors — all styling flows through design tokens
- 3 existing components rebuilt (Card, EmptyState, Skeleton)
- 20 new components created (Input, DataTable, Dialog, Toast, etc.)
- Full TypeScript types, forwardRef, ARIA accessibility

## Components
Core: Button, IconButton, Badge, StatusTag, Card
Form: Input, Textarea, Select, SearchInput, Checkbox, Switch
Data: DataTable, MetricCard, ProgressBar, Timeline
Navigation: Tabs, Breadcrumb, Pagination
Feedback: Dialog, Toast, Tooltip, EmptyState, Skeleton
Layout: Avatar, Separator

## Design spec
- Canvas theme: Violet #7C3AED, Stone neutrals, 16px radius, 14px body
- All colors via CSS variables for automatic theme propagation

## Test plan
- [x] Zero hardcoded hex colors (grep verified)
- [x] All components export from index.ts
- [ ] CI build passes
- [ ] Storybook renders all components

### DIFF
--- a/packages/ui/src/components/Card/Card.tsx
+++ b/packages/ui/src/components/Card/Card.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { cn } from '../../lib/cn';
 
-export type CardVariant = 'default' | 'elevated' | 'interactive';
-export type CardPadding = 'sm' | 'md' | 'lg';
+export type CardVariant = 'default' | 'elevated' | 'interactive' | 'outlined' | 'ghost';
+export type CardPadding = 'none' | 'sm' | 'md' | 'lg';
 
 export interface CardProps extends React.HTMLAttributes<HTMLDivElement> {
   variant?: CardVariant;
@@ -37,9 +37,14 @@ const variantClasses: Record<CardVariant, string> = {
     'bg-[var(--amp-semantic-bg-surface)] shadow-lg border border-[var(--amp-semantic-border-default)]',
   interactive:
     'bg-[var(--amp-semantic-bg-surface)] border border-[var(--amp-semantic-border-default)] cursor-pointer hover:shadow-lg transition-shadow duration-150',
+  outlined:
+    'bg-transparent border border-[var(--amp-semantic-border-default)]',
+  ghost:
+    'bg-transparent border-none',
 };
 
 const paddingClasses: Record<CardPadding, string> = {
+  none: 'p-0',
   sm: 'p-3',
   md: 'p-4',
   lg: 'p-6',
--- a/packages/ui/src/components/Card/Card.tsx
+++ b/packages/ui/src/components/Card/Card.tsx
@@ -1,72 +1,123 @@
 import React from 'react';
 import { cn } from '../../lib/cn';
 
-export type CardVariant = 'default' | 'elevated' | 'outlined' | 'ghost';
-export type CardPadding = 'none' | 'sm' | 'md' | 'lg';
+export type CardVariant = 'default' | 'elevated' | 'interactive';
+export type CardPadding = 'sm' | 'md' | 'lg';
 
-/**
- * Card component props.
- *
- * **Accessibility note:** When `onClick` is provided the Card renders with
- * `role="button"`. Screen-reader users need an accessible name, so callers
- * should supply an `aria-label` (or ensure the card's text content is
- * sufficiently descriptive) whenever the card is interactive.
- */
-export interface CardProps {
+export interface CardProps extends React.HTMLAttributes<HTMLDivElement> {
   variant?: CardVariant;
   padding?: CardPadding;
   children: React.ReactNode;
-  as?: 'div' | 'article' | 'section' | 'li';
-  onClick?: () => void;
-  /** Required for accessibility when `onClick` is provided. */
-  'aria-label'?: string;
-  className?: string;
+}
+
+export interface CardHeaderProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode;
+}
+
+export interface CardTitleProps extends React.HTMLAttributes<HTMLHeadingElement> {
+  children: React.ReactNode;
+}
+
+export interface CardDescriptionProps extends React.HTMLAttributes<HTMLParagraphElement> {
+  children: React.ReactNode;
+}
+
+export interface CardContentProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode;
+}
+
+export interface CardFooterProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode;
 }
 
 const variantClasses: Record<CardVariant, string> = {
-  default: 'bg-surface border border-border rounded-lg shadow-sm',
-  elevated: 'bg-surface border border-border rounded-lg shadow-md',
-  outlined: 'bg-transparent border-2 border-brand/20 rounded-lg',
-  ghost: 'bg-surface-raised rounded-lg',
+  default:
+    'border border-[var(--amp-semantic-border-default)] bg-[var(--amp-semantic-bg-surface)]',
+  elevated:
+    'bg-[var(--amp-semantic-bg-surface)] shadow-lg border border-[var(--amp-semantic-border-default)]',
+  interactive:
+    'bg-[var(--amp-semantic-bg-surface)] border border-[var(--amp-semantic-border-default)] cursor-pointer hover:shadow-lg transition-shadow duration-150',
 };
 
 const paddingClasses: Record<CardPadding, string> = {
-  none: '',
   sm: 'p-3',
   md: 'p-4',
   lg: 'p-6',
 };
 
-export const Card: React.FC<CardProps> = ({
-  variant = 'default',
-  padding = 'md',
-  children,
-  as: Tag = 'div',
-  onClick,
-  'aria-label': ariaLabel,
-  className,
-}) => {
-  return (
-    <Tag
-      className={cn(
-        'transition-all duration-150',
-        variantClasses[variant],
-        paddingClasses[padding],
-        onClick && 'cursor-pointer hover:shadow-md active:scale-[0.99]',
-        className
-      )}
-      onClick={onClick}
-      aria-label={ariaLabel}
-      role={onClick ? 'button' : undefined}
-      tabIndex={onClick ? 0 : undefined}
-      onKeyDown={onClick ? (e) => {
-        if (e.key === 'Enter' || e.key === ' ') {
-          e.preventDefault();
-          onClick();
-        }
-      } : undefined}
+export const Card = React.forwardRef<HTMLDivElement, CardProps>(
+  ({ variant = 'default', padding = 'md', children, className, ...props }, ref) => {
+    return (
+      <div
+        ref={ref}
+        className={cn(
+          'rounded-[16px]',
+          variantClasses[variant],
+          paddingClasses[padding],
+          className
+        )}
+        {...props}
+      >
+        {children}
+      </div>
+    );
+  }
+);
+Card.displayName = 'Card';
+
+export const CardHeader = React.forwardRef<HTMLDivElement, CardHeaderProps>(
+  ({ children, className, ...props }, ref) => (
+    <div ref={ref} className={cn('flex flex-col gap-1.5 pb-3', className)} {...props}>
+      {children}
+    </div>
+  )
+);
+CardHeader.displayName = 'CardHeader';
+
+export const CardTitle = React.forwardRef<HTMLHeadingElement, CardTitleProps>(
+  ({ children, className, ...props }, ref) => (
+    <h3
+      ref={ref}
+      className={cn('text-[16px] font-semibold text-[var(--amp-semantic-text-primary)]', className)}
+      {...props}
     >
       {children}
-    </Tag>
-  );
-};
+    </h3>
+  )
+);
+CardTitle.displayName = 'CardTitle';
+
+export const CardDescription = React.forwardRef<HTMLParagraphElement, CardDescriptionProps>(
+  ({ children, className, ...props }, ref) => (
+    <p
+      ref={ref}
+      className={cn('text-[14px] text-[var(--amp-semantic-text-secondary)]', className)}
+      {...props}
+    >
+      {children}
+    </p>
+  )
+);
+CardDescription.displayName = 'CardDescription';
+
+export const CardContent = React.forwardRef<HTMLDivElement, CardContentProps>(
+  ({ children, className, ...props }, ref) => (
+    <div ref={ref} className={cn('py-2', className)} {...props}>
+      {children}
+    </div>
+  )
+);
+CardContent.displayName = 'CardContent';
+
+export const CardFooter = React.forwardRef<HTMLDivElement, CardFooterProps>(
+  ({ children, className, ...props }, ref) => (
+    <div
+      ref={ref}
+      className={cn('flex items-center pt-3 border-t border-[var(--amp-semantic-border-default)]', className)}
+      {...props}
+    >
+      {children}
+    </div>
+  )
+);
+CardFooter.displayName = 'CardFooter';
--- a/packages/ui/src/components/DataTable/DataTable.tsx
+++ b/packages/ui/src/components/DataTable/DataTable.tsx
@@ -1,0 +1,130 @@
+import React, { useState, useMemo } from 'react';
+import { cn } from '../../lib/cn';
+
+export interface DataTableColumn<T> {
+  key: string;
+  header: string;
+  render?: (item: T) => React.ReactNode;
+  sortable?: boolean;
+}
+
+export interface DataTableProps<T> {
+  columns: DataTableColumn<T>[];
+  data: T[];
+  loading?: boolean;
+  emptyMessage?: string;
+  className?: string;
+}
+
+type SortDir = 'asc' | 'desc';
+
+export function DataTable<T extends Record<string, unknown>>({
+  columns,
+  data,
+  loading = false,
+  emptyMessage = 'No data available',
+  className,
+}: DataTableProps<T>) {
+  const [sortKey, setSortKey] = useState<string | null>(null);
+  const [sortDir, setSortDir] = useState<SortDir>('asc');
+
+  const handleSort = (key: string) => {
+    if (sortKey === key) {
+      setSortDir((d) => (d === 'asc' ? 'desc' : 'asc'));
+    } else {
+      setSortKey(key);
+      setSortDir('asc');
+    }
+  };
+
+  const sortedData = useMemo(() => {
+    if (!sortKey) return data;
+    return [...data].sort((a, b) => {
+      const aVal = a[sortKey];
+      const bVal = b[sortKey];
+      if (aVal == null && bVal == null) return 0;
+      if (aVal == null) return 1;
+      if (bVal == null) return -1;
+      if (aVal < bVal) return sortDir === 'asc' ? -1 : 1;
+      if (aVal > bVal) return sortDir === 'asc' ? 1 : -1;
+      return 0;
+    });
+  }, [data, sortKey, sortDir]);
+
+  if (loading) {
+    return (
+      <div className={cn('rounded-[16px] border border-[var(--amp-semantic-border-default)] overflow-hidden', className)}>
+        <div className="p-8 text-center text-[14px] text-[var(--amp-semantic-text-muted)]">
+          Loading...
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <div className={cn('rounded-[16px] border border-[var(--amp-semantic-border-default)] overflow-hidden', className)}>
+      <table className="w-full border-collapse">
+        <thead>
+          <tr className="bg-[var(--amp-semantic-bg-sunken)]">
+            {columns.map((col) => (
+              <th
+                key={col.key}
+                className={cn(
+                  'px-4 py-3 text-left text-[12px] font-semibold uppercase tracking-wide text-[var(--amp-semantic-text-secondary)]',
+                  col.sortable && 'cursor-pointer select-none hover:text-[var(--amp-semantic-text-primary)]'
+                )}
+                onClick={col.sortable ? () => handleSort(col.key) : undefined}
+                aria-sort={sortKey === col.key ? (sortDir === 'asc' ? 'ascending' : 'descending') : undefined}
+              >
+                <span className="inline-flex items-center gap-1">
+                  {col.header}
+                  {col.sortable && sortKey === col.key && (
+                    <svg className="w-3 h-3" fill="none" viewBox="0 0 24 24" stroke="currentColor" strokeWidth={2}>
+                      <path
+                        strokeLinecap="round"
+                        strokeLinejoin="round"
+                        d={sortDir === 'asc' ? 'M5 15l7-7 7 7' : 'M19 9l-7 7-7-7'}
+                      />
+                    </svg>
+                  )}
+                </span>
+              </th>
+            ))}
+          </tr>
+        </thead>
+        <tbody>
+          {sortedData.length === 0 ? (
+            <tr>
+              <td
+                colSpan={columns.length}
+                className="px-4 py-8 text-center text-[14px] text-[var(--amp-semantic-text-muted)]"
+              >
+                {emptyMessage}
+              </td>
+            </tr>
+          ) : (
+            sortedData.map((row, idx) => (
+              <tr
+                key={idx}
+                className={cn(
+                  'border-t border-[var(--amp-semantic-border-default)]',
+                  'hover:bg-[var(--amp-semantic-bg-raised)] transition-colors duration-100',
+                  idx % 2 === 1 && 'bg-[var(--amp-semantic-bg-sunken)]/30'
+                )}
+              >
+                {columns.map((col) => (
+                  <td
+                    key={col.key}
+                    className="px-4 py-3 text-[14px] text-[var(--amp-semantic-text-primary)]"
+                  >
+                    {col.render ? col.render(row) : (row[col.key] as React.ReactNode)}
+                  </td>
+                ))}
+              </tr>
+            ))
+          )}
+        </tbody>
+      </table>
+    </div>
+  );
+}
--- a/packages/ui/src/components/DataTable/DataTable.tsx
+++ b/packages/ui/src/components/DataTable/DataTable.tsx
@@ -11,6 +11,7 @@ export interface DataTableColumn<T> {
 export interface DataTableProps<T> {
   columns: DataTableColumn<T>[];
   data: T[];
+  rowKey?: keyof T | ((item: T) => string);
   loading?: boolean;
   emptyMessage?: string;
   className?: string;
@@ -21,6 +22,7 @@ type SortDir = 'asc' | 'desc';
 export function DataTable<T extends Record<string, unknown>>({
   columns,
   data,
+  rowKey,
   loading = false,
   emptyMessage = 'No data available',
   className,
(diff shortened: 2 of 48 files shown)


packages/ui/src/components/Card/Card.tsx
```diff
@@ -1,72 +1,123 @@
 import React from 'react';
 import { cn } from '../../lib/cn';
 
-export type CardVariant = 'default' | 'elevated' | 'outlined' | 'ghost';
-export type CardPadding = 'none' | 'sm' | 'md' | 'lg';
+export type CardVariant = 'default' | 'elevated' | 'interactive';
```
zenith-oi: **BLOCK** — [broken-contract] CardVariant removes 'outlined' and 'ghost', and CardProps removes the `as` prop and explicit `onClick`/`aria-label`. Any consumer in Brand, Creator, or Atmosphere using `variant='outlined'`, `variant='ghost'`, `as='article'`, or the old action API will break silently at compile time (TypeScript error) or runtime. This is a public package consumed by 3 products. Before merging, either keep the removed variants/props with a deprecation path, or confirm with all downstream consumers that no usage exists. The CLAUDE.md notes Pixel tracks cross-product dependencies — run `cross-product-deps.ts` against these removed types before merge.

packages/ui/src/components/Skeleton/Skeleton.tsx
```diff
@@ -1,34 +1,29 @@
...
 import { cn } from '../../lib/cn';
 
-export interface SkeletonProps {
-  width?: string | number;
-  height?: string | number;
-  variant?: 'text' | 'rectangular' | 'circular';
-  className?: string;
+export type SkeletonVariant = 'text' | 'circular' | 'rectangular';
+
+export interface SkeletonProps extends React.HTMLAttributes<HTMLDivElement> {
```
zenith-oi, line 6: **BLOCK** — [broken-contract] SkeletonProps removes `width` and `height` props. Consumers who sized Skeletons via `width={200} height={16}` will now get TypeScript errors (the props are spread onto HTMLDivElement so they'll still reach the DOM as attributes, but the typed API contract is gone). Confirm no existing usage before removing these from the public interface.

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+    document.addEventListener('keydown', handleKeyDown);
+    document.body.style.overflow = 'hidden';
+
+    return () => {
+      document.removeEventListener('keydown', handleKeyDown);
+      document.body.style.overflow = '';
+    };
+  }, [open, onClose]);
+
+  useEffect(() => {
```
zenith-oi, line 79: **FLAG** — [accessibility-focus-trap] The Dialog focuses the container div on open (line 44) but has no focus trap. A keyboard user can Tab past the dialog's actions and into background content, violating WCAG 2.1 SC 2.1.2. For a shared component library used across 3 products this is a real accessibility bug. Implement a basic focus trap: on mount, collect all focusable elements inside dialogRef, intercept Tab/Shift+Tab in the keydown handler to cycle within them. Additionally, `aria-labelledby='dialog-title'` and `aria-describedby='dialog-description'` use static string IDs — if two dialogs are ever rendered (unlikely given the modal pattern but possible in tests), IDs will collide. Prefer generated IDs via `useId()`.

packages/ui/src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,130 @@
...
+                colSpan={columns.length}
+                className="px-4 py-8 text-center text-[14px] text-[var(--amp-semantic-text-muted)]"
+              >
+                {emptyMessage}
+              </td>
+            </tr>
+          ) : (
+            sortedData.map((row, idx) => (
+              <tr
+                key={idx}
```
zenith-oi: **FLAG** — [missing-key-stability] Rows are keyed by array index (`key={idx}`). When data is sorted, all row keys change, causing React to remount every row instead of reordering. If rows contain interactive children (inputs, checkboxes) this will lose local state. The `DataTableColumn` interface has a `key` field but there's no `rowKey` prop or convention for data identity. Add a `rowKey: keyof T` prop to `DataTableProps` and use it as `key={String(row[rowKey])}`.

packages/ui/src/components/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export const Timeline: React.FC<TimelineProps> = ({ events, className }) => {
+  return (
+    <div className={cn('relative', className)}>
+      {events.map((event, idx) => {
+        const isLast = idx === events.length - 1;
+        const status = event.status || 'default';
+
+        return (
+          <div key={idx} className="relative flex gap-4 pb-6 last:pb-0">
```
zenith-oi: **FLAG** — [missing-key-stability] Timeline events are keyed by index (`key={idx}`). If events are reordered or filtered (common for activity feeds), React will reuse DOM nodes incorrectly. Add an optional `id` field to `TimelineEvent` and prefer it as the key, falling back to index only when absent.

packages/ui/src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onChange,
+  placeholder = 'Search...',
+  onClear,
+  debounceMs = 300,
+  className,
+}) => {
+  const [internal, setInternal] = useState(value);
+  const timerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  useEffect(() => {
```
zenith-oi, line 25: **FLAG** — [uncontrolled-external-sync] The `useEffect` that syncs external `value` → `internal` state (lines 24-26) will overwrite in-flight user input on every re-render where the parent re-provides the same prop. This is the classic "derived state" anti-pattern. The component should be either fully controlled (no internal state, just debounce the callback) or uncontrolled with `defaultValue`. As written, if the parent re-renders while the user is typing mid-debounce, the internal value will be reset to the last committed value, losing keystrokes.

zenith-oi (dismissed):
## Zenith Code Review

**Large batch addition of 20+ new UI components to the design system, plus a breaking refactor of Card into compound sub-components and migrations of Skeleton/EmptyState to CSS variable tokens. The work is generally high quality, but there are two real issues worth fixing: a breaking contract change on Card that removes the `as` prop and `'outlined'`/`'ghost'` variants, and a focus-trap omission in Dialog that makes it non-functional for keyboard users.**

Found 2 BLOCKs, 4 FLAGs. See inline comments for details.

**Positive:**
- Consistent use of CSS custom properties throughout — no hardcoded color values in any new component, respecting the CLAUDE.md rule #1.
- Dialog, IconButton, and form inputs all handle focus-visible ring styles correctly, and interactive elements have explicit aria-label, aria-invalid, aria-describedby wiring.
- React.forwardRef is correctly applied to all components that need ref forwarding (Card, Input, Textarea, Select, Separator, Skeleton, IconButton), with displayName set on each — good DX for React DevTools.
- ProgressBar correctly clamps the value and sets all three ARIA attributes (aria-valuenow, aria-valuemin, aria-valuemax).
- SearchInput correctly clears the pending debounce timer on unmount to avoid setState-after-unmount.

---

packages/ui/src/components/Card/Card.tsx
zenith-oi: **BLOCK** — [broken-contract: CardVariant removal] The `CardVariant` type dropped two values that were in the previous public API: `'outlined'` and `'ghost'`. Any consumer passing `variant="outlined"` or `variant="ghost"` will now silently hit an undefined entry in `variantClasses` and render with only the base rounded class — no visible border, no background. TypeScript will catch this at compile time in consumers that are also TypeScript, but any JS consumers or stringified values (e.g. from CMS/SDUI config) will break at runtime with no error. Similarly, `CardPadding` dropped `'none'`. Before merge, either keep the removed variants as deprecated aliases mapped to the nearest equivalent, or do a confirmed cross-repo audit showing no consumers of `outlined`, `ghost`, or `padding="none"`.

packages/ui/src/components/Skeleton/Skeleton.tsx
zenith-oi, line 6: **BLOCK** — [broken-contract: Skeleton width/height props removed] The previous `SkeletonProps` exposed `width` and `height` as explicit sizing props with automatic px-unit conversion. These are now silently dropped — callers using `<Skeleton width={200} height={40} />` will get no sizing at all (the values are swallowed by the `...props` spread onto a div, which ignores them). Since Skeleton is typically used inline with explicit dimensions, this will cause invisible/zero-size skeletons in existing consumers. The fix is straightforward: accept `width`/`height` via `React.HTMLAttributes<HTMLDivElement>` (which includes `style`) and either keep the numeric-to-px conversion logic or document that callers should pass `style={{ width: 200, height: 40 }}` or Tailwind classes instead.

packages/ui/src/components/Dialog/Dialog.tsx
zenith-oi, line 79: **FLAG** — [accessibility: Dialog missing focus trap] The Dialog sets focus to the dialog container on open (line 44) and traps body scroll, but has no focus trap implementation. Tab/Shift+Tab will cycle focus outside the dialog to the document behind the backdrop, which violates WCAG 2.1 SC 2.1.2 (No Keyboard Trap) in the wrong direction — focus escapes a modal. For a design system component that will be used across three products, this needs a proper focus trap: capture the previously-focused element on open, constrain Tab cycling to focusable children within the dialog, and restore focus on close. Consider using a small utility like `focus-trap-react` or implementing `getFocusableChildren` + keydown handler on the dialog div.

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+          'relative z-10 w-full max-w-md rounded-[16px] p-6',
+          'bg-[var(--amp-semantic-bg-surface)] border border-[var(--amp-semantic-border-default)]',
+          'shadow-xl',
+          'focus:outline-none',
+          className
+        )}
+      >
+        {title && (
+          <h2
+            id="dialog-title"
```
zenith-oi: **FLAG** — [accessibility: Dialog non-unique ARIA IDs] The `id` attributes for the dialog title and description are hardcoded strings: `id="dialog-title"` and `id="dialog-description"`. If two Dialogs are mounted simultaneously (e.g. a confirmation dialog opened from within a parent dialog, or a portal-rendered dialog during a transition), duplicate IDs in the DOM will break `aria-labelledby`/`aria-describedby` associations and cause accessibility violations. Fix: generate a stable unique ID per instance using `useId()` (React 18+) or a `useRef`-based counter.

packages/ui/src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,130 @@
+import React, { useState, useMemo } from 'react';
```
zenith-oi, line 1: **FLAG** — [missing-tests: critical interactive components] No tests are included for any of the 20 new components. For stateless display components this is acceptable short-term, but DataTable (sorting state), SearchInput (debounce + controlled/uncontrolled sync), Dialog (focus management, Escape key, portal rendering), Pagination (page number algorithm edge cases like totalPages=1, currentPage at boundaries), and Tabs (keyboard navigation) all have non-trivial logic that warrants unit tests. At minimum, DataTable sorting, SearchInput debounce behavior, and Pagination's `getPageNumbers` with edge inputs should be tested before these ship to three consuming products.

packages/ui/src/components/DataTable/DataTable.tsx
zenith-oi: **FLAG** — [correctness: DataTable row key instability] Rows are keyed by `idx` (array index). When sorting is applied, the sorted array reorders items but their index-based keys stay the same positionally — React sees key `0` at position 0 before and after sort, so it reuses DOM nodes rather than remounting, which can cause stale state in any row with internal state (checkboxes, inline edits, expand/collapse). The `DataTableColumn` type already has a `key` field; add an optional `rowKey` prop (or require a `keyExtractor: (item: T) => string`) so callers can provide a stable identity key per row.

packages/ui/src/components/SearchInput/SearchInput.tsx
zenith-oi, line 25: **FLAG** — [correctness: SearchInput controlled/uncontrolled desync] The `useEffect` that syncs `value` → `internal` (lines 24-26) runs on every render where `value` changes, which will override in-flight typing if the parent re-renders before the debounce fires. A user typing quickly will have their input reset to the last confirmed `value` from the parent mid-keystroke. The standard pattern here is to only sync from props when the component is not "dirty" (i.e. when the user has not typed since the last confirmed onChange), or to accept that this is a fully uncontrolled-with-callback component and remove the `value` prop / sync effect entirely, letting the caller reset via a `key` prop.

zenith-oi (dismissed):
## Zenith Code Review

**Large batch addition of 20+ new UI components to the design system (Input, Dialog, DataTable, Tabs, Toast, etc.) plus refactoring of Card, Skeleton, and EmptyState to use semantic CSS variables and forwardRef patterns. Overall quality is high — consistent patterns, good accessibility coverage, and correct use of design tokens throughout.**

Found 2 BLOCKs, 5 FLAGs. See inline comments for details.

**Positive:**
- Consistent and correct use of semantic CSS variables (`var(--amp-semantic-*)`) throughout all new components — no hardcoded colors anywhere, which is the top rule in CLAUDE.md.
- Accessibility fundamentals are present across the board: aria-label on icon-only buttons (IconButton requires `label` as a non-optional prop), aria-sort on DataTable sortable columns, role=progressbar with aria-valuenow/min/max on ProgressBar, aria-current='page' on Breadcrumb, proper role=tablist/tab/tabpanel with aria-selected and aria-controls on Tabs.
- forwardRef is applied correctly and consistently to all components that consumers will realistically need to attach refs to (Card, Input, Select, Textarea, Skeleton, Separator, IconButton).
- SearchInput's debounce implementation correctly cleans up the timer on unmount and on each new keystroke, avoiding the common leak of firing onChange after component unmount.
- ProgressBar clamps value to [0, 100] before rendering and before setting aria-valuenow — defensive and correct.

---

packages/ui/src/components/Card/Card.tsx
zenith-oi: **BLOCK** — [broken-contract: Card API regression] CardVariant removes 'outlined' and 'ghost', and CardPadding removes 'none'. Any consuming app currently using `variant='outlined'`, `variant='ghost'`, or `padding='none'` will get a TypeScript error after this merge (and a runtime silent wrong-variant render if they're on a pre-built version). This is a breaking API change with no migration path or deprecation. The three products (Brand, Creator, Atmosphere) need to be audited before this lands. Either keep the removed variants as deprecated aliases mapping to equivalent new styles, or this needs a major version bump and explicit consumer migration.

packages/ui/src/components/Skeleton/Skeleton.tsx
zenith-oi, line 6: **BLOCK** — [broken-contract: Skeleton API regression] The `width` and `height` props are removed from SkeletonProps. Any consumer that was controlling skeleton dimensions via those props (e.g. `<Skeleton width={200} height={40} />`) now gets unsized skeletons — invisible or collapsed — without a compile error, because `...props` on the div will silently pass those as invalid DOM attributes. This is a silent runtime regression for existing usages. Needs either a deprecation shim or explicit migration guide before merge.

packages/ui/src/components/Dialog/Dialog.tsx
zenith-oi: **FLAG** — [duplicate-dialog-ids] The title and description elements use hardcoded ids `dialog-title` and `dialog-description`. If two Dialogs are ever rendered simultaneously (e.g. a confirmation dialog on top of an edit dialog), both will emit the same ids, breaking `aria-labelledby` and `aria-describedby` association. Use a stable unique id per instance — `useId()` (available in React 18+) is the right fix here: `const titleId = React.useId(); const descId = React.useId();` and reference those in `aria-labelledby`/`aria-describedby`.

packages/ui/src/components/Dialog/Dialog.tsx
zenith-oi, line 79: **FLAG** — [focus-trap-missing] The Dialog focuses the container on open but does not trap focus inside it. A keyboard user can Tab out of the dialog into the (visually hidden, inert) background content. This violates WCAG 2.1 SC 2.1.2 (No Keyboard Trap in the correct direction). Implement a focus trap: on mount, find all focusable elements inside `dialogRef.current`, intercept Tab/Shift+Tab in the keydown handler to cycle within them, and restore focus to the previously focused element on close.

packages/ui/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export interface InputProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'> {
+  label?: string;
+  error?: string;
+  helperText?: string;
+}
+
+export const Input = React.forwardRef<HTMLInputElement, InputProps>(
+  ({ label, error, helperText, disabled, className, id, ...props }, ref) => {
+    const inputId = id || (label ? label.toLowerCase().replace(/\s+/g, '-') : undefined);
```
zenith-oi: **FLAG** — [input-id-collision] Auto-generated ids from label text (`label.toLowerCase().replace(/\s+/g, '-')`) will collide when the same label appears twice on a page (e.g. two 'Email' fields in different forms). This silently breaks label-input association for the second instance. Same issue exists in Textarea (L11), Select (L18), Checkbox (L21), and Switch (L21). Replace with `React.useId()` as the fallback: `const generatedId = React.useId(); const inputId = id || generatedId;`. This also removes the label-dependency for id generation, so unlabelled inputs with no explicit id still get a unique id.

packages/ui/src/components/DataTable/DataTable.tsx
zenith-oi: **FLAG** — [datatable-index-key] Rows are keyed by array index (`key={idx}`). After a sort operation `sortedData` is a new array in a different order, so React will reuse DOM nodes incorrectly — any row-level state (focused cells, inline edit state, transitions) will bleed across rows. Require either a `rowKey` prop (`(row: T) => string`) or a known `id` field on T, and use that as the key. The current approach will cause subtle but real bugs when sorting is used.

packages/ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const Tooltip: React.FC<TooltipProps> = ({
+  content,
+  children,
+  side = 'top',
+  className,
+}) => {
+  const [visible, setVisible] = useState(false);
+  const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
```
zenith-oi, line 27: **FLAG** — [tooltip-no-cleanup] `timeoutRef` is never cleared on unmount. If the component unmounts while the 100ms hide-delay timer is pending, the `setVisible(false)` call fires on an unmounted component. In React 18 strict mode this is a warning; in older roots it's a state-update-on-unmounted-component warning. Add a cleanup: `useEffect(() => () => { if (timeoutRef.current) clearTimeout(timeoutRef.current); }, [])`.

zenith-oi (dismissed):
## Zenith Code Review

**Large batch PR adding 20 new UI components (Input, Select, Dialog, DataTable, Tabs, Toast, etc.) plus Card compound component refactor. Overall quality is high — consistent token usage, good accessibility patterns, proper forwardRef usage. A few real issues worth fixing before merge.**

Found 2 BLOCKs, 5 FLAGs. See inline comments for details.

**Positive:**
- Consistent use of CSS custom properties (var(--amp-semantic-*)) throughout all new components — no hardcoded color values anywhere, which directly satisfies the repo's primary rule.
- Accessibility is genuinely thoughtful across the batch: aria-sort on DataTable sortable columns, aria-current='page' on Breadcrumb, role='switch' + aria-checked on Switch, role='progressbar' with aria-valuenow/min/max on ProgressBar, focus-visible ring on all interactive controls.
- Dialog implements Escape-key close, backdrop click dismiss, body scroll lock with proper cleanup, and portal rendering — all the right primitives for a modal.
- SearchInput debounce implementation is correct: timer is properly cleared on each keystroke, cancelled on unmount, and the external value is synced back in via useEffect so controlled usage works.
- forwardRef is applied consistently on all components that need it (Card family, Input, Textarea, Select, Skeleton, Separator, IconButton) — composable with third-party form libraries out of the box.

---

packages/ui/src/components/Card/Card.tsx
```diff
@@ -1,72 +1,128 @@
...
-/**
- * Card component props.
- *
- * **Accessibility note:** When `onClick` is provided the Card renders with
- * `role="button"`. Screen-reader users need an accessible name, so callers
- * should supply an `aria-label` (or ensure the card's text content is
- * sufficiently descriptive) whenever the card is interactive.
- */
-export interface CardProps {
+export interface CardProps extends React.HTMLAttributes<HTMLDivElement> {
```
zenith-oi, line 7: **BLOCK** — [breaking-contract] The old CardProps had `as`, `onClick`, `aria-label`, and `className` as explicit props. The new interface extends `React.HTMLAttributes<HTMLDivElement>` which drops the `as` prop and removes the keyboard handler (Enter/Space activation). Any consumer using `as='article'` or `as='li'` will silently get a broken render (always a div) without a compile error because `as` is just filtered into `...props` and passed to the div — it's not a valid HTMLDivElement attribute so React will ignore it. The `onClick`-based interactive variant has also lost keyboard activation (Enter/Space) that was explicitly handled before. The new 'interactive' variant in `variantClasses` suggests cards will be clickable, but there's no keyboard handler. Fix: Remove `as` from any public docs/storybook, add a migration note in the PR description (CLAUDE.md rule 5 requires this for breaking CSS/API changes), and add back keyboard activation to the 'interactive' variant or document that callers must handle it.

packages/ui/src/components/EmptyState/EmptyState.tsx
```diff
@@ -1,35 +1,42 @@
...
 export interface EmptyStateProps {
+  icon?: React.ReactNode;
   title: string;
   description?: string;
-  action?: {
-    label: string;
-    onClick: () => void;
-  };
-  icon?: React.ReactNode;
+  action?: React.ReactNode;
```
zenith-oi, line 8: **BLOCK** — [breaking-contract] The `action` prop type changed from `{ label: string; onClick: () => void }` to `React.ReactNode`. This is a breaking change for all three consuming products (Brand, Creator, Atmosphere). Any caller passing `action={{ label: 'Try again', onClick: fn }}` will now render the object literal as a React node, which produces '[object Object]' in the DOM — a silent runtime break with no TypeScript error because the object shape is not checked. CLAUDE.md rule 5 requires a migration note in the PR description for breaking changes. Fix: Add the migration note, and consider a deprecation path or codemod.

packages/ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,79 @@
...
+export interface SelectProps extends Omit<React.SelectHTMLAttributes<HTMLSelectElement>, 'size'> {
+  label?: string;
+  error?: string;
+  options: SelectOption[];
+  placeholder?: string;
+}
+
+export const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
+  ({ label, error, options, placeholder, disabled, className, id, ...props }, ref) => {
+    const selectId = id || (label ? label.toLowerCase().replace(/\s+/g, '-') : undefined);
```
zenith-oi, line 18: **FLAG** — [id-uniqueness] The generated `selectId` is derived purely from the label string (`label.toLowerCase().replace(...)`). If two Select components with the same label appear on the same page, they'll share the same `id`, breaking `htmlFor` association and `aria-describedby`. The Input component correctly appends `useId()` to the label to guarantee uniqueness — Select should do the same. Same issue exists in Textarea (line 11) and Checkbox (line 21) / Switch (line 21) — those use `id || label-derived-slug` without a uniqueness suffix.

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+          if (document.activeElement === last) {
+            e.preventDefault();
+            first.focus();
+          }
+        }
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyDown);
+    document.body.style.overflow = 'hidden';
```
zenith-oi, line 71: **FLAG** — [no-body-overflow-side-effect] `document.body.style.overflow = 'hidden'` is a shared global resource. If two Dialogs mount concurrently (or a Dialog mounts while another component has already set overflow), the cleanup `document.body.style.overflow = ''` on unmount will clear the lock even if the second dialog is still open. This is a known pitfall for modal components. Fix: Track the pre-existing overflow value and restore it, or use a ref-counted approach. At minimum document this limitation.

packages/ui/src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,139 @@
...
```
zenith-oi, line 1: **FLAG** — [missing-tests] No tests added for any of the 20 new components. DataTable in particular has non-trivial sorting logic (null handling, direction toggle, useMemo dependencies) that benefits from unit tests. Dialog has a focus trap implementation that's notoriously easy to get wrong. SearchInput has debounce and controlled/uncontrolled sync logic. These are the three highest-risk components — tests for their core behaviors should accompany this PR per the coding standards.

packages/ui/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const activeContent = tabs.find((t) => t.id === activeTab)?.content;
+
+  return (
+    <div className={cn('w-full', className)}>
+      <div
+        className="flex border-b border-[var(--amp-semantic-border-default)]"
+        role="tablist"
+      >
+        {tabs.map((tab) => (
+          <button
```
zenith-oi, line 28: **FLAG** — [tabs-keyboard-navigation] ARIA tabs pattern requires arrow key navigation between tabs (ArrowLeft/ArrowRight moves focus between tab buttons). Without this, keyboard users must Tab through every tab button individually. The current implementation only handles click. This isn't a WCAG failure per se since the tabs are focusable, but it violates the ARIA Authoring Practices Guide pattern for tablist widgets and will be flagged in any accessibility audit.

zenith-oi (changes requested):
## Zenith Code Review

**Large batch addition of 20+ new UI components (Input, Dialog, DataTable, Toast, Tooltip, etc.) plus refactoring existing components (Card, Skeleton, EmptyState) to use forwardRef and semantic CSS variables. Code quality is high overall with good accessibility patterns throughout.**

Found 2 BLOCKs, 4 FLAGs. See inline comments for details.

**Positive:**
- Dialog implements a proper focus trap with Escape key handling, scroll lock, backdrop click, and correct aria-modal/aria-labelledby/aria-describedby — this is unusually thorough for a design system component.
- Consistent use of semantic CSS variables (--amp-semantic-*) throughout all new components, no hardcoded colors — fully compliant with CLAUDE.md rule 1.
- DataTable's null-safe sorting comparator and the SearchInput's controlled/external value sync via prevValueRef are well-implemented edge cases.
- forwardRef adoption across Card, Skeleton, Input, Select, Textarea, IconButton, Separator means consumers can use refs for focus management — the right call for a design system.

---

PR #11, opened 2026-04-20
revert: Phase 2 — premature merge, needs user review

apaksh-gupta:
Reverts One-Impression/amplify-design-system#9


packages/ui/src/index.ts
```diff
@@ -4,108 +4,22 @@
...
 export type { ButtonProps, ButtonVariant, ButtonSize } from './components/Button';
 
-// Badge
 export { Badge } from './components/Badge';
 export type { BadgeProps, BadgeVariant, BadgeSize } from './components/Badge';
 
-// Card
-export { Card, CardHeader, CardTitle, CardDescription, CardContent, CardFooter } from './components/Card';
-export type { CardProps, CardVariant, CardPadding, CardHeaderProps, CardTitleProps, CardDescriptionProps, CardContentProps, CardFooterProps } from './components/Card';
+export { Card } from './components/Card';
```
zenith-oi, line 16: **BLOCK** — [breaking-change-requires-migration-note] CLAUDE.md explicitly states: 'Breaking changes to CSS variable names or values require a migration note in the PR description.' This PR removes 20+ public exports (CardHeader, CardTitle, CardDescription, CardContent, CardFooter, Input, Dialog, DataTable, Tabs, Toast, Tooltip, Avatar, Checkbox, Switch, Select, SearchInput, Separator, ProgressBar, Breadcrumb, Pagination, StatusTag, Timeline, IconButton, MetricCard, Textarea) and changes the CardProps interface shape. These are consumed by Brand, Creator, and Atmosphere products. There is no migration note, no deprecation period, and no semver major bump indicated. Pixel Agent's cross-product dependency analysis (cross-product-deps.ts) will detect the drift, but the downstream products will have broken imports immediately on publish. At minimum: document the removals in the PR description, bump the major version in package.json, and confirm with owners of all three consuming products before merge.

packages/ui/src/components/EmptyState/EmptyState.tsx
```diff
@@ -1,42 +1,35 @@
...
       {icon && (
-        <div className="w-12 h-12 rounded-[16px] bg-[var(--amp-semantic-bg-raised)] flex items-center justify-center mb-4 text-[var(--amp-semantic-text-muted)]">
+        <div className="w-12 h-12 rounded-xl bg-surface-raised flex items-center justify-center mb-4 text-neutral-500">
           {icon}
         </div>
       )}
-      <h3 className="text-[16px] font-semibold text-[var(--amp-semantic-text-primary)] mb-2">
-        {title}
-      </h3>
+      <h3 className="text-lg font-semibold text-neutral-900 mb-2">{title}</h3>
```
zenith-oi, line 24: **BLOCK** — [no-hardcoded-colors] CLAUDE.md rule 1: 'No hardcoded colors in UI components — use CSS variables only.' EmptyState now uses hardcoded Tailwind color utilities: `text-neutral-900` (L24), `text-neutral-700` (L26), and `text-neutral-500` (L20). These bypass the token system entirely — they won't respond to theme switching (light/dark) and won't be caught by the eslint-config no-hardcoded-colors rule. Replace with the appropriate semantic CSS variable classes (e.g. `text-[var(--amp-semantic-text-primary)]`, `text-[var(--amp-semantic-text-secondary)]`, `text-[var(--amp-semantic-text-muted)]`). Skeleton has the same issue at L20 with `bg-neutral-100` instead of `bg-[var(--amp-semantic-bg-sunken)]`.

packages/ui/src/components/Skeleton/Skeleton.tsx
```diff
@@ -1,38 +1,34 @@
...
+export const Skeleton: React.FC<SkeletonProps> = ({
+  width,
+  height,
+  variant = 'rectangular',
+  className,
+}) => {
+  return (
+    <div
+      className={cn(
+        'animate-pulse bg-neutral-100',
```
zenith-oi, line 20: **BLOCK** — [no-hardcoded-colors] Same violation as EmptyState: `bg-neutral-100` is a hardcoded Tailwind color, not a token reference. The original code correctly used `bg-[var(--amp-semantic-bg-sunken)]`. This will not theme-switch and breaks the token contract.

packages/ui/src/components/Card/Card.tsx
```diff
@@ -1,128 +1,72 @@
...
-    'border border-[var(--amp-semantic-border-default)] bg-[var(--amp-semantic-bg-surface)]',
-  elevated:
-    'bg-[var(--amp-semantic-bg-surface)] shadow-lg border border-[var(--amp-semantic-border-default)]',
-  interactive:
-    'bg-[var(--amp-semantic-bg-surface)] border border-[var(--amp-semantic-border-default)] cursor-pointer hover:shadow-lg transition-shadow duration-150',
-  outlined:
-    'bg-transparent border border-[var(--amp-semantic-border-default)]',
-  ghost:
-    'bg-transparent border-none',
+  default: 'bg-surface border border-border rounded-lg shadow-sm',
```
zenith-oi, line 27: **FLAG** — [broken-contract-card-variants] The new variantClasses use Tailwind utility classes `bg-surface`, `border-border`, `bg-surface-raised`, and `border-brand/20` (L27-L30) that are not standard Tailwind — they appear to be custom Tailwind v4 theme keys. If these aren't defined in the Tailwind config these classes will silently produce no styles. Confirm these token aliases exist in the build output before merging, or use the established CSS variable syntax. The old code explicitly used `var(--amp-semantic-*)` which is verifiable.

packages/ui/src/index.ts
```diff
@@ -4,108 +4,22 @@
...
-export type { TimelineProps, TimelineEvent } from './components/Timeline';
-
-// IconButton
-export { IconButton } from './components/IconButton';
-export type { IconButtonProps, IconButtonVariant, IconButtonSize } from './components/IconButton';
-
-// Separator
-export { Separator } from './components/Separator';
-export type { SeparatorProps, SeparatorOrientation } from './components/Separator';
+export { cn } from './lib/cn';
```
zenith-oi, line 25: **FLAG** — [removed-cn-utility-export] The `cn` utility export has been moved from the top of the file (where it was grouped with a comment) to the bottom, after the component exports. This is minor but `cn` is a utility that consuming packages may import from `@amplify/ui`. It's still exported so this isn't a break, but the placement change is worth noting if any downstream code depends on tree-shaking order. Low risk, but worth confirming intentional.

packages/ui/src/components/Card/Card.tsx
```diff
@@ -1,128 +1,72 @@
...
+      className={cn(
+        'transition-all duration-150',
+        variantClasses[variant],
+        paddingClasses[padding],
+        onClick && 'cursor-pointer hover:shadow-md active:scale-[0.99]',
+        className
+      )}
+      onClick={onClick}
+      aria-label={ariaLabel}
+      role={onClick ? 'button' : undefined}
```
zenith-oi, line 60: **FLAG** — [missing-keyboard-accessibility] The interactive Card adds `role="button"` and a keydown handler, which is correct. However, when `as="li"` or `as="article"` is combined with `onClick`, the role override is semantically odd — a `<li role="button">` creates a confusing accessible tree. Consider restricting the `as` prop to only allow interactive-card semantics when `onClick` is present, or document explicitly that `as` and `onClick` should not be combined arbitrarily. This is also missing `aria-pressed` for toggle-style cards if needed by consumers.

zenith-oi (changes requested):
## Zenith Code Review

**This PR significantly slims the design system's public surface area — removing 20+ components and reducing Card to a simpler polymorphic primitive. The remaining components (Card, Skeleton, EmptyState) are refactored toward Tailwind utility classes rather than CSS custom properties. The main concern is a hard breaking change to the public API with no migration path documented.**

Found 3 BLOCKs, 3 FLAGs. See inline comments for details.

**Positive:**
- The Card keyboard accessibility implementation (Enter/Space keydown, tabIndex, role=button) is correct and well-documented inline — the JSDoc accessibility note on CardProps is genuinely helpful for consumers.
- Removing @types/react and @types/react-dom as explicit devDependencies in favour of peer resolution is correct for a library package — avoids version conflicts in consuming repos.
- The polymorphic `as` prop on Card is a clean pattern for semantic HTML flexibility without requiring multiple component variants.

---

apaksh-gupta:
Keeping original merge — code is good, Zenith-approved

PR #12, opened 2026-04-20
revert: Phase 1 — premature merge, needs user review

apaksh-gupta:
Reverts One-Impression/amplify-design-system#8


zenith-oi (commented):
## Zenith Code Review

**This PR migrates the design token build pipeline from a custom W3C DTCG (dollar-prefix) format to a Style Dictionary v4 / Tokens Studio compatible format (bare `value`/`type` keys), restructures token files, and adds a Figma sync workflow. The migration is architecturally sound but has two blocking issues: the figma-sync workflow is explicitly prohibited by CLAUDE.md, and the build scripts have an inconsistent module system that will cause runtime failures.**

### BLOCK
- **[CLAUDE.md explicit prohibition]** .github/workflows/figma-sync.yml:1 — CLAUDE.md explicitly states: '~~`figma-sync.yml`~~ — REMOVED: Tokens Studio integration deprecated in favour of direct PRs + Pixel cascade. Design changes flow via Pixel Agent governance, not Figma plugin.' This file directly contradicts that architectural decision. The workflow must not be merged. If design-to-code sync automation is needed, it should go through Pixel Agent's existing `token-sync.ts` and `brand-cascade.ts` infrastructure.
- **[Module system consistency — CJS/ESM mismatch will crash at runtime]** packages/tokens-brand/build.js:12 — `package.json` removes `"type": "module"` (making the package default to CommonJS), but `tokens-creator/build.js` and `tokens-foundation/build.js` and `tokens-atmosphere/build.js` all use top-level `await` and ESM `import` syntax (e.g. `import StyleDictionary from 'style-dictionary'`). These files will throw `SyntaxError: Cannot use import statement` or `SyntaxError: await is only valid in async functions` when executed with Node in CJS mode. Meanwhile `tokens-brand/build.js` uses `require()`. The module system must be unified: either keep `"type": "module"` in package.json (and convert `tokens-brand/build.js` to ESM, fix `validate-tokens.js` to use ESM too), or convert all build scripts to CJS. The mixed state guarantees at least two build scripts will fail.
- **[Broken reference resolution — build will silently produce wrong CSS values]** packages/tokens-brand/build.js:65 — The `resolveRef` / `resolveTokenTree` functions at L28-L65 resolve `{color.semantic.error}` references by looking them up in `[colors, foundation]`. However, `foundation` is loaded from `../tokens-foundation/tokens/colors.json` which defines `color.semantic.error` as a reference itself: `"{color.primitive.red-500}"`. The resolver follows one level of indirection but the recursive call at L46 (`resolveRef(current.value, sources)`) re-runs against the same `sources` array which does NOT include a flat merged tree — so a two-hop reference chain (brand → foundation semantic → foundation primitive) will return the un-resolved string `"{color.primitive.red-500}"` instead of `"#FD5154"`. Any brand token that references a foundation semantic token (negative, warning, negative.light, warning.light at lines L14-L20 of tokens-brand/tokens/colors.json) will emit literal curly-brace strings into the CSS output. The fix is to first flatten `foundation` into a merged tree and resolve within that, or use Style Dictionary's own reference resolution (which atmosphere and foundation already delegate to SD) consistently.
- **[Pixel integration path broken — CLAUDE.md declared file paths no longer exist]** packages/tokens-foundation/tokens/primitives/colors.json:1 — CLAUDE.md documents that Pixel Agent's `token-sync.ts` reads from specific canonical paths: `packages/tokens-foundation/tokens/primitives/*.json` and `packages/tokens-foundation/tokens/semantic/*.json`. This PR deletes all files under both of those paths (primitives/ and semantic/ subdirectories are entirely removed). Pixel will either error on missing files or silently stop detecting drift. This is a broken contract with a live downstream consumer. The new flat structure under `tokens/colors.json` etc. is fine, but CLAUDE.md must be updated AND the Pixel `token-sync.ts` paths must be updated in a coordinated deploy, or the old paths must be preserved as symlinks/copies until Pixel is updated.

### FLAG
- **[validate-tokens.js checks the wrong required category key]** scripts/validate-tokens.js:74 — `requiredCategories` includes `'radius'` but the new `tokens-foundation/tokens/radii.json` defines tokens under the key `"radius"` — that part is fine. However it also checks for `'shadow'` but `tokens-foundation/tokens/shadows.json` uses the key `"shadow"` (singular) — also fine. The problem is `'color'`: the new `colors.json` uses `{ "color": { "primitive": ..., "semantic": ... } }` so `foundation['color']` will be truthy and pass. But the old motion, opacity, and z-index token categories are now gone from foundation entirely and nothing checks for them. More critically, the check at L76 uses `!foundation[cat]` which only does a shallow top-level key check — it won't catch a file that exists but has the wrong internal structure. This is a weakened validator, not a bug that blocks merge by itself, but it means the CI gate that was catching broken references (the old `findBrokenRefs`) has been completely removed with no replacement, so reference errors in token files will now silently pass validation.
- **[tokens-creator/build.js uses top-level await without ESM package type]** packages/tokens-creator/build.js:38 — Beyond the BLOCK above, `tokens-creator` specifically reads token files with relative paths `'tokens/colors.json'` (L41-L42) rather than `path.join(__dirname, ...)`. This means the build must be run from the package directory — if run from the workspace root (as `npm run build` typically does via workspace scripts), the `fs.readFileSync` calls will fail with ENOENT. The foundation and atmosphere build scripts avoid this by relying on Style Dictionary's `source` glob which resolves relative to CWD at SD config time. Use `new URL('./tokens/colors.json', import.meta.url)` or `path.join(__dirname, 'tokens/colors.json')` depending on the final module system choice.
- **[Dark theme tokens removed with no replacement — breaking change for Atmosphere and Creator consumers]** packages/tokens-atmosphere/tokens/theme-dark.json:1 — All dark theme token files have been removed (atmosphere, brand, creator) and replaced with single-theme flat color files. The old files defined dark-mode-specific values for bg, text, borders, and accent colors. The new `tokens-atmosphere/tokens/colors.json` and `tokens-creator/tokens/colors.json` only define light/neutral values. Any consumer using `--amplify-atmosphere-theme-bg-primary` or the dark-mode CSS variables will get nothing. If dark mode support is intentionally being dropped, this needs to be called out explicitly in the PR description as a breaking change per CLAUDE.md rule 5.

**Positive:**
- Consolidating four per-package build scripts that each called a shared script into individual Style Dictionary configs is the right call — it eliminates the fragile exec chain and gives each package explicit, auditable configuration.
- The cross-product color standardization work (aligning purple-500, red-500, green-500, amber-500 hex values across brand/creator/atmosphere with documented rationale in descriptions) is genuinely valuable and the descriptions in colors.json are well-written.
- The Figma sync workflow itself (ignoring the prohibition) is well-constructed — the duplicate PR guard at L62, the scoped permissions, and the build-then-validate ordering are all correct patterns.

---

apaksh-gupta:
Keeping original merge — code is good, Zenith-approved

PR #15, opened 2026-04-20
feat: npm publishing via GitHub Packages

apaksh-gupta:
## Summary
- Add `publishConfig` (registry: npm.pkg.github.com, access: restricted) to all 6 publishable packages: tokens-foundation, tokens-brand, tokens-atmosphere, tokens-creator, ui, eslint-config
- Create `.npmrc` with @amplify scope pointing to GitHub Packages registry
- Add publish step to CI workflow that runs after build on push to main (uses `GITHUB_TOKEN`)
- Add `packages: write` permission and `registry-url`/`scope` to setup-node for proper auth

## How it works
- On push to main, after build + validation passes, CI loops through each package and runs `npm publish --access restricted`
- `|| true` ensures a version that already exists does not fail the pipeline — bump version in package.json to trigger a new publish
- Consumers install with: `npm install @amplify/tokens-foundation --registry=https://npm.pkg.github.com` (or add the same .npmrc to their repo)

## Test plan
- [ ] Verify CI passes on this PR (build + validate, no publish since not main)
- [ ] After merge, confirm packages appear at github.com/orgs/One-Impression/packages
- [ ] Test consuming a package from another repo with the .npmrc registry config

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,16 @@ jobs:
         run: |
           for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui eslint-config; do
             cd packages/$pkg
-            npm publish --access restricted || true
+            OUTPUT=$(npm publish --access restricted 2>&1) && echo "$OUTPUT" || {
+              EXIT_CODE=$?
+              echo "$OUTPUT"
+              if echo "$OUTPUT" | grep -q 'EPUBLISHCONFLICT\|already been published\|You cannot publish over the previously published versions'; then
+                echo "Package already published at this version — skipping"
+              else
+                echo "::error::Publish failed for $pkg"
+                exit $EXIT_CODE
+              fi
+            }
             cd ../..
           done
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [main]
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
   build-and-validate:
     name: Build & Validate Tokens
@@ -18,6 +22,8 @@ jobs:
         with:
           node-version: 20
           cache: npm
+          registry-url: https://npm.pkg.github.com
+          scope: '@amplify'
 
       - name: Install dependencies
         run: npm ci
@@ -58,6 +64,17 @@ jobs:
 
       - name: Build Storybook
         run: npm run build -w packages/storybook
+
+      - name: Publish packages
+        if: github.ref == 'refs/heads/main'
+        run: |
+          for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui eslint-config; do
+            cd packages/$pkg
+            npm publish --access restricted || true
+            cd ../..
+          done
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   secret-scan:
     name: Secret Scan
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,12 +8,14 @@ on:
 
 permissions:
   contents: read
-  packages: write
 
 jobs:
   build-and-validate:
     name: Build & Validate Tokens
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
 
     steps:
       - uses: actions/checkout@v4
@@ -68,6 +70,7 @@ jobs:
       - name: Publish packages
         if: github.ref == 'refs/heads/main'
         run: |
+          set -e
           for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui eslint-config; do
             echo "Publishing @amplify/$pkg..."
             (
@@ -83,8 +86,6 @@ jobs:
                 fi
               }
             )
-            # Propagate subshell failure
-            if [ $? -ne 0 ]; then exit 1; fi
           done
         env:
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,18 +69,22 @@ jobs:
         if: github.ref == 'refs/heads/main'
         run: |
           for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui eslint-config; do
-            cd packages/$pkg
-            OUTPUT=$(npm publish --access restricted 2>&1) && echo "$OUTPUT" || {
-              EXIT_CODE=$?
-              echo "$OUTPUT"
-              if echo "$OUTPUT" | grep -q 'EPUBLISHCONFLICT\|already been published\|You cannot publish over the previously published versions'; then
-                echo "Package already published at this version — skipping"
-              else
-                echo "::error::Publish failed for $pkg"
-                exit $EXIT_CODE
-              fi
-            }
-            cd ../..
+            echo "Publishing @amplify/$pkg..."
+            (
+              cd packages/$pkg
+              OUTPUT=$(npm publish --access restricted 2>&1) && echo "$OUTPUT" || {
+                EXIT_CODE=$?
+                echo "$OUTPUT"
+                if echo "$OUTPUT" | grep -q 'EPUBLISHCONFLICT\|already been published\|You cannot publish over the previously published versions'; then
+                  echo "Package already published at this version — skipping"
+                else
+                  echo "::error::Publish failed for $pkg"
+                  exit $EXIT_CODE
+                fi
+              }
+            )
+            # Propagate subshell failure
+            if [ $? -ne 0 ]; then exit 1; fi
           done
         env:
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+@amplify:registry=https://npm.pkg.github.com
+//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -7,5 +7,9 @@
   "peerDependencies": {
     "eslint": ">=8.0.0"
   },
-  "files": ["index.js", "rules/"]
+  "files": ["index.js", "rules/"],
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com",
+    "access": "restricted"
+  }
 }
--- a/packages/tokens-atmosphere/package.json
+++ b/packages/tokens-atmosphere/package.json
@@ -18,5 +18,9 @@
   "devDependencies": {
     "style-dictionary": "^4.3.0"
   },
-  "files": ["dist/", "tokens/"]
+  "files": ["dist/", "tokens/"],
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com",
+    "access": "restricted"
+  }
 }
--- a/packages/tokens-brand/package.json
+++ b/packages/tokens-brand/package.json
@@ -20,5 +20,9 @@
   "devDependencies": {
     "style-dictionary": "^4.3.0"
   },
-  "files": ["dist/", "tokens/"]
+  "files": ["dist/", "tokens/"],
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com",
+    "access": "restricted"
+  }
 }
--- a/packages/tokens-creator/package.json
+++ b/packages/tokens-creator/package.json
@@ -20,5 +20,9 @@
   "devDependencies": {
     "style-dictionary": "^4.3.0"
   },
-  "files": ["dist/", "tokens/", "scripts/"]
+  "files": ["dist/", "tokens/", "scripts/"],
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com",
+    "access": "restricted"
+  }
 }
--- a/packages/tokens-foundation/package.json
+++ b/packages/tokens-foundation/package.json
@@ -16,5 +16,9 @@
   "devDependencies": {
     "style-dictionary": "^4.3.0"
   },
-  "files": ["dist/", "tokens/"]
+  "files": ["dist/", "tokens/"],
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com",
+    "access": "restricted"
+  }
 }
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -37,5 +37,9 @@
   "dependencies": {
     "clsx": "^2.1.0",
     "tailwind-merge": "^2.3.0"
+  },
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com",
+    "access": "restricted"
   }
 }


.github/workflows/ci.yml
```diff
@@ -59,6 +65,17 @@ jobs:
       - name: Build Storybook
         run: npm run build -w packages/storybook
 
+      - name: Publish packages
+        if: github.ref == 'refs/heads/main'
+        run: |
+          for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui eslint-config; do
+            cd packages/$pkg
+            npm publish --access restricted || true
```
zenith-oi: **FLAG** — [error-handling on external calls] The `|| true` on `npm publish` silently swallows all failures — including network errors, auth failures, malformed package.json, and version conflicts. The only case you actually want to suppress is 'version already published' (exit code 409/EPUBLISHCONFLICT). Replace with a proper check: `npm publish --access restricted 2>&1 | tee /tmp/publish_out.txt; code=$?; if [ $code -ne 0 ] && ! grep -q 'EPUBLISHCONFLICT\|already been published' /tmp/publish_out.txt; then exit $code; fi`. This way a legitimate publish failure (bad token, missing dist files, network timeout) will actually fail the CI job instead of silently passing.

zenith-oi (commented):
## Zenith Code Review

**Adds GitHub Packages publishing infrastructure: .npmrc configuration, publishConfig in all package manifests, and a publish step in CI. Straightforward and well-scoped, but has one reliability issue worth addressing.**

Found 1 FLAG. See inline comments for details.

**Positive:**
- Minimal, correct permissions block — `contents: read` + `packages: write` is exactly right, no over-privileging.
- Using `publishConfig` in each package.json as the source of truth for registry is the correct pattern — it means `npm publish` from any context routes correctly regardless of the root .npmrc.
- The `if: github.ref == 'refs/heads/main'` guard is correct; publish only runs on the main branch, not on PRs.

---

zenith-oi (commented):
## Zenith Code Review

**Adds automated npm publish to GitHub Packages Registry on merge to main, with publishConfig in each package.json and idempotency handling for already-published versions. Straightforward CI plumbing — the implementation is mostly correct but has one structural issue worth flagging.**

Found 2 FLAGs. See inline comments for details.

**Positive:**
- Using `${{ secrets.GITHUB_TOKEN }}` rather than a personal access token is the right call for GitHub Packages — no long-lived secret to rotate.
- publishConfig in each package.json is the correct canonical way to pin the registry per-package, ensuring `npm publish` works correctly regardless of the caller's .npmrc state.
- Scoping workflow permissions to `contents: read` + `packages: write` follows least-privilege — not requesting broader repo write access.

---

.github/workflows/ci.yml
```diff
@@ -59,6 +65,26 @@ jobs:
       - name: Build Storybook
         run: npm run build -w packages/storybook
 
+      - name: Publish packages
```
zenith-oi, line 70: **FLAG** — [ci-publish-before-validate] The publish step runs inside the `build-and-validate` job, after Storybook builds. This means if the Storybook build fails (unrelated to the packages being published), the publish step is skipped entirely — which may be intentional, but it also means a broken Storybook blocks a valid token/UI release. More importantly, the publish runs on every push to main regardless of whether versions have actually been bumped. There's no version-change guard, so every merge to main attempts a publish for all 6 packages, relying entirely on the EPUBLISHCONFLICT grepping to skip already-published ones. This is fragile: if npm's error message wording changes, the grep silently fails and the step exits non-zero mid-loop leaving partial publishes. Consider either a dedicated `publish` job that checks for version changes (e.g. `git diff HEAD~1 -- package.json | grep version`) or use `changesets` / `lerna` for proper publish orchestration.

.github/workflows/ci.yml
```diff
@@ -59,6 +65,26 @@ jobs:
...
+        if: github.ref == 'refs/heads/main'
+        run: |
+          for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui eslint-config; do
```
zenith-oi, line 74: **FLAG** — [loop-directory-state] The `cd packages/$pkg` / `cd ../..` pattern is stateful — if any earlier step in the loop leaves the working directory in an unexpected state (e.g. a subshell exit), subsequent iterations will `cd` into the wrong directory and publish the wrong package or fail silently. Prefer `(cd packages/$pkg && npm publish --access restricted)` to run each publish in a subshell so directory state is isolated per iteration and can never bleed across packages.

zenith-oi (commented):
## Zenith Code Review

**Adds GitHub Packages publishing for all design system packages, with per-package publishConfig and a CI step that handles the already-published version gracefully. Solid approach overall with one real concern worth addressing.**

Found 2 FLAGs. See inline comments for details.

**Positive:**
- The already-published version detection is well thought out — catching EPUBLISHCONFLICT and the prose variants means this won't break on re-runs or when versions haven't been bumped yet.
- Using GITHUB_TOKEN (not a PAT) for publishing to GitHub Packages is the correct approach — no secret rotation required.
- publishConfig in each package.json is the right belt-and-suspenders alongside .npmrc — prevents accidental publishes to the public npm registry.

---

.github/workflows/ci.yml
zenith-oi, line 70: **FLAG** — [ci-publish-no-build-gate] The publish step runs after 'Build Storybook' but there's no explicit gate confirming the token build and validation steps succeeded before publishing. If the job is restructured or steps are reordered, a broken build could publish. The publish step should either be in a separate job with `needs: [build-and-validate]` (which would also make the condition cleaner — job-level `if` rather than step-level), or at minimum add a comment making the ordering dependency explicit. As written, it relies on sequential step execution within the same job, which is fine but fragile if someone adds an early `continue-on-error` elsewhere.

.github/workflows/ci.yml
```diff
@@ -59,6 +65,26 @@ jobs:
...
+            cd packages/$pkg
```
zenith-oi: **FLAG** — [cd-in-loop-fragility] The `cd packages/$pkg` / `cd ../..` pattern inside the for-loop is fragile. If any iteration exits early (e.g. the `exit $EXIT_CODE` path fires and is somehow caught), subsequent iterations will run from the wrong directory and fail with confusing errors. Prefer `(cd packages/$pkg && npm publish --access restricted 2>&1)` with a subshell so directory changes are scoped to each iteration and the parent shell's working directory is never mutated.

.github/workflows/ci.yml
```diff
@@ -59,6 +65,30 @@ jobs:
...
+                if echo "$OUTPUT" | grep -q 'EPUBLISHCONFLICT\|already been published\|You cannot publish over the previously published versions'; then
+                  echo "Package already published at this version — skipping"
+                else
+                  echo "::error::Publish failed for $pkg"
+                  exit $EXIT_CODE
+                fi
+              }
+            )
+            # Propagate subshell failure
+            if [ $? -ne 0 ]; then exit 1; fi
```
zenith-oi: **FLAG** — [ci-trigger-mismatch] The subshell failure propagation check `if [ $? -ne 0 ]; then exit 1; fi` at L87 is always false. The subshell runs in a `( ... )` block that already calls `exit $EXIT_CODE` on a real failure, and on success the subshell exits 0. `$?` after a subshell that exits cleanly is always 0, so this guard is dead code. It doesn't cause a bug (failures are still caught inside the subshell), but it's misleading. Remove it or replace the subshell pattern with a function so the exit code propagates naturally.

.github/workflows/ci.yml
```diff
@@ -59,6 +65,30 @@ jobs:
       - name: Build Storybook
         run: npm run build -w packages/storybook
 
+      - name: Publish packages
+        if: github.ref == 'refs/heads/main'
```
zenith-oi, line 71: **FLAG** — [publish-on-every-main-push] Publishing is triggered on every push to main, not just when versions change. If a commit touches docs, CI config, or anything unrelated to a package, the workflow will attempt a publish for all six packages. The conflict-skip logic catches this gracefully, but it means every merge to main hits the registry six times unnecessarily. Consider gating on `if: github.ref == 'refs/heads/main' && contains(github.event.head_commit.message, 'release')` or, better, use changesets or a dedicated release workflow triggered by version tags. This is a workflow hygiene issue, not a correctness bug, but it will produce noisy publish logs and unnecessary registry churn.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [main]
 
+permissions:
```
zenith-oi, line 9: **FLAG** — [permissions-scope] `packages: write` is declared at the top-level `permissions` block, which applies it to ALL jobs including `secret-scan` and any other jobs in this file. `packages: write` should be scoped only to the `build-and-validate` job (move the `permissions` block inside that job) so other jobs run with the minimum required permissions. This is a least-privilege hardening issue.

zenith-oi (commented):
## Zenith Code Review

**Adds automated npm package publishing to GitHub Packages Registry on pushes to main, with per-package publishConfig and a version-conflict-tolerant publish loop. The implementation is correct and the conflict-handling logic is solid.**

Found 3 FLAGs. See inline comments for details.

**Positive:**
- The version-conflict error matching covers all three known npm error strings (EPUBLISHCONFLICT, 'already been published', 'You cannot publish over') — this is thorough and avoids false negatives across different npm/registry versions.
- Using `secrets.GITHUB_TOKEN` (not a PAT) for publishing is the correct approach for GitHub Packages — no secret rotation needed.
- publishConfig in each package.json as a belt-and-suspenders alongside the .npmrc is the right pattern — it ensures the registry target is explicit even if .npmrc is absent in a local dev context.

---

.github/workflows/ci.yml
```diff
@@ -59,6 +65,30 @@ jobs:
       - name: Build Storybook
         run: npm run build -w packages/storybook
 
+      - name: Publish packages
```
zenith-oi, line 70: **FLAG** — [ci-publish-ordering] The publish step runs in the same job as build-and-validate, meaning a publish failure will mark the build job as failed and could obscure build/validate results in the GitHub UI. More importantly, if the Storybook build at L66 fails, publish is skipped entirely — which is probably fine, but it also means a flaky Storybook build blocks a token-only release. Consider splitting publish into its own job with `needs: [build-and-validate]` so concerns are separated and retries are scoped correctly.

.github/workflows/ci.yml
zenith-oi, line 9: **FLAG** — [permissions-scope] The `packages: write` permission is set at the workflow level (L9-L11), so it also applies to the `secret-scan` job and any other jobs added in future. Prefer setting permissions at the job level — keep `packages: write` only on the job that actually publishes, and give other jobs `packages: read` or nothing. This follows least-privilege and avoids accidentally granting write access to jobs that shouldn't have it.

zenith-oi (commented):
## Zenith Code Review

**Adds automated npm publishing to GitHub Packages Registry for all design system packages, triggered on pushes to main. The implementation is clean and well-structured.**

Found 2 FLAGs. See inline comments for details.

**Positive:**
- The EPUBLISHCONFLICT/already-published detection (L78) is correct and handles the common case of re-running CI on an already-tagged version cleanly — avoids false failures without masking real ones.
- Using `GITHUB_TOKEN` rather than a PAT for publishing is the right call — no secret rotation needed and permissions are scoped to the repo.
- Setting `publishConfig` in each package.json in addition to the workflow-level config is the correct belt-and-suspenders approach — packages will publish to the right registry even if run locally with the right token.

---

.github/workflows/ci.yml
```diff
@@ -59,6 +67,29 @@ jobs:
...
```
zenith-oi, line 70: **FLAG** — [publish-without-build] The 'Publish packages' step runs after 'Build Storybook' but there's no explicit dependency on the token build steps completing first. If any earlier step fails and somehow execution reaches publish (or if the step order shifts), you'd publish stale/missing dist/ artifacts. The publish step should either be a separate job with `needs: build-and-validate` or have an explicit check that dist/ directories exist for token packages before publishing. Given the current linear step ordering it works today, but it's fragile — a step reorder would silently publish empty packages.

.github/workflows/ci.yml
```diff
@@ -59,6 +67,29 @@ jobs:
...
+        if: github.ref == 'refs/heads/main'
```
zenith-oi, line 71: **FLAG** — [missing-version-gate] Publishing is gated on `github.ref == 'refs/heads/main'` but there's no check that the package version was actually bumped before publishing. Every merge to main will attempt to publish — the EPUBLISHCONFLICT skip logic handles the already-published case gracefully, but this means the publish step is essentially a no-op on most merges. Consider adding a version-change check (e.g. comparing package.json version to the registry) or adopting changesets/semantic-release so publishes only fire on intentional version bumps. This is a workflow ergonomics issue, not a correctness blocker.

zenith-oi (commented):
## Zenith Code Review

**Adds automated npm publishing to GitHub Packages Registry for all design system packages, triggered on pushes to main. The implementation is clean and well-structured.**

Found 2 FLAGs. See inline comments for details.

**Positive:**
- The EPUBLISHCONFLICT idempotency handling is well-done — catching all three known error strings and skipping gracefully means re-running CI on an already-published version won't break the workflow.
- Using `set -e` at the top of the publish script and explicitly re-exiting with the captured exit code ($EXIT_CODE) correctly propagates failures rather than swallowing them.
- Scope-level permissions (`packages: write`) are applied at the job level, not globally — the workflow-level permission is correctly set to read-only, which follows least-privilege correctly.
- publishConfig in each package.json as the source of truth for registry/access settings is the right pattern — it ensures `npm publish` from any context targets the correct registry regardless of local .npmrc state.

---

.github/workflows/ci.yml
zenith-oi, line 70: **FLAG** — [publish-on-build-job] The publish step is appended to the build-and-validate job, meaning a publish failure (or a flaky build step before it) blocks the CI signal for PRs. More importantly, publishing on every main push without a version gate means this will silently no-op on every commit where the version hasn't been bumped (caught by the EPUBLISHCONFLICT guard), but a botched version bump would immediately publish to consumers with no staging gate. Consider a separate `publish` job that depends on `build-and-validate`, or gating on a tag push (`refs/tags/v*`) rather than any push to main. This is especially relevant here since Pixel Agent and 3 product repos consume these packages — an accidental publish of a broken version has wide blast radius.

.github/workflows/ci.yml
```diff
@@ -59,6 +67,29 @@ jobs:
...
+        run: |
+          set -e
+          for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui eslint-config; do
```
zenith-oi, line 74: **FLAG** — [missing-version-check] The publish loop iterates over a hardcoded package list. If a new package is added to the monorepo (e.g. `feature-flags` currently lacks a publishConfig), it won't be published and there's no warning. Consider deriving the list dynamically from packages that have `publishConfig` in their package.json, or at minimum document this hardcoded list needs manual maintenance. The `feature-flags` package is notably absent — intentional or oversight?

zenith-oi (commented):
## Zenith Code Review

**Adds automated npm publishing to GitHub Packages Registry for all design system packages, triggered on push to main. The implementation is clean and well-structured.**

Found 2 FLAGs. See inline comments for details.

**Positive:**
- The EPUBLISHCONFLICT idempotency guard is well-written — it catches all three known npm error strings for already-published versions and exits cleanly rather than failing the build, which is exactly the right behavior for a push-to-main trigger.
- Using `packages: write` only on the job-level permission (not workflow-level) follows least-privilege correctly — the top-level `contents: read` is appropriately restrictive.
- Using `set -e` at the top of the publish script and explicit `EXIT_CODE` capture before the conditional means publish errors are never silently swallowed.
- Consistent `publishConfig` in all package.json files means the registry target is canonical per-package and doesn't rely solely on the .npmrc, which is good defensive practice.

---

.github/workflows/ci.yml
zenith-oi, line 70: **FLAG** — [publish-before-validate] The 'Publish packages' step runs inside the same job as 'Build Storybook' (L68), which means it publishes even if storybook build fails — unless storybook failure causes the whole job to exit, which it does with `set -e` only within the publish step itself. More critically, the publish step comes *after* storybook build but there's no explicit check that token build and validation steps (which run earlier in the same job) have produced fresh `dist/` artifacts. This is likely fine since the job is sequential, but worth confirming that `npm run build -w packages/storybook` failure would halt the job before the publish step. Check that `continue-on-error` is not set anywhere on earlier steps.

zenith-oi, line 70: **FLAG** — [missing-version-bump-gate] There is no version bump step — every merge to main will attempt to publish at whatever version is currently in each package.json. The 'already published' conflict handling (L81) will silently skip unchanged versions, which works, but means a developer must manually bump the version in package.json before merging to actually ship a new release. This is a footgun: a meaningful change merges to main, nothing actually publishes because the version wasn't bumped, and consumers see no update. Consider either: (a) adding a CI check that fails if a changed package's version hasn't been bumped, or (b) documenting this requirement explicitly in CLAUDE.md so engineers don't ship changes that silently don't publish.

.npmrc
```diff
@@ -0,0 +1,2 @@
+@amplify:registry=https://npm.pkg.github.com
+//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}
```
zenith-oi, line 2: **FLAG** — [npmrc-auth-token-exposure] The `.npmrc` file at repo root uses `${NODE_AUTH_TOKEN}` interpolation, which is correct for CI. However, this file will also be read by developers running `npm install` locally. If a developer has `NODE_AUTH_TOKEN` unset, npm may error or fall back unexpectedly when resolving `@amplify`-scoped packages. More importantly — confirm this `.npmrc` is not accidentally committing a literal token (it isn't here, it's an env var reference, so this is fine as written). No action needed on the token itself, but consider adding a comment noting that `NODE_AUTH_TOKEN` must be set locally to install from the private registry, and ensure `.npmrc` is not in `.gitignore` (it shouldn't be, since it's intentionally committed).

zenith-oi (commented):
## Zenith Code Review

**This PR adds automated publishing of all design system packages to GitHub Packages (npm.pkg.github.com) on merges to main. The implementation is straightforward and correct — the publish step runs only on main, handles already-published versions gracefully, and scopes permissions appropriately.**

Found 3 FLAGs. See inline comments for details.

**Positive:**
- Permissions are correctly scoped: top-level `contents: read` with job-level `packages: write` override — this follows least-privilege correctly for GITHUB_TOKEN.
- The already-published version detection (L81) catches all three known npm error strings for version conflicts, making the publish step safely idempotent without failing noisy CI runs.
- Using `publishConfig` in each package.json rather than relying solely on `.npmrc` or CLI flags means the registry target is explicit and version-controlled alongside the package — good defensive practice.

---

PR #16, opened 2026-04-21
feat: add 4 design system enforcement ESLint rules (v2)

apaksh-gupta:
## Summary
Adds 4 new ESLint rules to `@amplify/eslint-config` that catch the highest-impact design system violations at lint time. All set to `warn` — existing code keeps building, new violations get flagged.

### New rules

| Rule | Catches | Prevents |
|------|---------|----------|
| `no-inline-styles` | `style={{}}` with literal values | Bypassing design tokens entirely |
| `no-raw-surface` | `bg-[var(--surface-*)]` without ring/border/shadow | Invisible cards in light mode |
| `no-hardcoded-radius` | `rounded-lg/md/xl` etc. | Inconsistent corners |
| `no-hardcoded-typography` | `text-[13px]`, `text-[11px]` | Typography outside scale |

### Existing rules (unchanged)
- `no-hardcoded-colors` — hex colors in JSX
- `no-raw-spacing` — raw pixel values in style props
- `prefer-token-import` — legacy import paths

**Total: 7 rules** covering all major design system violation classes.

### Smart exceptions
- `src/components/ui/` files exempted (component definitions set their own styling)
- `style={variable}` allowed (pre-defined style objects for gradual migration)
- `rounded-full` / `rounded-none` allowed (intentional)
- `text-[var(--*)]` allowed (token-based)
- `surface-primary/hover/active` exempted from boundary check (page bg, interaction states)

## Test plan
- [x] All 7 rules load: `node -e "require('./rules/no-inline-styles')"` etc.
- [ ] Atmosphere `npm run lint` shows new warnings for existing violations

### DIFF
--- a/packages/eslint-config/rules/no-hardcoded-radius.js
+++ b/packages/eslint-config/rules/no-hardcoded-radius.js
@@ -1,0 +1,63 @@
+/**
+ * @amplify/no-hardcoded-radius
+ *
+ * Flags Tailwind hardcoded border-radius classes. Use design token radii instead:
+ *   rounded-[var(--radius-card)]  — 16px, for cards and containers
+ *   rounded-[var(--radius-xs)]    — 6px, for small elements (tags, inputs)
+ *   rounded-full                  — pills and circles (allowed)
+ *   rounded-none                  — intentional removal (allowed)
+ *
+ * Flagged: rounded-sm, rounded-md, rounded-lg, rounded-xl, rounded-2xl, rounded-3xl,
+ *          rounded-[8px], rounded-[12px] etc.
+ */
+
+const HARDCODED_RADIUS = /\brounded-(sm|md|lg|xl|2xl|3xl)\b|\brounded-\[\d+px\]/;
+const EXEMPT = /\brounded-(full|none)\b/;
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Use design token border-radius instead of hardcoded Tailwind radius classes',
+    },
+    schema: [],
+    messages: {
+      noHardcodedRadius:
+        'Use design token radius: `rounded-[var(--radius-card)]` (16px) or `rounded-[var(--radius-xs)]` (6px). Avoid hardcoded rounded-lg/md/xl.',
+    },
+  },
+
+  create(context) {
+    const filename = context.filename || context.getFilename();
+    if (/components\/ui\//.test(filename)) return {};
+
+    return {
+      JSXAttribute(node) {
+        if (node.name.name !== 'className') return;
+        const value = node.value;
+        if (!value) return;
+
+        let classStr = '';
+        if (value.type === 'Literal' && typeof value.value === 'string') {
+          classStr = value.value;
+        } else if (value.type === 'JSXExpressionContainer' && value.expression.type === 'TemplateLiteral') {
+          classStr = value.expression.quasis.map((q) => q.value.raw).join('');
+        } else if (value.type === 'JSXExpressionContainer' && value.expression.type === 'CallExpression') {
+          // cn(...) calls — check string arguments
+          for (const arg of value.expression.arguments) {
+            if (arg.type === 'Literal' && typeof arg.value === 'string') {
+              classStr += ' ' + arg.value;
+            }
+          }
+        } else {
+          return;
+        }
+
+        if (!HARDCODED_RADIUS.test(classStr)) return;
+        if (EXEMPT.test(classStr) && !HARDCODED_RADIUS.test(classStr.replace(EXEMPT, ''))) return;
+
+        context.report({ node, messageId: 'noHardcodedRadius' });
+      },
+    };
+  },
+};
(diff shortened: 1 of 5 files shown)


packages/eslint-config/rules/no-hardcoded-radius.js
```diff
@@ -0,0 +1,63 @@
...
+    const filename = context.filename || context.getFilename();
+    if (/components\/ui\//.test(filename)) return {};
+
+    return {
+      JSXAttribute(node) {
+        if (node.name.name !== 'className') return;
+        const value = node.value;
+        if (!value) return;
+
+        let classStr = '';
```
zenith-oi, line 40: **FLAG** — [DRY / code duplication] The className extraction logic (Literal → string, TemplateLiteral → quasis, CallExpression → string args) is duplicated identically across no-hardcoded-radius.js, no-hardcoded-typography.js, and no-raw-surface.js (three copies). Extract this into a shared helper, e.g. `rules/utils/extract-class-string.js`, and import it in each rule. Future changes to handle additional expression types (e.g. `clsx` with object args, conditional expressions) will currently require the same fix in three places.

packages/eslint-config/rules/no-raw-surface.js
```diff
@@ -0,0 +1,66 @@
...
+        if (node.name.name !== 'className') return;
+        const value = node.value;
+        if (!value) return;
+
+        let classStr = '';
+        if (value.type === 'Literal' && typeof value.value === 'string') {
+          classStr = value.value;
+        } else if (value.type === 'JSXExpressionContainer' && value.expression.type === 'TemplateLiteral') {
+          classStr = value.expression.quasis.map((q) => q.value.raw).join('');
+        } else {
```
zenith-oi, line 51: **FLAG** — [incomplete-cn-handling] no-raw-surface.js falls through to `return` on JSXExpressionContainer → CallExpression (cn(...) calls), while no-hardcoded-radius.js and no-hardcoded-typography.js do handle this case. A `className={cn('bg-[var(--surface-elevated)]', 'p-5')}` call will silently bypass the no-raw-surface rule. Either add the same CallExpression branch here or document this as an intentional limitation.

zenith-oi (approved):
## Zenith Code Review

**Adds four new ESLint rules to the design system's eslint-config package: no-inline-styles, no-raw-surface, no-hardcoded-radius, and no-hardcoded-typography. The rules are well-structured and logically sound for enforcing design token usage.**

**Positive:**
- The EXEMPT exemption logic in no-hardcoded-radius.js (L57) correctly handles the case where a class string contains both an exempt class (rounded-full) and a flagged class (rounded-lg) — it strips the exempt match before re-testing, preventing false negatives.
- The components/ui/ file path exemption is consistently applied to all three className-checking rules, correctly excluding primitive component definitions from the rules they enforce on consumers.
- The no-inline-styles rule correctly distinguishes between literal-valued properties (flagged) and computed CSS variable injection via bracket notation (allowed), which is the right call for supporting CSS custom property injection patterns.

---

packages/eslint-config/rules/no-hardcoded-radius.js
zenith-oi, line 40: **FLAG** — [DRY / code duplication] The className extraction logic (Literal → string, TemplateLiteral → quasis join, CallExpression → string args) is copy-pasted identically across no-hardcoded-radius.js, no-hardcoded-typography.js, and no-raw-surface.js (which also omits the CallExpression branch — see related flag). This was flagged in the previous round and remains unresolved. Extract a shared `extractClassString(node)` helper in a `utils.js` file and require it in each rule. Besides being a DRY violation, a bug fix (e.g. supporting nested cn() calls) currently requires three identical edits.

packages/eslint-config/rules/no-raw-surface.js
zenith-oi, line 51: **FLAG** — [incomplete-cn-handling] no-raw-surface.js falls through to `return {}` on any JSXExpressionContainer that isn't a TemplateLiteral — this silently skips `className={cn('bg-[var(--surface-elevated)]', someClass)}` (CallExpression), which is the most common real-world usage. The other three rules handle this case. This was flagged in the previous round and remains unresolved. Add the CallExpression → string-arg accumulation branch that the other rules already implement (ideally via the shared helper mentioned above).

zenith-oi (approved):
## Zenith Code Review

**This PR adds four new ESLint rules to the design system's eslint-config package (no-inline-styles, no-raw-surface, no-hardcoded-radius, no-hardcoded-typography) and bumps the plugin version to 2.0.0. The rules are well-structured and purposeful, but two previously flagged issues remain unresolved.**

**Positive:**
- The exempt-path check (`components/ui/`) is a thoughtful design decision — component primitives need to own their own styling without triggering these rules.
- The EXEMPT regex in no-hardcoded-radius correctly handles the mixed case where a string contains both an allowed class (rounded-full) and a flagged one — the `replace(EXEMPT, '')` re-check on L57 is a nice touch.
- The no-inline-styles rule correctly distinguishes `style={variable}` (allowed) from `style={{ literal: value }}` (flagged), and the CSS variable injection carve-out (`prop.computed` check) prevents false positives on `style={{ ['--foo']: val }}`.

---

PR #36, opened 2026-04-27
feat(mcp-server): expose Canvas design system via MCP (PR-A1, Wave 1)

apaksh-gupta:
## Summary

First leg of Wave 1 toward making the Canvas design system 100x more powerful — this PR introduces `@amplify/mcp-server`, a queryable contract surface for the design system that Pixel and Claude Code can consume directly instead of carrying hardcoded component lists in their prompts.

Replaces:
- Pixel's `src/prompts/component-gen.ts` AMPLIFY_UI_COMPONENTS hardcoded list (43 components, missing 20+ Canvas primitives)
- Claude Code's lack of any structured Canvas knowledge

## What's in here

**New package: `packages/mcp-server`**

5 tools:
| Tool | Purpose |
|------|---------|
| `list_components` | All 46 components with variants/sizes/forwardRef metadata, filterable |
| `get_props` | Full prop signature — types, defaults, enum values, suggestions for typos |
| `find_block` | Templates package search by use case (checkout, stepper, wallet…) |
| `validate_usage` | Lint a JSX snippet against contracts — flags unknown props + invalid enum values |
| `suggest_token` | Nearest Canvas token for a raw hex/spacing/etc. value |

Two transports:
- **stdio** (`amplify-mcp` bin) — for Claude Code via `.mcp.json`
- **HTTP** (`amplify-mcp-http` bin, port 3500) — for Pixel runtime

**How the catalog is built** — Filesystem extraction from `packages/ui/src/components/` at startup. Regex-based prop extraction (string-literal unions + interface body + default values from React.forwardRef destructure). PR-A3 will replace this with JSON contracts emitted at build time.

**Root scripts added** — `npm run build:mcp`, `npm run mcp`, `npm run mcp:http`.

## Test plan

- [x] `npm run test -w packages/mcp-server` — 9/9 pass
- [x] `npm run build -w packages/mcp-server` — clean tsup build, dts emitted
- [x] Smoke: list_components returns 46 components incl. Button/Card/Badge/Input/Dialog/DataTable
- [x] Smoke: get_props extracts Button variants `[primary, secondary, ghost, destructive, outline]` + sizes `[sm, md, lg]` + default `variant=primary`
- [x] Smoke: validate_usage flags `<Button variant="plumber">` as invalid enum
- [x] Smoke: find_block finds checkout in templates
- [x] Smoke: suggest_token returns near matches for `#7C3AED`
- [ ] Reviewer to verify stdio transport works via Claude Code `.mcp.json` after merge + publish

[...]

PR #37, opened 2026-04-27
feat(llms-docs): emit llms.txt + per-component .md at build (PR-A2, Wave 1)

apaksh-gupta:
## Summary

Second leg of Wave 1 Tier 1 — every UI component now has a machine-readable rule sheet generated from its TSX source on every build. Replaces \"AI agents have to read source code or guess\" with \"AI agents read a structured spec\".

## What's in here

**New script: `scripts/generate-llms-docs.mjs`**

Walks `packages/ui/src/components/` at build time, reuses the same regex extractor as `@amplify/mcp-server`, and emits three artifacts under `packages/ui/dist/`:

| Artifact | Format | Purpose |
|----------|--------|---------|
| `llms.txt` | llmstxt.org spec | Root index, lists all 46 components with variant/size/prop counts |
| `llms/<Name>.md` | Markdown | Per-component spec — variants, sizes, props (table with required/default/allowed), subcomponents, forwardRef, example, import |
| `llms.json` | JSON | Machine-readable mirror for tools that prefer structured data |

**Hook:** Wired into `packages/ui` build script — runs after tsup. Output ships in the published npm package alongside `dist/index.js`, so consumers can `cat node_modules/@amplify/ui/dist/llms.txt`.

**Override extension point:** Optional per-component semantic overrides at `packages/ui/src/components/<Name>/<Name>.llm.md` get merged in verbatim under a \"Guidance\" section (when-NOT-to-use rules, accessibility notes, etc.). None ship in this PR — extension reserved for follow-up.

## Sample output

`llms.txt`:
```
# Amplify Canvas Design System
> Canvas is the unified design system for Amplify products...
## Components
- [Button](llms/Button.md): 5 variants, 3 sizes, 5 props
- [Card](llms/Card.md): 5 variants, 3 props
...
```

`llms/Button.md` includes a full prop table:
```
| Name | Type | Required | Default | Allowed values |
|------|------|----------|---------|----------------|
| variant | ButtonVariant | no | primary | primary, secondary, ghost, destructive, outline |
| size | ButtonSize | no | md | sm, md, lg |
...
```

## Test plan

- [x] Generator runs clean: 46 component docs + llms.txt + llms.json, zero failures
- [x] `npm run build -w packages/ui` succeeds end-to-end (tsup → llms generator)
- [x] Spot check: Button/Card props, variants, defaults, allowed values all correctly extracted
- [ ] Reviewer to verify llms.txt is accessible on Storybook GH Pages deploy after merge (separate follow-up if needed)

## Wave 1 progress

- [x] PR-A1 — MCP server (#36, merged)
- [x] PR-A2 — llms.txt generator (this)
- [ ] PR-A3 — JSON component contracts at build (will subsume llms.json + extend with variant×size×state matrix)
- [ ] PR-A4 — Bump `@amplify/ui` to 1.0.0
- [ ] PR-P1..P4, PR-AT1..AT4

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -14,7 +14,7 @@
   },
   "files": ["dist"],
   "scripts": {
-    "build": "tsup src/index.ts --format esm --dts --external react --external react-dom",
+    "build": "tsup src/index.ts --format esm --dts --external react --external react-dom && node ../../scripts/generate-llms-docs.mjs",
     "dev": "tsup src/index.ts --format esm --dts --external react --external react-dom --watch",
     "lint": "eslint src/",
     "test": "vitest run"
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -14,7 +14,8 @@
   },
   "files": ["dist"],
   "scripts": {
-    "build": "tsup src/index.ts --format esm --dts --external react --external react-dom && node ../../scripts/generate-llms-docs.mjs",
+    "build": "tsup src/index.ts --format esm --dts --external react --external react-dom",
+    "postbuild": "node ../../scripts/generate-llms-docs.mjs",
     "dev": "tsup src/index.ts --format esm --dts --external react --external react-dom --watch",
     "lint": "eslint src/",
     "test": "vitest run"
--- a/scripts/generate-llms-docs.mjs
+++ b/scripts/generate-llms-docs.mjs
@@ -19,12 +19,61 @@ const componentsDir = resolve(repoRoot, 'packages/ui/src/components');
 const distDir = resolve(repoRoot, 'packages/ui/dist');
 const docsDir = resolve(distDir, 'llms');
 
+const log = (msg) => process.stdout.write(`[llms-docs] ${msg}\n`);
+const warn = (msg) => process.stderr.write(`[llms-docs] ${msg}\n`);
+
 const TYPE_ALIAS_RE = /export\s+type\s+(\w+)\s*=\s*([^;]+);/g;
-const INTERFACE_RE = /export\s+interface\s+(\w+Props)(?:\s+extends\s+[^{]+)?\s*\{([^}]*)\}/s;
+const INTERFACE_HEADER_RE = /export\s+interface\s+(\w+Props)(?:<[^>]*>)?(?:\s+extends\s+[^{]+)?\s*\{/g;
 const PROP_LINE_RE = /^\s*(\w+)(\??):\s*([^;]+);?\s*$/;
 const FORWARD_REF_RE = /React\.forwardRef\s*</;
 const SUBCOMPONENT_RE = /export\s+(?:const|function)\s+(\w+(?:Header|Title|Description|Content|Footer|Body|Item|Trigger|Group))/g;
 const DEFAULT_VALUE_RE = /(\w+)\s*=\s*([^,\n}]+?)(?:[,\n}])/g;
+
+/**
+ * Extract the interface body using brace-depth counting so nested types
+ * (e.g. `defaultValue?: { label: string; value: string }`) don't truncate
+ * the match early. Returns null if no Props interface is found or braces
+ * are unbalanced.
+ */
+const extractInterfaceBody = (source) => {
+  INTERFACE_HEADER_RE.lastIndex = 0;
+  const headerMatch = INTERFACE_HEADER_RE.exec(source);
+  if (!headerMatch) return null;
+  const start = headerMatch.index + headerMatch[0].length;
+  let depth = 1;
+  for (let i = start; i < source.length; i++) {
+    const ch = source[i];
+    if (ch === '{') depth++;
+    else if (ch === '}') {
+      depth--;
+      if (depth === 0) return source.slice(start, i);
+    }
+  }
+  return null;
+};
+
+/**
+ * Split an interface body into prop declarations, respecting brace depth so
+ * a multi-line nested object type stays on one logical "line".
+ */
+const splitPropDeclarations = (body) => {
+  const out = [];
+  let buf = '';
+  let depth = 0;
+  for (const ch of body) {
+    if (ch === '{' || ch === '<' || ch === '(') depth++;
+    else if (ch === '}' || ch === '>' || ch === ')') depth = Math.max(0, depth - 1);
+    if ((ch === ';' || ch === '\n') && depth === 0) {
+      const trimmed = buf.trim();
+      if (trimmed) out.push(trimmed);
+      buf = '';
+      continue;
+    }
+    buf += ch;
+  }
+  if (buf.trim()) out.push(buf.trim());
+  return out;
+};
 
 const extractEnum = (typeBody) => {
   const stripped = typeBody.trim();
@@ -46,17 +95,19 @@ const extractFromSource = (filePath, name) => {
   const variants = aliases.get(`${name}Variant`) ?? null;
   const sizes = aliases.get(`${name}Size`) ?? null;
   const props = [];
-  const interfaceMatch = INTERFACE_RE.exec(source);
-  if (interfaceMatch) {
-    for (const line of interfaceMatch[2].split('\n')) {
-      const propMatch = PROP_LINE_RE.exec(line);
+  const body = extractInterfaceBody(source);
+  if (body) {
+    for (const decl of splitPropDeclarations(body)) {
+      // Match: `name?: type` (type may span multiple lines, no semicolon since splitter strips it)
+      const propMatch = /^(\w+)(\??):\s*([\s\S]+)$/.exec(decl);
       if (!propMatch) continue;
       const [, propName, optional, type] = propMatch;
+      const cleanType = type.trim();
       props.push({
         name: propName,
-        type: type.trim(),
+        type: cleanType,
         optional: optional === '?',
-        enumValues: aliases.get(type.trim()) ?? extractEnum(type),
+        enumValues: aliases.get(cleanType) ?? extractEnum(cleanType),
       });
     }
   }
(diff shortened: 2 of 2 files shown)


scripts/generate-llms-docs.mjs
```diff
@@ -0,0 +1,232 @@
...
+      variants: c.variants,
+      sizes: c.sizes,
+      props: c.props,
+      hasForwardRef: c.hasForwardRef,
+      subcomponents: c.subcomponents,
+    })),
+  };
+  writeFileSync(join(distDir, 'llms.json'), JSON.stringify(json, null, 2));
+
+  console.log(`[llms-docs] wrote ${components.length} component docs → ${docsDir}`);
```
zenith-oi: **FLAG** — [pii-logging-discipline / console.log in production scripts] The script uses `console.log` / `console.error` directly (lines 223–229). Per Amplify coding standards, production code must use a structured logger, not `console.log`. For a build script this is lower risk than in a running service, but the standard applies repo-wide. Either accept this intentionally (add a comment noting it's a one-shot build script) or swap to a minimal structured output. If this runs in CI, raw console output is fine in practice, but it should be acknowledged rather than left implicit.

packages/ui/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "files": ["dist"],
   "scripts": {
-    "build": "tsup src/index.ts --format esm --dts --external react --external react-dom",
+    "build": "tsup src/index.ts --format esm --dts --external react --external react-dom && node ../../scripts/generate-llms-docs.mjs",
```
zenith-oi: **FLAG** — [build-coupling] The doc-generation script is chained directly to the `build` command with `&&`. If `generate-llms-docs.mjs` fails (e.g., a component file with unusual syntax trips a regex), it will fail the entire UI build and block publishing. Consider running it as a post-build step that exits 0 on failures (already soft-handled internally for per-component errors, but a top-level uncaught exception or the `process.exit(1)` at L144 will still hard-fail the build). Either wire it via a separate `postbuild` npm lifecycle hook so a doc-gen failure doesn't kill `dist/` artifact production, or ensure the script never exits non-zero in environments where the components directory may not exist (e.g. partial builds, storybook-only builds).

scripts/generate-llms-docs.mjs
```diff
@@ -0,0 +1,232 @@
...
+import { fileURLToPath } from 'node:url';
+
+const here = dirname(fileURLToPath(import.meta.url));
+const repoRoot = resolve(here, '..');
+const componentsDir = resolve(repoRoot, 'packages/ui/src/components');
+const distDir = resolve(repoRoot, 'packages/ui/dist');
+const docsDir = resolve(distDir, 'llms');
+
+const TYPE_ALIAS_RE = /export\s+type\s+(\w+)\s*=\s*([^;]+);/g;
+const INTERFACE_RE = /export\s+interface\s+(\w+Props)(?:\s+extends\s+[^{]+)?\s*\{([^}]*)\}/s;
```
zenith-oi: **FLAG** — [regex-fragility] The `INTERFACE_RE` regex (L23) uses `[^}]*` in the body match with the `s` flag, which means any interface whose body contains a nested type with a `}` (e.g., `defaultValue?: { label: string; value: string }`) will silently truncate prop extraction at the first `}`. This is a silent data-correctness issue: the generated docs will drop props without any warning. Consider using a proper bracket-depth counter or a multi-pass approach for interfaces with complex prop types. At minimum, emit a warning when a parsed interface's prop count is suspiciously low (0 props on a known-complex component).

zenith-oi (dismissed):
## Zenith Code Review

**This PR adds a build-time script that generates LLM-readable documentation (llms.txt, per-component Markdown, and llms.json) from the design system's component source files, wired into the ui package build step. The script is well-structured and the approach is sound.**

**Positive:**
- Per-component failures are caught and soft-skipped with descriptive messages — the script degrades gracefully rather than hard-failing on a single malformed component.
- The optional `.llm.md` override mechanism is a clean extension point that avoids polluting component source files with doc-only content.
- The llmstxt.org spec is followed correctly, and the parallel JSON mirror for tool-based consumers is a thoughtful addition.
- The script is self-contained pure Node.js with no added runtime dependencies — zero impact on the package's production bundle.

---

scripts/generate-llms-docs.mjs
```diff
@@ -0,0 +1,303 @@
...
+import { join, resolve, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const here = dirname(fileURLToPath(import.meta.url));
+const repoRoot = resolve(here, '..');
+const componentsDir = resolve(repoRoot, 'packages/ui/src/components');
+const distDir = resolve(repoRoot, 'packages/ui/dist');
+const docsDir = resolve(distDir, 'llms');
+
+const log = (msg) => process.stdout.write(`[llms-docs] ${msg}\n`);
```
zenith-oi, line 25: **FLAG** — [pii-logging-discipline] The previous round flagged `console.log` / `console.error` usage. The re-review shows the author replaced those with `process.stdout.write` / `process.stderr.write` via the `log` and `warn` helpers (L22-L23) — that is an improvement, but per Amplify coding standards, production scripts should still use a structured logger rather than raw stdio writes. `process.stdout.write` is marginally better than `console.log` but not meaningfully different for log aggregation purposes — CloudWatch and similar agents pick up both stdout and stderr. The real requirement is structured JSON output (event name + fields) so log queries are reliable. Since this is a build-time Node script (not a long-lived service), the blast radius is low, but the standard still applies. Consider a minimal structured wrapper: `process.stdout.write(JSON.stringify({level:'info', event:'llms_docs_generated', componentCount}) + '\n')` at the summary lines, at minimum.

zenith-oi (dismissed):
## Zenith Code Review

**Adds a postbuild script that auto-generates LLM-readable documentation (llms.txt, per-component .md files, llms.json) from the UI component source tree. The implementation is solid — graceful degradation on missing dirs, brace-depth-aware interface parsing, and explicit soft-fail wrapping. Previous review issues have been addressed.**

**Positive:**
- Brace-depth-aware interface body extraction (L38-L53) and prop-declaration splitting (L59-L76) correctly handle nested generic types — this is materially better than the regex-only approach flagged in the prior round.
- Graceful soft-fail design: missing componentsDir exits cleanly (L197-L200), individual component extraction failures are collected rather than thrown (L216-L221), and the top-level try/catch (L299-L303) ensures the parent build never hard-fails due to this script.
- The suspicious-props warning (L225-L228) — surfacing when a Props interface parses to zero props — is a good defensive signal that will catch regex regressions without failing the build.
- Re-reading the source file only once per component (with the exception of the declaresPropsInterface check at L225, which is a second read) is clean; the small second read at L225 is isolated and low-cost.

---

scripts/generate-llms-docs.mjs
```diff
@@ -0,0 +1,313 @@
...
+    try {
+      extracted = extractFromSource(filePath, dir);
+    } catch (err) {
+      failures.push({ name: dir, reason: err instanceof Error ? err.message : String(err) });
+      continue;
+    }
+    // Safety net: a Props interface that yields zero props is almost always
+    // an extraction failure (regex truncation, unusual syntax, etc.) rather
+    // than a genuine zero-prop component. Surface as a warning.
+    const declaresPropsInterface = /export\s+interface\s+\w+Props\b/.test(readFileSync(filePath, 'utf8'));
```
zenith-oi: **FLAG** — [error-handling-silent-catch] The suspicious-props check at L230 calls `readFileSync` on `filePath` a second time after already reading it in `extractFromSource` at L92. This is a redundant I/O call on every component. The source string is available inside `extractFromSource` — expose it as part of the return value (e.g. add a `declaresPropsInterface` boolean to the returned object computed at parse time) so the second read is eliminated. Not blocking, but this scales linearly with the number of components on every build.

zenith-oi (dismissed):
## Zenith Code Review

**This PR adds a build-time script that generates LLM-readable documentation (llms.txt, per-component Markdown, llms.json) from the design system's component source files. The implementation is thoughtful — graceful degradation, error isolation, structured logging, and a clear soft-fail contract. One resolved logging issue from the previous round is confirmed addressed.**

**Positive:**
- The top-level try/catch with `exit 0` (implicit — no `process.exit(1)`) is the right pattern for a postbuild hook: doc generation must never block artifact delivery.
- Per-component failure isolation (L221-L226) is excellent — one malformed component file doesn't abort generation of all others.
- Structured final summary line (L301-L305) is a clean pattern for log aggregators to index build telemetry without parsing prose.
- The previous-round PII/console-log issue is genuinely resolved: all output goes through the `log`/`warn`/`summary` wrappers using `process.stdout.write` / `process.stderr.write`, never `console.log`.
- Brace-depth-counting in `extractInterfaceBody` and `splitPropDeclarations` is the correct approach for handling nested generic types without a full AST — far more robust than a naive single-regex approach.

---

zenith-oi (approved):
## Zenith Code Review

**This PR adds a build-time script that auto-generates LLM-readable documentation (llms.txt, per-component Markdown, llms.json) from the UI component source tree, hooked in as a postbuild step. The implementation is well-structured and handles failures gracefully with exit-0 semantics.**

No issues found.

**Positive:**
- The brace-depth-aware interface body extractor (L43–58) and prop-declaration splitter (L64–81) are genuinely solid — they correctly handle nested generic types and multi-line object literals that naive regex would truncate.
- Excellent failure isolation: per-component try/catch (L222–227) ensures one malformed component never aborts the entire run, and all failures are collected and reported in the structured summary line.
- The 'soft no-op when componentsDir is missing' guard (L203–206) makes the script safe to run in partial build contexts like Storybook-only builds — avoids a class of accidental CI failures.
- The structured JSON summary line (L301–305) emitted to stdout is well-designed for log aggregators and makes doc-gen runs queryable in CloudWatch/Datadog without parsing prose.

---

PR #54, opened 2026-04-28
ci: skip publish step gracefully when NPM_TOKEN not set

apaksh-gupta:
## Summary

Without this guard, every push to main fails CI because the publish step runs `npm publish` against an empty `NODE_AUTH_TOKEN` and exits 1 with ENEEDAUTH. That's avoidable noise — builds and tests pass cleanly, only publishing is gated on a secret that's intentionally not set yet (we're awaiting npm org creation).

### After this PR

- Builds + tests + secret scan run on every push (unchanged)
- Publish step runs **only when `secrets.NPM_TOKEN` is set**
- When `NPM_TOKEN` is missing, a GitHub Actions notice surfaces the exact setup steps to enable publishing

### Setup steps (also embedded in the workflow notice)

1. Create org `amplify` at https://www.npmjs.com/org/create (free, public packages)
2. Generate automation token with publish scope
3. `gh secret set NPM_TOKEN -R One-Impression/amplify-design-system`
4. Push any commit to main (or manually re-run "Design System CI") → publish runs

### Why this matters

Currently main shows red on every commit. After merging this, main goes green for everything except publishing — clearer signal of actual code health vs. infra-secret state.

No code changes — workflow only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,9 @@ jobs:
         run: npm run build -w packages/storybook
 
       - name: Publish packages
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' && env.NPM_TOKEN_SET == 'true'
+        env:
+          NPM_TOKEN_SET: ${{ secrets.NPM_TOKEN != '' }}
         run: |
           set -e
           for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui mcp-server eslint-config; do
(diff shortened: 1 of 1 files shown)


zenith-oi (approved):
## Zenith Code Review

**This PR adds a graceful skip for the npm publish step when NPM_TOKEN is not yet configured, preventing CI failures on forks or repos without the secret. The change is small, correct, and well-documented.**

No issues found.

**Positive:**
- The `env.NPM_TOKEN_SET` guard pattern is the right approach for GitHub Actions — secrets evaluate to empty string when unset, and the `!= ''` expression produces a reliable boolean that can be consumed in `if:` conditions without exposing the secret value.
- The `::notice` annotation on the skip step surfaces the exact setup instructions inline in the GitHub Actions UI, which will save time for anyone enabling publishing for the first time.

---

PR #51, opened 2026-04-28
feat(release): migrate to public npm with @amplify scope (v2.0.0) — reverses GH Packages constraint

apaksh-gupta:
## Summary — Phase 0.5 / Phase 2.A.1 finalized: @amplify scope on public npm

This reverses PR #42 (the @amplify → @one-impression GH Packages rename) and switches Canvas distribution to **public npm with `@amplify` scope** — fully available, zero claimants verified.

### Why this approach (Option B from the architectural review)

Per user directive 2026-04-28: brand "Amplify" everywhere, no "one-impression" naming in user-visible artefacts. The GH Packages "scope must match org" constraint was real but solvable by going to public npm where scope ownership is first-publish-wins. This honors the brand directive **and** eliminates auth friction for every consumer.

| | Before (GH Packages) | After (public npm) |
|---|---|---|
| Scope | `@one-impression/*` | `@amplify/*` |
| Registry | npm.pkg.github.com | registry.npmjs.org |
| Consumer install | `.npmrc` + `NODE_AUTH_TOKEN` required | `npm install @amplify/ui` — no auth |
| Visibility | private | public (typical for design systems — Polaris/MUI/Carbon all public) |
| Version | 1.x | 2.0.0 (major — registry change is breaking) |

## Files changed

- 9 `packages/*/package.json` — name → `@amplify/*`, version → `2.0.0`, publishConfig → public npm + access:public, internal deps pinned `^2.0.0`
- `.github/workflows/ci.yml` — registry-url, NODE_AUTH_TOKEN→NPM_TOKEN, removed `--access restricted`, updated 409-conflict regex for public npm error variants
- 48 source/doc files: `@one-impression` → `@amplify` (CLAUDE.md, README, contracts loader, llms-docs generator, Storybook stories, mcp-server, etc.)

## Required setup BEFORE merging

This PR cannot publish until these are done:

- [ ] **Create npm org** `amplify` at https://www.npmjs.com/org/create (free for public packages)
- [ ] **Generate automation token** with publish scope on the new org
- [ ] **Add NPM_TOKEN secret** to this repo: `gh secret set NPM_TOKEN -R One-Impression/amplify-design-system`
- [ ] **Add team members** to `amplify` npm org

Once those are done, merge this PR and CI publishes all 9 packages to public npm — claiming the `@amplify` scope permanently for One-Impression.

## What this unblocks

[...]

PR #55, opened 2026-04-28
chore(release): @amplify-ai/* v2.0.1 published to public npm (follow-up to #51)

apaksh-gupta:
## Summary — @amplify-ai/* v2.0.1 LIVE on public npm

Follow-up to #51 (the @amplify → @amplify-ai scope migration). All 9 packages now published and installable zero-auth from public npm.

### Verified live

| Package | Version | Install |
|---|---|---|
| @amplify-ai/tokens-foundation | 2.0.1 | ✅ |
| @amplify-ai/tokens-brand | 2.0.1 | ✅ |
| @amplify-ai/tokens-atmosphere | 2.0.1 | ✅ |
| @amplify-ai/tokens-creator | 2.0.1 | ✅ |
| @amplify-ai/ui | 2.0.1 | ✅ |
| @amplify-ai/mcp-server | 2.0.1 | ✅ |
| @amplify-ai/eslint-config | 2.0.1 | ✅ |
| @amplify-ai/feature-flags | 2.0.1 | ✅ |
| @amplify-ai/templates | 2.0.1 | ✅ |

### Changes in this PR

1. Versions bumped 2.0.0 → 2.0.1 across all packages
2. Removed `publishConfig` block from each package.json — it was triggering 404 in publish (conflicted with registry resolution path)
3. Cleaned `.npmrc` — was referencing `${NODE_AUTH_TOKEN}` which broke local publishes; now just the scope mapping
4. Internal cross-package deps pinned to `^2.0.1`

### Path used to publish (for future reference)

The publish required:
- Manual `npm login --auth-type=web` for session auth
- Disabling org-level "require 2FA for writes" was needed (free tier auto-enables it)
- Each first publish prompts EOTP web-auth flow; subsequent within ~5 min reuse cached auth
- Removed publishConfig was the key fix — with it, publishes 404'd before reaching the auth check

### What unblocks next

- Atmosphere PR #99 (PR-AT2 Button migration): flip imports `@one-impression/ui@^1.1.0` → `@amplify-ai/ui@^2.0.1`, drop `.npmrc`, drop `NODE_AUTH_TOKEN` wiring — clean public-npm install
- All future product repos (brand-platform, creator-app) install Canvas zero-auth

### DIFF
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@amplify/eslint-config",
-  "version": "2.0.0",
+  "name": "@amplify-ai/eslint-config",
+  "version": "2.0.1",
   "description": "Shared ESLint config enforcing Amplify design token usage",
   "main": "index.js",
   "scripts": {},
@@ -10,9 +10,5 @@
   "files": [
     "index.js",
     "rules/"
-  ],
-  "publishConfig": {
-    "registry": "https://registry.npmjs.org",
-    "access": "public"
-  }
+  ]
 }
--- a/packages/mcp-server/src/server.ts
+++ b/packages/mcp-server/src/server.ts
@@ -10,7 +10,7 @@ import { z } from 'zod';
 const tools = [
   {
     name: 'list_components',
-    description: 'List components in @amplify/ui with their variants, sizes, and forwardRef status.',
+    description: 'List components in @amplify-ai/ui with their variants, sizes, and forwardRef status.',
     schema: listComponentsSchema,
     handler: listComponents,
   },
@@ -60,7 +60,7 @@ const zodToInputSchema = (schema: z.ZodTypeAny): Record<string, unknown> => {
 
 export const createCanvasServer = (): Server => {
   const server = new Server(
-    { name: '@amplify/mcp-server', version: '0.1.0' },
+    { name: '@amplify-ai/mcp-server', version: '0.1.0' },
     { capabilities: { tools: {} } }
   );
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
           node-version: 20
           cache: npm
           registry-url: https://registry.npmjs.org
-          scope: '@amplify'
+          scope: '@amplify-ai'
 
       - name: Install dependencies
         run: npm ci
@@ -37,10 +37,10 @@ jobs:
           npm run build -w packages/tokens-creator
           npm run build -w packages/tokens-atmosphere
 
-      - name: Build @amplify/ui
+      - name: Build @amplify-ai/ui
         run: npm run build -w packages/ui
 
-      - name: Build @amplify/mcp-server
+      - name: Build @amplify-ai/mcp-server
         run: npm run build -w packages/mcp-server
 
       - name: Validate cross-package consistency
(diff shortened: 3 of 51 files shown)


packages/eslint-config/package.json
```diff
@@ -10,9 +10,5 @@
   "files": [
     "index.js",
     "rules/"
-  ],
-  "publishConfig": {
-    "registry": "https://registry.npmjs.org",
-    "access": "public"
-  }
-}
+  ]
```
zenith-oi, line 13: **FLAG** — [publishConfig removed] The `publishConfig` block (registry + access:public) has been removed from every package.json. Without `"access": "public"` in publishConfig, npm will default to restricted access for scoped packages on first publish to a new scope. Since @amplify-ai is a new scope that has never published these packages before, the first `npm publish` will fail with a 402/403 unless the CI command explicitly passes `--access public`, or the org default is set to public. The CI publish loop in ci.yml does not pass `--access public` either. Either restore `publishConfig` with `"access": "public"` to each package.json, or add `--access public` to the publish command in ci.yml (line 83 area). Same issue affects: packages/feature-flags, packages/mcp-server, packages/templates, packages/tokens-atmosphere, packages/tokens-brand, packages/tokens-creator, packages/tokens-foundation, packages/ui.

packages/eslint-config/package.json
```diff
@@ -10,9 +10,5 @@
...
+}
```
zenith-oi, line 14: **FLAG** — [missing-newline-eof] Multiple package.json files are missing a trailing newline (diff shows `\ No newline at end of file`): eslint-config, feature-flags, mcp-server, templates, tokens-atmosphere, tokens-brand, tokens-creator, tokens-foundation, ui. This is a minor tooling concern but can cause noisy diffs and some npm/yarn tools emit warnings. Not blocking, but worth fixing in the same pass given everything is already touched.

packages/mcp-server/src/server.ts
```diff
@@ -60,7 +60,7 @@ const zodToInputSchema = (schema: z.ZodTypeAny): Record<string, unknown> => {
 
 export const createCanvasServer = (): Server => {
   const server = new Server(
-    { name: '@amplify/mcp-server', version: '0.1.0' },
+    { name: '@amplify-ai/mcp-server', version: '0.1.0' },
```
zenith-oi, line 63: **FLAG** — [stale-version-string] The MCP server's `createCanvasServer()` call hardcodes `version: '0.1.0'` in the Server metadata even though the package.json version is now 2.0.1. This will cause the MCP protocol handshake to report the wrong version to connected agents (Pixel, Claude Code), making it impossible to correlate logs or debug version-specific behaviour. Should be `version: '2.0.1'` or, better, imported from package.json so it stays in sync automatically.

zenith-oi (approved):
## Zenith Code Review

**Scope rename from @amplify/* to @amplify-ai/* across all packages in the monorepo — a mechanical find-and-replace touching package names, imports, documentation, CI config, and the lock file. The change is clean and consistent.**

**Positive:**
- The rename is exhaustive and consistent — every import, comment, documentation string, CI step name, lock file entry, and runtime string literal has been updated. No stale @amplify/* references are visible in the diff.
- Version bumps (2.0.0 → 2.0.1) are applied uniformly across all packages, and inter-package dependency version pins have been updated to match (e.g. tokens-atmosphere now requires @amplify-ai/tokens-foundation ^2.0.1 rather than ^2.0.0).

---

PR #56, opened 2026-04-28
fix(ui): add 'use client' to DensityProvider for Next.js RSC compat (v2.0.2)

apaksh-gupta:
## Summary

Canvas's `DensityProvider` uses `React.createContext` — a client-component-only API. Without an explicit `'use client'` at the module entry, Next.js App Router server components fail at SSR with `createContext is not a function`.

Atmosphere PR #106 worked around this with a Client Component bridge — this PR makes that bridge unnecessary going forward.

## Changes

| File | Change |
|---|---|
| `packages/ui/src/lib/density.tsx` | Added `'use client'` at top of module |
| `packages/ui/package.json` | 2.0.1 → 2.0.2 (patch — RSC interop fix) |

## What this unlocks

Products can import `DensityProvider` directly from `@amplify-ai/ui` in their server-component layouts — no client-component bridge needed.

```tsx
// Works after this lands:
import { DensityProvider } from '@amplify-ai/ui';

export default function RootLayout({ children }) {
  return (
    <html>
      <body>
        <DensityProvider density="compact">
          {children}
        </DensityProvider>
      </body>
    </html>
  );
}
```

## Verified

- ✅ `npm run build -w packages/ui` clean
- ✅ `npm test -w packages/ui` — 20 tests passing

## Out of scope

A sweep across other Canvas components that use `createContext` (when they're added) — for now Density is the only one in production use.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10785,21 +10785,21 @@
     },
     "packages/eslint-config": {
       "name": "@amplify-ai/eslint-config",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "peerDependencies": {
         "eslint": ">=8.0.0"
       }
     },
     "packages/feature-flags": {
       "name": "@amplify-ai/feature-flags",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "peerDependencies": {
         "react": ">=17.0.0"
       }
     },
     "packages/mcp-server": {
       "name": "@amplify-ai/mcp-server",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.0.4",
         "zod": "^3.23.8"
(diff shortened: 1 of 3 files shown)


zenith-oi (approved):
## Zenith Code Review

**This PR adds `'use client'` to `density.tsx` for Next.js RSC compatibility and bumps the `@amplify-ai/ui` package version from 2.0.1 to 2.0.2. The change is minimal and correct — adding the directive to a file that uses React context hooks is the right fix.**

No issues found.

**Positive:**
- Correctly identifies that `createContext`/`useContext` require the `'use client'` directive in Next.js App Router environments — this is a real and necessary fix, not a reflexive addition.
- Version bump is appropriately scoped: only `packages/ui/package.json` moves to 2.0.2 while the lock file reflects the broader 2.0.0→2.0.1 workspace sync, indicating the author understands the difference between this change and the prior release.

---

PR #60, opened 2026-04-29
feat(ui): Phase B Wave 2 — motion + conversational components

apaksh-gupta:
## Summary

Adds **8 new client-side components** to `@amplify/ui` (v2.2.0) for the Phase B Wave 2 motion + conversational primitives.

### Conversational primitives
- **MessageBubble** — incoming / outgoing / system variants with avatar, status (sending/sent/delivered/read/failed), reactions, directional tail
- **ChatInput** — auto-resizing multi-line input; Enter / Cmd+Enter to send; attach / voice / emoji slots; controlled + uncontrolled
- **TypingIndicator** — three animated dots with sm/md/lg sizes and configurable color; honors reduced-motion

### Motion primitives
- **Reveal** — mount or scroll-into-view trigger; up / down / left / right / fade directions; polymorphic `as` prop
- **Stagger** — sequential reveal of children with configurable `interval` + `initialDelay`
- **Parallax** — scroll-coupled translation (vertical / horizontal), rAF-throttled scroll listener
- **AnimatedNumber** — eased counting tween, custom `format`, optional `animateOnView`
- **ScrollProgress** — top/bottom linear bar or radial dial; tracks window or a target ref

All motion respects `prefers-reduced-motion` via a shared `useReducedMotion` hook. Durations / easings use sensible defaults with `TODO(phase-a)` markers to swap onto Phase A motion tokens once they land.

### What did **not** change
- Wave 1 marketing components (separate PR, separate files)
- Phase A token additions (separate PR — motion tokens land there first)

## Test plan

- [x] `npm run build:ui` — green (54 contracts written, dts clean)
- [x] `npx tsc --noEmit` — no new errors (only pre-existing `Button.test.tsx` failure shared with main)
- [x] `npm run build:storybook` — green, all 8 new stories built
- [ ] Visual review of each story in Storybook
- [ ] Reduced-motion verification in OS settings (stories include explicit RM stories)
- [ ] Cross-Phase-A merge: re-anchor TODOs onto motion tokens once Phase A merges

## Stats
- 8 components, 17 new files in `packages/ui/src/components/*`
- 1 new lib hook `useReducedMotion`
- 8 story files, 49 stories total
- Bumps `@amplify-ai/ui` 2.0.2 → 2.2.0 (Wave 1 owns 2.1.0)


packages/ui/src/components/MessageBubble/MessageBubble.tsx
```diff
@@ -0,0 +1,158 @@
...
+            bubbleStyles[variant],
+            tail && tailStyles[variant]
+          )}
+        >
+          {children}
+        </div>
+        {reactions && reactions.length > 0 && (
+          <div className="flex flex-wrap gap-1 px-1">
+            {reactions.map((r, i) => (
+              <button
```
zenith-oi, line 121: **FLAG** — [missing-aria-label-on-reaction-buttons] Reaction buttons have no accessible label. A screen reader will announce the emoji character and count but nothing meaningful about the action. Add `aria-label={`React with ${r.emoji}, ${r.count} reaction${r.count !== 1 ? 's' : ''}`}` (or similar) to each reaction button so keyboard/AT users know what the button does and its current state. The `r.reacted` state is also not communicated to AT — consider `aria-pressed={r.reacted}` on each button.

packages/ui/src/components/AnimatedNumber/AnimatedNumber.tsx
```diff
@@ -0,0 +1,114 @@
...
+        previousRef.current = value;
+        rafRef.current = null;
+      }
+    };
+
+    rafRef.current = requestAnimationFrame(tick);
+    return () => {
+      if (rafRef.current != null) cancelAnimationFrame(rafRef.current);
+    };
+  }, [value, duration, easing, reduced, shouldAnimate]);
```
zenith-oi: **FLAG** — [raf-not-cancelled-on-value-change] The animation effect depends on `easing` in its dependency array, but `easing` defaults to the module-level `easeOutCubic` constant so it's stable. However, if a consumer passes an inline arrow function as `easing` prop (e.g. `easing={(t) => t}`), a new function reference is created every render, causing the effect to restart every render cycle. Consider either omitting `easing` from deps and using a ref to hold it, or documenting that callers must memoize a custom easing function. This is a latent footgun for consumers who don't know to `useCallback` their easing prop.

zenith-oi (dismissed):
## Zenith Code Review

**Adds 8 new alpha components (MessageBubble, ChatInput, TypingIndicator, Reveal, Stagger, Parallax, AnimatedNumber, ScrollProgress) plus a shared useReducedMotion hook, with full Storybook coverage. The implementation is solid — good accessibility patterns, consistent reduced-motion handling, proper SSR guards, and clean component contracts. One genuine issue worth flagging.**

**Positive:**
- Consistent reduced-motion handling across all motion components via a single shared useReducedMotion hook — correctly SSR-safe (initialises false, hydrates on mount) and subscribes to live media query changes including the older Safari addListener fallback.
- AnimatedNumber cancels in-flight RAF on cleanup and on value change — no stale animation frames leaking between renders.
- Parallax and ScrollProgress both use RAF-throttled scroll listeners with passive:true, and cancel pending frames on cleanup. Clean pattern throughout.
- ChatInput correctly implements controlled/uncontrolled duality, exposes an imperative handle via forwardRef, and auto-sizes via scrollHeight without layout thrash on every keystroke (driven by value change effect, not input event).
- ScrollProgress covers both window scroll and arbitrary scrollable element via the target ref prop — genuinely useful for in-page panels.
- TypingIndicator's inline keyframe injection (ensureKeyframes idempotency check) avoids needing Tailwind config additions while keeping the animation co-located with the component.
- Stagger composing Reveal rather than duplicating scroll-intersection logic is the right call — single responsibility maintained.

---

packages/ui/src/components/MessageBubble/MessageBubble.tsx
zenith-oi, line 121: **FLAG** — [missing-aria-label-on-reaction-buttons] Still unresolved from the previous round. Reaction buttons render `<span>{r.emoji}</span><span>{r.count}</span>` with no accessible label. A screen reader will announce the raw emoji character and the count integer with no context — e.g. '🔥 3' — with no indication that this is a reaction button or that clicking it toggles the user's own reaction. Add an `aria-label` like `aria-label={\`React with ${r.emoji} (${r.count})\``} or, if the user has already reacted, `\`Remove ${r.emoji} reaction (${r.count})\``. The `onReactionClick` prop and `reacted` flag make the toggling behaviour clear enough to warrant the distinction.

packages/ui/src/components/AnimatedNumber/AnimatedNumber.tsx
zenith-oi: **FLAG** — [raf-not-cancelled-on-value-change] Still unresolved from the previous round. The animation effect lists `easing` in its dependency array. When the caller does not pass a custom `easing` prop the default is the module-level `easeOutCubic` reference, which is stable, so there's no practical problem in that path. However, if a caller passes an inline arrow `easing={(t) => t}` the function reference changes every render, causing the effect to re-run (and re-fire an animation) on every parent render even when `value` hasn't changed. The previous round's note about 'previous RAF not being cancelled' is the consequence: the cleanup from the prior run cancels the old RAF, but the new one starts from `previousRef.current` (which was correctly updated only on completion), so a value change mid-animation resets from wherever the counter stopped rather than from the visually-displayed value. Fix: move the `easing` prop out of the dep array and capture it in a ref (`easingRef.current = easing`) so the running animation always uses the latest easing without restarting. The `format` prop has the same issue at line 111 (used inside render, not in the effect, so it doesn't cause spurious restarts, but worth noting for consistency).

zenith-oi (dismissed):
## Zenith Code Review

**Adds 8 new alpha-status UI components (MessageBubble, ChatInput, TypingIndicator, Reveal, Stagger, Parallax, AnimatedNumber, ScrollProgress) plus a shared useReducedMotion hook, all with Storybook stories. The implementation is solid overall — correct RAF cleanup, proper IntersectionObserver teardown, SSR-safe patterns, and consistent reduced-motion handling across every animated component.**

**Positive:**
- Consistent reduced-motion handling across all animated components via a single shared useReducedMotion hook — no component manually calls matchMedia inline.
- Every scroll/RAF listener is properly cleaned up in effect return functions, including the in-flight frame cancellation on unmount. No listener leaks.
- IntersectionObserver.disconnect() called inside the callback after first intersection (animateOnView, Reveal with once=true), preventing redundant callbacks on re-entry.
- ChatInput correctly implements both controlled and uncontrolled modes with the isControlled flag, and the imperative handle's clear() correctly calls onChange('') so controlled parents stay in sync.
- TypingIndicator's inline keyframes injection via ensureKeyframes() is idempotent (document.getElementById guard) — safe to render multiple instances without duplicating the style block.
- ScrollProgress's getProgress() clamps to [0, 1] and guards against zero-height documents — no divide-by-zero or out-of-range values reaching the SVG strokeDashoffset calculation.

---

packages/ui/src/components/MessageBubble/MessageBubble.tsx
zenith-oi, line 121: **FLAG** — [missing-aria-label-on-reaction-buttons] Reaction buttons still have no accessible label. A screen reader announces them as '🔥 3', '🎉 2', etc. — emoji are not reliably vocalised, and the count alone is meaningless. Add an aria-label that describes the action, e.g. `aria-label={`React with ${r.emoji}, ${r.count} reaction${r.count !== 1 ? 's' : ''}`}`. Still unresolved from the previous round.

packages/ui/src/components/AnimatedNumber/AnimatedNumber.tsx
zenith-oi: **FLAG** — [raf-not-cancelled-on-value-change] The animation effect lists `easing` in its dependency array. If the caller passes an inline arrow function (e.g. `easing={(t) => t}`) the reference changes every render, re-triggering the effect and spawning overlapping RAF loops — the cleanup only cancels the previous frame ID but the new loop starts from `previousRef.current` which may not have been committed yet for the interrupted tween. Either (a) document that `easing` must be a stable reference (useCallback / module-level const), or (b) remove `easing` from the dep array and access it via a ref instead. Still unresolved from the previous round.

zenith-oi (dismissed):
## Zenith Code Review

**Adds 8 new UI components (conversational: MessageBubble, ChatInput, TypingIndicator; motion: Reveal, Stagger, Parallax, AnimatedNumber, ScrollProgress) plus a shared useReducedMotion hook, with full Storybook stories. Implementation quality is high — consistent use of design tokens, proper accessibility attributes, SSR safety, and reduced-motion support throughout. Two issues from the previous round remain unresolved.**

**Positive:**
- useReducedMotion is SSR-safe, handles the deprecated addListener/removeListener Safari path, and is shared across all motion components — clean single-responsibility hook.
- AnimatedNumber cancels the RAF loop on cleanup and tracks previousRef correctly so re-triggers animate from the last displayed value rather than resetting to zero.
- ChatInput correctly implements the controlled/uncontrolled duality (isControlled flag, forwardRef + useImperativeHandle), textarea autosize via scrollHeight, and the expected Enter/Shift+Enter keyboard contract.
- Parallax and ScrollProgress both use a RAF-throttled scroll listener (coalesces rapid scroll events to one frame) and remove listeners cleanly on unmount.
- Stagger delegates entirely to Reveal rather than reimplementing animation logic, keeping the motion system DRY.
- All motion components immediately show content when IntersectionObserver is unavailable (SSR / old browsers) rather than leaving elements invisible.

---

zenith-oi (approved):
## Zenith Code Review

**Adds 8 new components (MessageBubble, ChatInput, TypingIndicator, Reveal, Stagger, Parallax, AnimatedNumber, ScrollProgress) plus a shared `useReducedMotion` hook, with full Storybook coverage. The implementation is solid — reduced-motion handling is thorough, RAF loops are properly cancelled, and the controlled/uncontrolled input pattern in ChatInput is clean.**

No issues found.

**Positive:**
- Thorough reduced-motion handling across every motion component — `useReducedMotion` is shared, SSR-safe, and correctly degrades to immediate/static rendering everywhere.
- The AnimatedNumber RAF loop correctly stores `easing` in a ref so inline arrow-function props don't spawn overlapping animation loops, and the eslint-disable comment explains exactly why the omission is intentional.
- ChatInput's controlled/uncontrolled duality is well-implemented: the `isControlled` flag is computed once, `defaultValue` is only read on mount, and the imperative handle's `clear()` correctly respects both modes.
- ScrollProgress handles both window-scroll and element-scroll targets cleanly, and the radial variant's SVG math (stroke-dasharray/offset from circumference) is correct.
- Stagger's `wrapChildren=false` escape hatch via `React.cloneElement` with `transitionDelay` is a nice affordance for callers who supply their own Reveal.
- TypingIndicator injects its keyframes lazily and idempotently via a single DOM `<style>` tag guarded by `document.getElementById`, avoiding duplicate injection on re-mount.

---

PR #70, opened 2026-05-04
feat(ui): page affordances — Banner, Kbd, Notification, Tour

apaksh-gupta:
## Summary

Adds 4 page-level affordance components flagged in the Canvas audit. All four ship as `beta` (since `2.6.0`) per the project's component-status convention.

### `<Banner>`
Sticky inline page banner. Distinct from `<Toast>` (transient) and `<AnnouncementBar>` (top-of-app marketing strip). Use for deploy notices, system status, action-required messages.
- Variants: `info | warning | error | success`
- `dismissible` + `onDismiss`
- Right-aligned `actions` slot for inline CTAs
- a11y: `role="alert"` + `aria-live="assertive"` for warning/error, `role="status"` + `aria-live="polite"` otherwise

### `<Kbd>`
Inline keyboard-shortcut hint. Pure visual primitive.
- Static mode: `<Kbd>K</Kbd>`
- Platform-aware mode: `<Kbd keys={["mod","K"]} />` → renders ⌘ K on Mac, Ctrl K on Windows/Linux
- Optional `isMac` prop overrides detection (handy for cross-platform docs + tests)
- a11y: `role="img"` with `aria-label` derived from the rendered sequence

### `<Notification>` + `<NotificationList>`
Persistent in-app notification row (distinct from `<Toast>`).
- Unread state with indicator dot, marked-read on click
- Optional `avatar` (or `icon` fallback), `title`, `body`, `timestamp`
- Renders as `<a href>` when `href` is set; otherwise as `role="button"` with keyboard activation (Enter/Space)
- `<NotificationList>` is a thin composition with `role="list"` + accessible label + empty slot

### `<Tour>` + `<TourStepMarker>`
Multi-step product tour overlay.
- Portal-rendered to `document.body` so the backdrop covers the entire viewport
- Spotlight backdrop with cutout + ring around the anchored target (4-panel technique, no clip-path needed)
- Anchored placement (`top | bottom | left | right`) with viewport clamping; `center` fallback for unanchored steps
- Auto re-measures target on `resize` + capture-phase `scroll` so the card follows during in-page navigation
- Keyboard: `Esc` closes, `Enter` / `→` advances, `←` goes back
- Persists step index to `localStorage[amp:tour:{tourId}]` (opt-out via `persist={false}`); clears on completion

## Implementation notes

- Followed existing patterns from `Toast.tsx`, `Tooltip.tsx`, `Dialog.tsx`, `EmptyState.tsx`, `AnnouncementBar.tsx`
- TypeScript strict — no `any`
- Canvas semantic tokens only (with sensible fallbacks for tokens that may not exist yet, e.g. `--amp-semantic-bg-warning-subtle` falls back to `--amp-semantic-bg-raised`)
- No new external dependencies
- 91 components in the contracts manifest after this change (was 87)

## Test plan

- [x] `npm run test` — 25 new tests across 4 files all passing (45/45 total)
- [x] `npm run build` — full repo build green; tsup ESM + dts emit clean; contracts + llms-docs regenerated for 91 components
- [x] `npm run validate` — token reference integrity intact
- [ ] Storybook visual review — `npm run storybook`, check Banner/Kbd/Notification/Tour stories
- [ ] Chromatic visual regression on this PR

## Audit reference

Flagged in the Canvas audit as missing page-level affordance components — Banner (page-level message distinct from Toast/AnnouncementBar), Kbd (keyboard shortcut hint), Notification (persistent in-app notification distinct from Toast), Tour (product onboarding overlay).

### DIFF
--- a/packages/ui/src/components/Banner/Banner.tsx
+++ b/packages/ui/src/components/Banner/Banner.tsx
@@ -1,0 +1,140 @@
+import React from 'react';
+import { cn } from '../../lib/cn';
+
+export type BannerVariant = 'info' | 'warning' | 'error' | 'success';
+
+export interface BannerProps {
+  /** Visual + semantic tone. `warning` and `error` set role="alert" for assertive a11y. */
+  variant?: BannerVariant;
+  /** Short headline. */
+  title?: string;
+  /** Body copy — supports rich content. */
+  message: React.ReactNode;
+  /** Show a dismiss (×) affordance. */
+  dismissible?: boolean;
+  /** Fired when the user dismisses the banner. */
+  onDismiss?: () => void;
+  /** Right-aligned action slot — typically Buttons or links. */
+  actions?: React.ReactNode;
+  className?: string;
+}
+
+const variantStyles: Record<
+  BannerVariant,
+  { container: string; iconColor: string; icon: React.ReactNode }
+> = {
+  info: {
+    container:
+      'bg-[var(--amp-semantic-bg-info-subtle,var(--amp-semantic-bg-raised))] border-[var(--amp-semantic-border-info,var(--amp-semantic-status-info))]',
+    iconColor: 'text-[var(--amp-semantic-status-info)]',
+    icon: (
+      <path strokeLinecap="round" strokeLinejoin="round" d="M13 16h-1v-4h-1m1-4h.01M21 12a9 9 0 11-18 0 9 9 0 0118 0z" />
+    ),
+  },
+  warning: {
+    container:
+      'bg-[var(--amp-semantic-bg-warning-subtle,var(--amp-semantic-bg-raised))] border-[var(--amp-semantic-border-warning,var(--amp-semantic-status-warning))]',
+    iconColor: 'text-[var(--amp-semantic-status-warning)]',
+    icon: (
+      <path strokeLinecap="round" strokeLinejoin="round" d="M12 9v2m0 4h.01m-6.938 4h13.856c1.54 0 2.502-1.667 1.732-3L13.732 4c-.77-1.333-2.694-1.333-3.464 0L3.34 16c-.77 1.333.192 3 1.732 3z" />
+    ),
+  },
+  error: {
+    container:
+      'bg-[var(--amp-semantic-bg-error-subtle,var(--amp-semantic-bg-raised))] border-[var(--amp-semantic-border-error,var(--amp-semantic-status-error))]',
+    iconColor: 'text-[var(--amp-semantic-status-error)]',
+    icon: (
+      <path strokeLinecap="round" strokeLinejoin="round" d="M12 9v3m0 4h.01M5.07 19h13.86c1.54 0 2.5-1.67 1.73-3L13.73 4a2 2 0 00-3.46 0L3.34 16c-.77 1.33.19 3 1.73 3z" />
+    ),
+  },
+  success: {
+    container:
+      'bg-[var(--amp-semantic-bg-success-subtle,var(--amp-semantic-bg-raised))] border-[var(--amp-semantic-border-success,var(--amp-semantic-status-success))]',
+    iconColor: 'text-[var(--amp-semantic-status-success)]',
+    icon: (
+      <path strokeLinecap="round" strokeLinejoin="round" d="M9 12l2 2 4-4m6 2a9 9 0 11-18 0 9 9 0 0118 0z" />
+    ),
+  },
+};
+
+/**
+ * Sticky page-level banner. Renders inline at the top of a page or section
+ * and persists until dismissed. Use for deploy notices, system status, or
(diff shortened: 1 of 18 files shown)


packages/ui/src/components/Tour/Tour.tsx
```diff
@@ -0,0 +1,366 @@
...
+  }, [onClose]);
+
+  // Keyboard handling
+  useEffect(() => {
+    if (!open) return;
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') {
+        e.preventDefault();
+        handleClose();
+      } else if (e.key === 'Enter' || e.key === 'ArrowRight') {
```
zenith-oi, line 151: **FLAG** — [keyboard-navigation-correctness] The global `keydown` listener treats `Enter` as 'advance step', but the Enter key also activates focused buttons (Back, Next, Done, Close). This means when the user focuses the Back button and presses Enter, two handlers fire: the button's native click handler AND the document keydown handler, both decrementing the step index or (worse) conflicting when the Close button fires `handleClose` at the same time as `goNext`. The `goNext` guard in the keyboard handler should check `e.target` isn't an interactive element inside the card, or the handler should only handle Enter when the card container itself is focused. A simple fix: add `if ((e.target as HTMLElement).closest('[role="dialog"]') && e.target !== cardRef.current) return;` before processing Enter/ArrowRight, so button clicks inside the dialog aren't double-counted.

packages/ui/src/components/Tour/Tour.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    expect(screen.queryByRole('dialog')).toBeNull();
+  });
+
+  it('renders first step + step counter when opened', () => {
+    render(<Tour tourId="t1" steps={steps} open persist={false} />);
+    expect(screen.getByText('Step 1')).toBeDefined();
+    expect(screen.getByText('1 of 3')).toBeDefined();
+  });
+
+  it('Enter key advances to next step', () => {
```
zenith-oi, line 28: **FLAG** — [missing-test-coverage] The Enter-key test fires `keyDown` on the document without verifying it doesn't conflict with button-click behavior. No test covers the case where focus is on the Next button and Enter is pressed — which would trigger both the button onClick and the global keydown handler (double-advance). Given the behavioral bug flagged above, a regression test covering this conflict path is important before shipping.

packages/ui/src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,140 @@
...
+          </p>
+        )}
+        <div className="text-[14px] text-[var(--amp-semantic-text-secondary)] leading-snug">
+          {message}
+        </div>
+      </div>
+      {actions && (
+        <div className="flex items-center gap-2 shrink-0 ml-2">{actions}</div>
+      )}
+      {dismissible && (
```
zenith-oi, line 114: **FLAG** — [missing-dismissed-state] The dismiss button calls `onDismiss` but the Banner has no internal `dismissed` state. When `onDismiss` is not provided (prop is optional), clicking the dismiss button is a no-op that keeps the banner visible. The contract in the doc comment says `dismissible` shows a dismiss affordance, but there's no guidance that callers must provide `onDismiss`. Either make `onDismiss` required when `dismissible={true}` at the type level (`dismissible?: false; onDismiss?: never` | `dismissible: true; onDismiss: () => void` discriminated union), or add internal dismissed state so the component hides itself when the button is clicked without a handler. As-is, a caller can render `<Banner dismissible />` and the ✕ button does nothing.

zenith-oi (approved):
## Zenith Code Review

**Adds four new UI components (Banner, Kbd, Notification, Tour) as Phase B Wave 5 additions to the design system. The implementation is solid overall — good a11y patterns, design-token-only styling, comprehensive tests — with one behavioral issue in the Tour component worth addressing.**

**Positive:**
- Consistent use of CSS custom properties throughout — no hardcoded color values anywhere in the four components, fully honoring the CLAUDE.md no-hardcoded-colors rule.
- Excellent a11y decisions across the board: `role="alert"` vs `role="status"` differentiation in Banner based on urgency, `role="img"` + `aria-label` on Kbd for screen-reader key sequence pronunciation, `useId()` for `aria-labelledby`/`aria-describedby` linkage in Tour dialog, spotlight backdrop panels all marked `aria-hidden`.
- Tour localStorage helpers (`readProgress`, `writeProgress`, `clearProgress`) each have try/catch with silent-ignore on quota errors — correct defensive handling for a non-critical storage path.
- Tour step positioning correctly clamps card coordinates within the viewport with an 8px margin, avoiding cards that render partially off-screen.
- Test coverage is genuinely comprehensive: keyboard navigation, persistence behavior, empty state slot, link vs button rendering in Notification, platform-aware key substitution in Kbd.

---

PR #71, opened 2026-05-04
feat(ui): RadioGroup + Combobox primitives

apaksh-gupta:
## Summary

Adds the two missing form primitives to `@amplify-ai/ui` so products stop reinventing them. Both ship at status `beta` (since `2.6.0`) and are exported from the package barrel.

## `<RadioGroup>` + `<Radio>`

Children-based composition with full keyboard a11y.

```tsx
<RadioGroup name=\"color\" value={color} onValueChange={setColor} label=\"Pick a color\">
  <Radio value=\"red\">Red</Radio>
  <Radio value=\"green\">Green</Radio>
  <Radio value=\"blue\">Blue</Radio>
</RadioGroup>
```

**Behavior**
- Arrow keys cycle (skip disabled radios, wrap around).
- Space / Enter selects the focused radio.
- Roving `tabIndex` (only the selected radio is tabbable).
- Hidden native `<input type=\"radio\">` keeps form submission + autofill working.

**Props (RadioGroup)** — `name` (required), `value` / `defaultValue`, `onValueChange`, `orientation: 'horizontal' | 'vertical'`, `disabled`, `readOnly`, `label`, `error`, `helperText`.

**Props (Radio)** — `value` (required), `disabled`, `label` / children, `id`, `className`.

**ARIA** — `role=\"radiogroup\"` + `aria-orientation` + `aria-labelledby` on the container; `role=\"radio\"` + `aria-checked` + `aria-disabled` on each item.

## `<Combobox>`

Searchable single-value select.

```tsx
// Static items
<Combobox label=\"Platform\" items={platforms} value={v} onValueChange={setV} />

// Async items
<Combobox label=\"Country\" loadItems={async (q) => fetch(...).then(r => r.json())} />

// Children composition
<Combobox label=\"Currency\">
  <Combobox.Item value=\"usd\" label=\"USD — US Dollar\" hint=\"Default\" />
  <Combobox.Item value=\"eur\" label=\"EUR — Euro\" />
</Combobox>
```

**Behavior**
- Click / focus → opens listbox.
- Type → filters items (sync) or fires debounced `loadItems(query)` (async).
- Arrow keys, Home/End → navigate; Enter → select; Esc → close; Tab → close + commit focus.
- Empty state when no match (overridable via `emptyMessage`).
- Clear button when a value is set.
- Async loader is debounced (default 250ms) and request-versioned to prevent stale results.

**Props** — `value` / `defaultValue`, `onValueChange`, `items` OR `<Combobox.Item>` children OR `loadItems`, `debounceMs`, `label`, `placeholder`, `error`, `helperText`, `emptyMessage`, `disabled`, `readOnly`, `name`.

**ARIA** — `role=\"combobox\"` on input with `aria-autocomplete=\"list\"`, `aria-expanded`, `aria-controls`, `aria-activedescendant`; `role=\"listbox\"` + `role=\"option\"` with `aria-selected` and `aria-disabled`. Loading + empty states use `role=\"status\"` + `aria-live=\"polite\"`.

## Tests

23 new vitest tests added (10 RadioGroup + 13 Combobox), covering keyboard a11y, ARIA wiring, controlled / uncontrolled state, async loading, error / readOnly variants. Full UI suite: 43/43 green.

## Build / typecheck

- `npm run build --workspaces --if-present` — green; `dist/index.js` 339.84 KB, contracts now report 89 components (was 87).
- `tsc --noEmit` — no new errors on these files; only the pre-existing React-19 ↔ Storybook `ComponentType<any>` mismatch present on every component (Switch, Tooltip, Toast etc.).

## Audit

- No raw colors / spacing — Canvas tokens only (`--amp-semantic-*`).
- No new external deps.
- Both components added to `component-status.json` at `beta` (since `2.6.0`).
- Existing `recipes/SearchCombobox` (async-only recipe) is intentionally left in place; the new `<Combobox>` is the lower-level primitive that products should adopt going forward.

## Test plan

- [ ] Storybook renders all stories (default / with-error / readonly / horizontal / async-search).
- [ ] Keyboard a11y verified manually: arrow keys, space, Enter, Escape, Tab.
- [ ] Brand / Creator / Atmosphere products can drop in `<RadioGroup>` and `<Combobox>` and remove their bespoke radio/searchable-select forks.


packages/ui/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+  React.useEffect(() => {
+    ctx.registerRadio(value, buttonRef.current);
+    return () => ctx.registerRadio(value, null);
+  }, [ctx, value]);
+
+  // Roving tabindex: only the selected radio (or the first if none selected) is tabbable.
+  // We approximate "first" by making radios with no group selection tabbable when this is mounted first.
+  // To stay correct under StrictMode + dynamic children, we also accept tabindex=0 when nothing is selected.
+  const tabIndex = isSelected ? 0 : ctx.value === null ? 0 : -1;
```
zenith-oi, line 224: **FLAG** — [tabindex-all-unselected-tabbable] The roving tabindex logic `const tabIndex = isSelected ? 0 : ctx.value === null ? 0 : -1` makes every radio tabbable when nothing is selected (`ctx.value === null`). Standard roving-tabindex for a radiogroup requires exactly ONE element to be in the tab sequence at any time — when no selection exists the first radio (index 0) should receive `tabIndex=0` and all others `-1`. As written, all radios are simultaneously tabbable on first render with no selection, violating the ARIA Authoring Practices for radiogroup (which is also what the in-file comment says it's trying to do). Fix by tracking a `firstRadioValue` in context (set once on first `registerRadio`) and using `tabIndex = isSelected || (ctx.value === null && value === firstRadioValue) ? 0 : -1`.

packages/ui/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,466 @@
...
+    if (!open) return;
+    const onDocMouseDown = (e: MouseEvent) => {
+      if (!wrapperRef.current?.contains(e.target as Node)) setOpen(false);
+    };
+    document.addEventListener('mousedown', onDocMouseDown);
+    return () => document.removeEventListener('mousedown', onDocMouseDown);
+  }, [open]);
+
+  const selectedItem = useMemo<ComboboxItemData | undefined>(() => {
+    const pool = asyncItems ?? staticItems;
```
zenith-oi, line 182: **FLAG** — [stale-closure-on-async-items] `selectedItem` resolves against `asyncItems ?? staticItems`, but `asyncItems` is only populated after the first successful `loadItems` call. For an async combobox with a pre-set `value` (e.g. `defaultValue='us'`), `selectedItem` will be `undefined` until the dropdown is opened and the initial query completes — so the input shows an empty string instead of the label of the pre-selected item. This affects the `readOnly` variant in stories too. Fix: when `loadItems` is provided and `value` is set but `asyncItems` is null, either hydrate on mount with an empty-query load, or accept a `selectedItemLabel` escape-hatch prop for SSR/pre-populated scenarios.

packages/ui/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,466 @@
...
+
+  const onKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (disabled) return;
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
+      if (!open) {
+        openAndPreparQuery();
+        return;
+      }
+      setHighlightIdx((i) => Math.min(i + 1, Math.max(0, visibleItems.length - 1)));
```
zenith-oi, line 231: **FLAG** — [keyboard-nav-skips-disabled-in-combobox] `ArrowDown` increments `highlightIdx` by 1 unconditionally, including onto disabled items. The ARIA combobox pattern requires that keyboard navigation skips `aria-disabled` options (same discipline the RadioGroup correctly applies). A user pressing ArrowDown on the last enabled item before a disabled one will land highlight on the disabled item and be unable to select it with Enter — confusing without any visual skip. Fix: implement a helper that walks `visibleItems` from the current index, skipping `item.disabled === true`, wrapping as needed.

zenith-oi (dismissed):
## Zenith Code Review

**Adds two new beta UI components — Combobox (searchable single-value select with async support) and RadioGroup — with stories, tests, and public exports. Implementation quality is high: clean ARIA wiring, roving tabindex, controlled/uncontrolled dual-mode, request-cancellation guard on async loads, and solid test coverage.**

**Positive:**
- Request-cancellation guard on async loads (`reqRef` integer generation) correctly prevents stale results from racing async calls — this is subtle and done right.
- RadioGroup's `focusByOffset` correctly skips disabled buttons by checking both `hasAttribute('disabled')` and `aria-disabled`, covering the dual path.
- The `onMouseDown` `e.preventDefault()` on listbox items to prevent input blur before click fires is the correct technique and is explicitly commented.
- Controlled/uncontrolled dual-mode is implemented cleanly with a stable `isControlled = controlledValue !== undefined` guard in both components.
- Hidden native `<input type='radio'>` with `tabIndex={-1}` and `aria-hidden` preserves native form submission without polluting the focus order — solid pattern.
- Test coverage is thorough: disabled items, readOnly, async loading with fake timers, ARIA attribute wiring, and keyboard navigation are all exercised.

---

packages/ui/src/components/RadioGroup/RadioGroup.tsx
zenith-oi, line 224: **FLAG** — [tabindex-all-unselected-tabbable] The roving tabindex logic `const tabIndex = isSelected ? 0 : ctx.value === null ? 0 : -1` makes every radio tabbable when nothing is selected (ctx.value === null). In that state all three or more radios get tabIndex=0, which breaks the roving-tabindex contract — only one element in the group should be in the tab sequence at a time. Fix: pick a single radio to be tabbable when nothing is selected (e.g. track a 'tabbable value' ref in the group, initialise it to the first registered radio's value, and use that for the null-selection case rather than making all radios tabbable).

packages/ui/src/components/Combobox/Combobox.tsx
zenith-oi, line 182: **FLAG** — [stale-closure-on-async-items] `selectedItem` resolves against `asyncItems ?? staticItems`. When `loadItems` is in use and the dropdown has not yet been opened (asyncItems is still null), the fallback is staticItems — but staticItems is empty when no `items` prop was passed and children composition is not used. This means a pre-set `value` / `defaultValue` will not display a label until the first async load resolves. More critically, after the listbox is closed and asyncItems are cleared on the next open cycle, the selected label momentarily disappears. Fix: maintain a `selectedItemCache` ref that is updated whenever a successful async load or item pick occurs, and fall back to that cache for label display when asyncItems is null.

packages/ui/src/components/Combobox/Combobox.tsx
zenith-oi, line 231: **FLAG** — [keyboard-nav-skips-disabled-in-combobox] ArrowDown/ArrowUp increment/decrement highlightIdx by 1 unconditionally, so keyboard navigation can land on and then commit a disabled item via Enter. The ARIA combobox pattern (APG) requires that keyboard navigation skips disabled options. Fix: after computing the candidate index, walk forward/backward until a non-disabled item is found (or stop at the boundary), similar to the skip-disabled logic already implemented in RadioGroup's focusByOffset.

zenith-oi (dismissed):
## Zenith Code Review

**Adds Combobox and RadioGroup components with full ARIA wiring, keyboard navigation, controlled/uncontrolled support, async loading, and comprehensive test coverage. Quality is high overall; three FLAG-level issues from the previous round remain unresolved.**

**Positive:**
- Race-condition guard on async loadItems using a request counter (reqRef) is a clean pattern that prevents stale responses from overwriting fresher results.
- mousedown preventDefault on listbox items to prevent input blur before click fires is a subtle but correct fix for a common combobox pitfall.
- RadioGroup's focusByOffset filters disabled elements from the navigation pool correctly, and the hidden native <input> for form submission is a thoughtful touch.
- Test coverage is thorough: ARIA attributes, keyboard navigation, disabled state, readOnly, async loading, empty state, and children composition are all exercised.

---

packages/ui/src/components/RadioGroup/RadioGroup.tsx
zenith-oi, line 224: **FLAG** — [tabindex-all-unselected-tabbable] The roving tabindex logic `const tabIndex = isSelected ? 0 : ctx.value === null ? 0 : -1` makes every radio tabbable when no value is selected (`ctx.value === null`). The ARIA roving tabindex pattern requires exactly one radio to be in the tab sequence at a time — the selected one, or only the first one when nothing is selected. With this implementation, tabbing into a group with no selection will tab through all radios instead of landing on the first and then using arrow keys. Fix: track which radio is the 'first registered' in the group (e.g. by index in the Map) and apply `tabIndex={0}` only to that one when `ctx.value === null`.

packages/ui/src/components/Combobox/Combobox.tsx
zenith-oi, line 231: **FLAG** — [keyboard-nav-skips-disabled-in-combobox] ArrowDown/ArrowUp increment/decrement `highlightIdx` unconditionally, so keyboard navigation can land on a disabled item and Enter will silently do nothing (the `onItemPick` guard returns early). This is confusing UX — users pressing ArrowDown repeatedly will appear to skip items. The fix is to advance `highlightIdx` in a loop, skipping items where `item.disabled === true`, matching the behavior of the RadioGroup's `focusByOffset` which already filters disabled items.

zenith-oi (approved):
## Zenith Code Review

**Adds Combobox and RadioGroup components to the design system with full ARIA wiring, keyboard navigation, async loading support, and comprehensive tests. Implementation quality is high overall. Three issues from the previous review remain unresolved.**

**Positive:**
- Race condition handling in the async loader is done correctly — `reqRef` counter ensures stale responses from superseded requests are discarded without introducing a state inconsistency.
- The `onMouseDown` preventDefault trick on listbox items to prevent input blur before the click fires is the correct ARIA combobox implementation pattern and is well-commented.
- Hidden native `<input type='radio'>` alongside the custom button element is a smart form-submission compatibility layer that keeps the accessible widget decoupled from native form semantics.
- ARIA wiring is thorough on both components: `aria-activedescendant`, `aria-controls`, `aria-autocomplete`, `aria-invalid`, `aria-describedby`, `aria-orientation` — all correctly wired and tested.
- Test coverage is broad and exercises the right edge cases: disabled item rejection, readOnly mode, async debounce with fake timers, empty state, and ARIA attribute assertions.

---

PR #74, opened 2026-05-04
feat(ui): Card consolidation — composable slots + 7 preset wrappers

apaksh-gupta:
## Summary

Audit **Tier B1** — consolidate the 8 card variants into one composable `<Card>` primitive plus thin opinionated preset wrappers. **Fully backward compatible** — every existing public prop on every wrapper still works.

### `<Card>` v2 — composable with named slots
- `<Card.Header>` (title / subtitle / icon / badge slots, or free-form children)
- `<Card.Media>` (image / video / iframe)
- `<Card.Body>` (main content — alias `Card.Content` kept for v1 compat)
- `<Card.Footer>` (meta info; renders top divider by default)
- `<Card.Actions>` (right-aligned buttons row)

Variants: `default | elevated | interactive | outlined | ghost`.
Padding: `compact | default | comfortable` (v2) **and** `none | sm | md | lg` (v1) — same classes.
Hover: `interactive | static`, auto-on when `onClick` or `as="button"`.
Polymorphic root via `as=`. When interactive, auto-attaches `role=button`, `tabIndex=0`, Enter/Space activation, focus-visible ring.

### 7 preset wrappers now compose `<Card>`
ContentTypeCard, ScriptPreviewCard, GoalCard, PackageCard, MetricCard, WalletCard, CollapsibleCard — each ~20–60 LOC, all original props preserved. Visual/behavioural Card upgrades flow through automatically.

### `component-status.json`
Added `replacedBy: "Card"` + explanatory `notes` on the 7 preset wrappers. They remain `status: "stable"` (not deprecated) so existing consumers see no churn — Pixel surfaces a "prefer the canonical primitive" hint when scaffolding new code. Schema updated to allow `replacedBy` on non-deprecated entries.

## Backward compat verification
- `tsc --noEmit`: same 20 pre-existing errors on main, no new ones from this PR.
- `vitest run`: **37/37 pass** (12 new Card + 5 new ContentTypeCard + 20 pre-existing).
- `npm run build`: tokens + ui + storybook all green. (mcp-server pre-existing failure on main — missing `@types/node`, unrelated.)
- Top-level v1 exports preserved: `CardHeader`, `CardTitle`, `CardDescription`, `CardContent`, `CardFooter` still importable from `@amplify-ai/ui`.
- All 7 preset wrappers preserve **exact** prop shapes.

## Test plan
[...]

PR #73, opened 2026-05-04
feat(ui): marketing primitives — CaseStudyCard, Marquee, IconCallout/Grid, FAQ, Footer, MediaShowcase

apaksh-gupta:
## Summary

Adds 6 marketing-page primitives to `@amplify-ai/ui` to fill the gaps the marketing audit identified between the existing Wave 1 marketing tier (Hero / CTABand / LogoCloud / PricingTable) and a full landing page. These are the **Tier C** primitives from that audit.

| Component | What it is |
|---|---|
| `<CaseStudyCard>` | Customer story card — logo, quote, 2-3 stat cards, attribution, optional CTA. Hoverable. Optional `href` makes the whole card a link. |
| `<Marquee>` | Continuously scrolling row of items (logos / testimonials / social proof). CSS animation only — no JS interval. `direction`, `speed`, `pauseOnHover`, `fadeEdges`. Children duplicated internally for a seamless `translateX(-50%)` loop. |
| `<IconCallout>` + `<IconGrid>` | 3- or 4-up feature blocks. `IconCallout` is one block; `IconGrid` lays out the row responsively (`columns={2 \| 3 \| 4}`, `gap={"default" \| "comfortable"}`). |
| `<FAQ>` | Accordion specifically for the FAQ pattern. `single` or `multi` mode. Renders `schema.org/FAQPage` JSON-LD inline for SEO automatically (toggle-able). |
| `<Footer>` | Composition with `Footer.Brand`, `Footer.LinkColumn`, `Footer.Social`, `Footer.Newsletter` (uses `Input` — no Form deps), `Footer.Legal`. 3/4/5-column grid. Uses `--amp-semantic-bg-sunken` so the footer reads visually heavier. |
| `<MediaShowcase>` | Hero-class media block — video / image / iframe with overlay text + CTA + optional play button. Auto-pauses video when scrolled out of view (IntersectionObserver). Supports caption tracks (`<track>` elements) for accessibility. |

All 6 ship at status `beta` since 2.6.0 in `component-status.json`.

## Constraints honoured

- **Canvas tokens only** — every colour and spacing uses `var(--amp-semantic-*)`. No hardcoded brand colours in component code.
- **TypeScript strict** — no `any`, narrow prop types, full prop documentation in JSDoc.
- **a11y**:
  - All primitives are keyboard-friendly (button focus rings, anchor focus rings, native `<button>`/`<a>` semantics throughout).
[...]

PR #72, opened 2026-05-04
feat(ui): Form scaffolding layer — Form, Field, Label, FieldError, FormSection

apaksh-gupta:
## Summary

Adds the missing **form scaffolding** tier to `@amplify-ai/ui`. Canvas had 87 components but no shared form layer — every product (Brand, Creator, Atmosphere) was reinventing form state, validation, and error display. This PR ships five composable primitives that back the platform on a single, schema-validated, fully accessible foundation.

Tier A1 of the Canvas design-system audit (`~/Desktop/canvas-design-system-audit.md`).

## What's in the box

| Component | Purpose | Key props |
|---|---|---|
| `<Form>` | Owns submission lifecycle. React state + zod-shape validation. **No** `react-hook-form`. | `schema?`, `defaultValues?`, `onSubmit({values, isValid})`, `validateOnSubmitOnly?` |
| `<Field>` | Slot composition: `<Label>` + input + hint + `<FieldError>`. Auto-wires `id`/`htmlFor`/`aria-describedby`/`aria-invalid`. | `name`, `label?`, `hint?`, `error?`, `required?`, `layout?` (`stacked`/`inline`), `children: (props) => ReactNode` |
| `<Label>` | Standalone accessible `<label>` for non-`<Field>` cases. | `required?` (renders aria-hidden asterisk), `emphasis?` (`primary`/`secondary`) |
| `<FieldError>` | Inline error with `role="alert"`, AlertCircle SVG, error semantic-token color. Renders nothing when empty. | `id?`, `children` |
| `<FormSection>` | Visual grouping for related fields. Optional collapsible variant (uses existing `<CollapsibleSection>` under the hood). | `title`, `description?`, `collapsible?`, `defaultOpen?` |

## Design decisions

- **No new runtime dep.** `Form#schema` is typed against a duck-typed `FormSchema` interface that exactly matches `z.ZodSchema['safeParse']`. Consumers can pass a real zod schema or any structurally compatible adapter (valibot wrapper, hand-rolled validator). Keeps the UI package free of zod's runtime cost while preserving the ergonomic API.
- **Slot pattern for `<Field>`.** Render-prop child receives `{id, name, 'aria-describedby', 'aria-invalid', 'aria-required', required}`. Spread onto any input primitive — `<Input>`, `<Select>`, `<Textarea>`, `<Checkbox>`, etc. Avoids tightly coupling the form layer to specific input variants.
[...]

PR #80, opened 2026-05-04
chore(canvas): commit contracts.json snapshot — Pixel needs this to see new components

apaksh-gupta:
## Summary

Pixel's \`canvas-contracts-loader\` fetches \`packages/ui/dist/contracts.json\` from origin/main via the GitHub Contents API. \`dist/\` is .gitignored so the file was never available — Pixel always fell back to a stale snapshot bundled in pixel-agent that was missing every component shipped in v2.5+.

## Impact

Without this PR, every Pixel-generated design today uses an outdated component vocabulary. Generations don't know about Form, Field, Banner, RadioGroup, Combobox, Card v2 slots, or any of the marketing primitives.

## What's in this PR

- Force-add \`packages/ui/dist/contracts.json\` (1676 lines, 105 components — all 87 stable + 18 newly shipped across PRs #70 / #71 / #72 / #73 / #74).

## Going forward

Future PRs to \`@amplify-ai/ui\` must regenerate this file (the \`prebuild\` script already does so) and include the diff in the PR.

## Test plan

- [ ] After merge: \`POST /api/tokens/refresh\` on Pixel (or wait 60 min for natural cache TTL)
- [ ] Verify Pixel-generated designs use the new components (e.g. ask Pixel to design a form using a CaseStudyCard)

### DIFF
--- a/packages/ui/dist/contracts.json
+++ b/packages/ui/dist/contracts.json
@@ -1,0 +1,1676 @@
+{
+  "generatedAt": "2026-05-04T12:51:21.428Z",
+  "tsVersion": "5.9.3",
+  "componentCount": 105,
+  "statusBreakdown": {
+    "stable": 46,
+    "alpha": 15,
+    "beta": 44
+  },
+  "components": [
+    {
+      "name": "ActionFooter",
+      "contract": "contracts/ActionFooter.json",
+      "filePath": "packages/ui/src/components/ActionFooter/ActionFooter.tsx",
+      "kind": "forwardRef",
+      "variants": null,
+      "sizes": null,
+      "propCount": 7,
+      "subcomponents": [],
+      "lifecycle": {
+        "status": "stable",
+        "since": "1.0.0"
+      }
+    },
+    {
+      "name": "AddonTeaser",
+      "contract": "contracts/AddonTeaser.json",
+      "filePath": "packages/ui/src/components/AddonTeaser/AddonTeaser.tsx",
+      "kind": "forwardRef",
+      "variants": null,
+      "sizes": null,
+      "propCount": 2,
+      "subcomponents": [],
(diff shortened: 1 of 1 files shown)


packages/ui/dist/contracts.json
```diff
@@ -0,0 +1,1676 @@
+{
```
zenith-oi, line 1: **FLAG** — [build-artifacts-in-vcs] This file lives under `dist/`, which is conventionally a build output directory and is typically gitignored. Committing generated build artifacts into version control creates maintenance burden: the file will silently drift from source (components added/removed without updating this index), and every regeneration produces a noisy diff. The `generatedAt` timestamp (L2) will change on every build, polluting git history.

If this file is intentionally committed as a published artifact (e.g., for consumers who install the npm package to read the contract index at runtime), the intent should be documented and the generation step should be part of the build pipeline with CI enforcing freshness. If it is purely a build output, add `packages/ui/dist/` to `.gitignore` and let the publish step handle it. Either way, clarify the intent — right now it is ambiguous whether this is a curated contract or an auto-generated snapshot.

zenith-oi (approved):
## Zenith Code Review

**This PR adds a generated `contracts.json` index file to `packages/ui/dist/`, enumerating 105 UI components with their metadata. The file is clean data with no security or logic concerns, but it belongs in `dist/` which is typically a build artifact directory that should be gitignored rather than committed.**

**Positive:**
- The deprecation/replacement metadata (`replacedBy`, `notes`) on legacy card-shaped components (CollapsibleCard, ContentTypeCard, GoalCard, MetricCard, PackageCard, ScriptPreviewCard, WalletCard) is a genuinely useful signal — it communicates migration intent directly in the machine-readable contract without removing the components, which is the correct backwards-compatible approach.

---

PR #112, opened 2026-05-18
feat(sdui-runtime): runtime foundation — SduiNode, Interpreter, registries, providers

mridulgupta-oi:
## Summary
- **Brief #264, Task 11** — SDUI runtime foundation package (`@amplify-ai/sdui-runtime`)
- Core `SduiNode` wrapper: Zod schema validation, `SduiErrorBoundary`, Clickable/Viewable HOCs, lifecycle triggers (on_load, on_view, on_dismount)
- Recursive `Interpreter` for node→renderer dispatch via component registries
- `PageRoot` for page layout dispatch via page container registry
- Empty registries (ui-components, snippets, pages) — populated by Tasks 24-26
- `SduiRuntimeProvider`, `IconStoreProvider`, `BottomSheetHost` providers
- Zustand stores: form state, section state, bottom-sheet, BFF client, navigation stack
- CI workflow updated for `sdui-runtime` package

## Test plan
- [ ] `npm run build` passes for sdui-runtime package
- [ ] TypeScript strict-mode type check passes
- [ ] SduiNode renders children when valid schema data provided
- [ ] SduiErrorBoundary catches and displays fallback on invalid data
- [ ] Interpreter dispatches to correct renderer per node type

Closes #264

### DIFF
--- a/packages/sdui-runtime/package.json
+++ b/packages/sdui-runtime/package.json
@@ -35,6 +35,7 @@
     "zod": "^3.22.0"
   },
   "peerDependenciesMeta": {
+    "@one-impression/sdk-native-sdui": { "optional": true },
     "@gorhom/bottom-sheet": { "optional": true }
   },
   "devDependencies": {
--- a/packages/sdui-runtime/package.json
+++ b/packages/sdui-runtime/package.json
@@ -1,0 +1,42 @@
+{
+  "name": "@amplify-ai/sdui-runtime",
+  "version": "1.0.0",
+  "description": "SDUI runtime for Amplify Creator App — interpreter, action engine, bottom-sheet manager, loaders, providers",
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "default": "./dist/index.js"
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup && tsc --emitDeclarationOnly --noCheck",
+    "dev": "tsup --watch",
+    "lint": "eslint src/",
+    "test": "jest"
+  },
+  "peerDependencies": {
+    "react": ">=18.0.0",
+    "react-native": ">=0.72.0",
+    "@one-impression/sdk-native-sdui": ">=1.0.0",
+    "@amplify-ai/ui-native": ">=1.0.0",
+    "@amplify-ai/tokens-creator": ">=2.0.3"
+  },
+  "dependencies": {
+    "@gorhom/bottom-sheet": "^5.0.0",
+    "zustand": "^4.5.0",
+    "@tanstack/react-query": "^5.0.0",
+    "zod": "^3.22.0"
+  },
+  "devDependencies": {
+    "@types/react": "^18.2.0",
+    "tsup": "^8.0.0",
+    "typescript": "^5.5.0"
+  }
+}
--- a/packages/sdui-runtime/package.json
+++ b/packages/sdui-runtime/package.json
@@ -26,16 +26,18 @@
     "react-native": ">=0.72.0",
     "@one-impression/sdk-native-sdui": ">=1.0.0",
     "@amplify-ai/ui-native": ">=1.0.0",
-    "@amplify-ai/tokens-creator": ">=2.0.3"
+    "@amplify-ai/tokens-creator": ">=2.0.3",
+    "@gorhom/bottom-sheet": "^5.0.0"
   },
   "dependencies": {
-    "@gorhom/bottom-sheet": "^5.0.0",
     "zustand": "^4.5.0",
     "@tanstack/react-query": "^5.0.0",
     "zod": "^3.22.0"
   },
+  "peerDependenciesMeta": {
+    "@gorhom/bottom-sheet": { "optional": true }
+  },
   "devDependencies": {
-    "@types/react": "^18.2.0",
     "tsup": "^8.0.0",
     "typescript": "^5.5.0"
   }
--- a/packages/sdui-runtime/src/action-engine/action-engine.ts
+++ b/packages/sdui-runtime/src/action-engine/action-engine.ts
@@ -1,0 +1,29 @@
+import type { Action } from "@one-impression/sdk-native-sdui";
+import type { ActionEngine, ActionEngineConfig, ActionHandler } from "./types.js";
+import { actionHandlerRegistry } from "../registries/actions.js";
+import { capabilityHandlerRegistry } from "../registries/capabilities.js";
+
+export function createActionEngine(config: ActionEngineConfig): ActionEngine {
+  const dispatch = (action: Action): void => {
+    const { type } = action;
+
+    // Capability actions: type starts with "capability:"
+    if (type.startsWith("capability:")) {
+      const capName = type.slice("capability:".length);
+      const handler = capabilityHandlerRegistry[capName];
+      if (handler) {
+        handler(action, config);
+      }
+      return;
+    }
+
+    // Standard action verbs
+    const handler: ActionHandler | undefined = actionHandlerRegistry[type];
+    if (handler) {
+      handler(action, config);
+    }
+    // Unknown action type — silent skip per forward-compat contract
+  };
+
+  return { dispatch };
+}
--- a/packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
+++ b/packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
@@ -1,0 +1,28 @@
+import { useState, useCallback } from "react";
+
+interface FormSubmissionState {
+  loading: boolean;
+  submit: (fn: () => Promise<void>) => Promise<void>;
+}
+
+/**
+ * Manages loading state during async form submissions.
+ * Prevents double-submit by guarding with the loading flag.
+ * Ported 1:1 from legacy.
+ */
+export function useFormSubmissionLoading(): FormSubmissionState {
+  const [loading, setLoading] = useState(false);
+
+  const submit = useCallback(async (fn: () => Promise<void>) => {
+    if (loading) return;
+    setLoading(true);
+    try {
+      await fn();
+    } finally {
+      setLoading(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loading]);
+
+  return { loading, submit };
+}
--- a/packages/sdui-runtime/src/registries/actions.ts
+++ b/packages/sdui-runtime/src/registries/actions.ts
@@ -1,0 +1,9 @@
+import type { ActionHandler } from "../action-engine/types.js";
+
+/**
+ * Registry mapping action verb strings to handler functions.
+ * Populated by task 023 (sdui-action-handlers).
+ *
+ * Shape: { "navigate": navigateHandler, "bff_call": bffCallHandler, ... }
+ */
+export const actionHandlerRegistry: Record<string, ActionHandler> = {};
--- a/packages/sdui-runtime/src/sdui-node/SduiNode.tsx
+++ b/packages/sdui-runtime/src/sdui-node/SduiNode.tsx
@@ -1,0 +1,85 @@
+import React, { useEffect, useCallback } from "react";
+import type { ReactNode } from "react";
+import type { z } from "zod";
+import type { Action } from "@one-impression/sdk-native-sdui";
+import { SduiErrorBoundary } from "./SduiErrorBoundary.js";
+import { SduiFallback } from "./SduiFallback.js";
+import { Clickable } from "../clickable/Clickable.js";
+import { Viewable } from "../viewable/Viewable.js";
+import { useActionEngine } from "../action-engine/useActionEngine.js";
+import { useTelemetry } from "../telemetry/useTelemetry.js";
+
+interface TrackEvent {
+  name: string;
+  params?: Record<string, unknown>;
+}
+
+interface SduiNodeProps<TSchema extends z.ZodTypeAny> {
+  data: unknown;
+  schema: TSchema;
+  id: string;
+  on_click?: Action;
+  on_load?: Action;
+  on_view?: Action;
+  on_dismount?: Action;
+  view_events?: TrackEvent[];
+  load_events?: TrackEvent[];
+  children: (validated: z.infer<TSchema>) => ReactNode;
+}
+
+export function SduiNode<TSchema extends z.ZodTypeAny>(
+  props: SduiNodeProps<TSchema>,
+): React.ReactElement {
+  const validated = props.schema.parse(props.data);
+  const actionEngine = useActionEngine();
+  const telemetry = useTelemetry();
+
+  useEffect(() => {
+    telemetry.emit("sdui.node.rendered", { id: props.id });
+
+    if (props.on_load) {
+      actionEngine.dispatch(props.on_load);
+    }
+    if (props.load_events) {
+      for (const e of props.load_events) {
+        telemetry.emit(e.name, e.params);
+      }
+    }
+
+    return () => {
+      if (props.on_dismount) {
+        actionEngine.dispatch(props.on_dismount);
+      }
+    };
+    // Intentionally run only on mount/unmount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const handleViewed = useCallback(() => {
+    if (props.on_view) {
+      actionEngine.dispatch(props.on_view);
+    }
+    if (props.view_events) {
+      for (const e of props.view_events) {
+        telemetry.emit(e.name, e.params);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return (
+    <SduiErrorBoundary nodeId={props.id} fallback={<SduiFallback />}>
+      <Viewable onView={handleViewed}>
+        <Clickable
+          onPress={
+            props.on_click
+              ? () => actionEngine.dispatch(props.on_click!)
+              : undefined
+          }
+        >
+          {props.children(validated)}
+        </Clickable>
+      </Viewable>
+    </SduiErrorBoundary>
+  );
+}
--- a/.changeset/sdui-runtime-foundation.md
+++ b/.changeset/sdui-runtime-foundation.md
@@ -1,0 +1,17 @@
+---
+"@amplify-ai/sdui-runtime": major
+---
+
+feat(sdui-runtime): add @amplify-ai/sdui-runtime@1.0.0 — SDUI runtime foundation
+
+New package providing the foundation layer for the SDUI runtime in amplify-creator-app:
+- SduiNode base wrapper (Zod validation, error boundary, click/view/load/dismount lifecycle, telemetry)
+- Interpreter dispatcher (type-based renderer lookup with forward-compat fallback)
+- PageRoot page container dispatcher
+- Clickable and Viewable HOCs
+- Action engine (pluggable verb handlers, capability dispatch)
+- Bottom-sheet manager (Zustand-driven, stack depth=2, sheet-aware context for gorhom v5+ scrollable/input swaps)
+- 7 hooks ported from legacy (useAppStateSession, useBottomSheetData, useBottomSheetFormSync, useFormSubmissionLoading, useHydrateParams, useKeyboardStatus, usePageRefresh)
+- 6 skeleton loaders + ContainerLoader + 3-tier resolution (action hint → endpoint hint → default)
+- Registry shells for ui-components, snippets, pages, actions, capabilities (populated by tasks 023-026)
+- SduiRuntimeProvider root provider
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,12 @@ jobs:
       - name: Build @amplify-ai/ui
         run: npm run build -w packages/ui
 
+      - name: Build @amplify-ai/ui-native
+        run: npm run build -w packages/ui-native
+
+      - name: Build @amplify-ai/sdui-runtime
+        run: npm run build -w packages/sdui-runtime
+
       - name: Build @amplify-ai/mcp-server
         run: npm run build -w packages/mcp-server
 
@@ -53,7 +59,7 @@ jobs:
       - name: Verify build outputs
         run: |
           MISSING=""
-          for pkg in tokens-foundation tokens-brand tokens-creator tokens-atmosphere tokens-marketing ui mcp-server; do
+          for pkg in tokens-foundation tokens-brand tokens-creator tokens-atmosphere tokens-marketing ui ui-native sdui-runtime mcp-server; do
             DIR="packages/$pkg/dist"
             if [ ! -d "$DIR" ]; then
               MISSING="$MISSING\n  - packages/$pkg/dist/ (directory missing)"
@@ -81,7 +87,7 @@ jobs:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: |
           set -e
-          for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui mcp-server eslint-config; do
+          for pkg in tokens-foundation tokens-brand tokens-atmosphere tokens-creator ui ui-native sdui-runtime mcp-server eslint-config; do
             echo "Publishing @amplify-ai/$pkg..."
             (
               cd packages/$pkg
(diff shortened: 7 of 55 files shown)


packages/sdui-runtime/src/registries/actions.ts
```diff
@@ -0,0 +1,9 @@
+import type { ActionHandler } from "../action-engine/types.js";
+
+/**
+ * Registry mapping action verb strings to handler functions.
+ * Populated by task 023 (sdui-action-handlers).
```
zenith-oi, line 5: **FLAG** — [no-task-refs-in-durable-artifacts] Comment references `task 023 (sdui-action-handlers)` — a transient task ordinal that will be meaningless once Brief #264 closes. Same pattern appears in `capabilities.ts` (L5), `pages.ts` (L6), `snippets.ts` (L6), `ui-components.ts` (L6), and the changeset file (L8). Replace with timeless descriptions of what the registry is for, e.g. 'Populated by the action-handlers package in subsequent releases.' Task numbers belong in commit messages and PR descriptions, not in durable source files.

packages/sdui-runtime/src/action-engine/action-engine.ts
```diff
@@ -0,0 +1,29 @@
...
+      }
+      return;
+    }
+
+    // Standard action verbs
+    const handler: ActionHandler | undefined = actionHandlerRegistry[type];
+    if (handler) {
+      handler(action, config);
+    }
+    // Unknown action type — silent skip per forward-compat contract
```
zenith-oi, line 25: **FLAG** — [silent-error-handling] Unknown action types are silently skipped. While the comment says 'forward-compat contract', this means mistyped or misconfigured action verbs will fail without any signal in dev/test. At minimum, emit a telemetry event or `console.warn` in non-production environments so contract violations surface during development before BFF action documents reach production.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
```diff
@@ -0,0 +1,85 @@
...
+  on_dismount?: Action;
+  view_events?: TrackEvent[];
+  load_events?: TrackEvent[];
+  children: (validated: z.infer<TSchema>) => ReactNode;
+}
+
+export function SduiNode<TSchema extends z.ZodTypeAny>(
+  props: SduiNodeProps<TSchema>,
+): React.ReactElement {
+  const validated = props.schema.parse(props.data);
```
zenith-oi, line 33: **FLAG** — [missing-error-handling] `props.schema.parse(props.data)` is called outside the `SduiErrorBoundary` — it is called in the function body before the JSX is returned, so a Zod parse error will throw synchronously and propagate up past the error boundary wrapping in the return value. The error boundary only catches errors thrown during render of the tree it wraps. Move the `schema.parse()` call inside the render tree (e.g. render a wrapper component that performs validation internally) or wrap the parse in a try/catch and return the fallback directly on failure.

packages/sdui-runtime/package.json
```diff
@@ -0,0 +1,42 @@
...
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup && tsc --emitDeclarationOnly --noCheck",
+    "dev": "tsup --watch",
+    "lint": "eslint src/",
+    "test": "jest"
```
zenith-oi, line 22: **FLAG** — [missing-tests] The package declares `"test": "jest"` but no jest configuration, no test files, and no jest dependency exist in this PR. The action engine dispatch logic, Zustand store stack depth enforcement, loader resolution tiers, and Zod parse error path in SduiNode are all critical paths with no test coverage. At minimum, unit tests for `useBottomSheetStore` (stack depth cap, close/replace), `resolveLoader` (3-tier resolution), and `action-engine` dispatch routing should accompany this foundation PR.

packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  const submit = useCallback(async (fn: () => Promise<void>) => {
+    if (loading) return;
+    setLoading(true);
+    try {
+      await fn();
+    } finally {
+      setLoading(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
zenith-oi, line 24: **FLAG** — [missing-error-handling] The `eslint-disable-next-line react-hooks/exhaustive-deps` suppression on `[loading]` is masking a real bug: `loading` closes over the state value at the time `useCallback` was created, so the guard `if (loading) return` in the callback captures the stale value. Since `loading` IS the guarding variable, it must be in deps — or the pattern must be restructured to use a `useRef` for the guard and `useState` only for the exposed `loading` value. The current code does not reliably prevent double-submit when `submit` is called multiple times in rapid succession.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces the `@amplify-ai/sdui-runtime` package (step 11 in Brief #264), establishing the SDUI runtime foundation for the Amplify Creator App. The code is well-structured, clearly authored, and represents a solid foundation layer with sensible abstractions. A few real issues worth addressing before merge.**

**5 FLAGs** — advisory, not merge-blocking. See inline comments.

**Positive:**
- The Zustand bottom-sheet store stack depth cap (MAX_STACK_DEPTH=2) is enforced at push time with a clear warning — correct defensive design for a shared singleton.
- SduiErrorBoundary isolates per-node failures cleanly — one bad node cannot crash the full page tree.
- The 3-tier loader resolution (action hint → endpoint hint → default) is a clean, extensible pattern and is easy to extend without touching callers.
- Registry shells are correctly typed and empty — downstream tasks (024-026) have a well-typed surface to populate without changing the public API.
- Forward-compat Fallback on unknown SDUI node types is the right contract — renders nothing rather than crashing on nodes from a newer BFF schema version.
- BottomSheetContext with insideSheet flag for gorhom v5 ScrollView/FlatList/TextInput swaps is a clean solution to the documented gorhom v5 requirement.
- ActionEngineContext and TelemetryContext both provide noop defaults — components are safe without a provider, good for unit testing isolation.

---

packages/sdui-runtime/src/registries/actions.ts
zenith-oi, line 5: **FLAG** — [no-task-refs-in-durable-artifacts] Comment still reads `Populated by task 023 (sdui-action-handlers)` — a transient task ordinal that will be meaningless after Brief #264 closes. Same pattern exists in capabilities.ts (line 5), pages.ts (line 6), snippets.ts (line 6), and ui-components.ts (line 6). Replace with a domain-descriptive comment, e.g. `Populated when the action-handler package registers its verbs at startup.` Commit messages are the right place for task provenance.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [missing-error-handling] `props.schema.parse(props.data)` throws a `ZodError` synchronously in the function body, before the `SduiErrorBoundary` in the JSX return — React error boundaries only catch errors thrown during rendering (i.e. from child component render calls), not from the parent's render-phase function body before the JSX tree is constructed. A parse failure here will propagate up uncaught. Wrap the parse in a try/catch and either return `<SduiFallback />` directly, or move the parse inside the children render callback so it happens within the boundary's subtree.

packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Manages loading state during async form submissions.
+ * Prevents double-submit by guarding with the loading flag.
+ * Ported 1:1 from legacy.
+ */
+export function useFormSubmissionLoading(): FormSubmissionState {
+  const [loading, setLoading] = useState(false);
+
+  const submit = useCallback(async (fn: () => Promise<void>) => {
```
zenith-oi, line 16: **FLAG** — [missing-error-handling] The `eslint-disable-next-line react-hooks/exhaustive-deps` suppression on `[loading]` hides a real bug: `loading` captured in the `useCallback` closure will always be the stale value from the render when the callback was memoized. Concurrent rapid calls can bypass the double-submit guard because the captured `loading` is `false`. Fix: use a `useRef` for the in-flight guard (`const inFlight = useRef(false)`) instead of reading the state variable inside the callback, which avoids the stale-closure issue and removes the need for the eslint suppression.

zenith-oi (dismissed):
## Zenith Code Review

**This PR ships the `@amplify-ai/sdui-runtime` foundation package — interpreter, action engine, bottom-sheet manager, hooks, skeleton loaders, and registry shells. The implementation is well-structured and clear. Three issues from the previous review remain unresolved.**

**Positive:**
- The `SduiErrorBoundary` per-node isolation design is correct — one bad node renders `SduiFallback` without crashing the page tree.
- Zustand bottom-sheet store with MAX_STACK_DEPTH=2 guard and explicit `replace` action is clean and forward-safe.
- The three-tier loader resolution (action hint → endpoint hint → default) is a well-structured and extensible approach.
- Noop defaults for both `ActionEngineContext` and `TelemetryContext` mean the runtime degrades gracefully when providers are absent — good defensive design.
- Forward-compat silent-skip on unknown action types is correctly documented with an explicit comment rather than left as an unexplained empty branch.

---

packages/sdui-runtime/src/registries/actions.ts
zenith-oi, line 5: **FLAG** — [no-task-refs-in-durable-artifacts] Comment references `task 023 (sdui-action-handlers)` — a transient task ordinal that will be meaningless once Brief #264 closes. Replace with a description of what populates this registry and why: e.g. 'Populated by the action-handler package — maps action verb strings like "navigate" and "bff_call" to handler functions.' Same applies in capabilities.ts, snippets.ts, ui-components.ts, and pages.ts which carry similar task-reference comments.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [missing-error-handling] `props.schema.parse(props.data)` is called in the function body before the JSX return, placing it outside the `SduiErrorBoundary` that wraps the rendered output. A Zod `ZodError` thrown here will propagate up to the nearest error boundary in the *parent* tree, not to the `SduiErrorBoundary` defined on line 71. Wrap the parse inside the children render callback, or use `safeParse` and handle the error case explicitly: `const result = props.schema.safeParse(props.data); if (!result.success) return <SduiFallback />;` — this keeps the error contained to the node's own fallback.

packages/sdui-runtime/package.json
```diff
@@ -0,0 +1,44 @@
...
```
zenith-oi, line 22: **FLAG** — [missing-tests] The package declares `"test": "jest"` but there is no jest configuration, no `jest` dependency (not in dependencies, devDependencies, or peerDependencies), and no test files anywhere in this PR. The CI script at `.github/workflows/ci.yml` builds the package but doesn't run tests. Either add jest + tests, or change the script to `"test": "echo 'no tests yet'"` to avoid a confusing silent failure when someone runs `npm test`.

packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
zenith-oi, line 16: **FLAG** — [missing-error-handling] The `useCallback` dependency array contains `[loading]`, which is suppressed by `// eslint-disable-next-line react-hooks/exhaustive-deps`. This creates a stale-closure bug: the `submit` function captured at render time closes over the `loading` value at that render, but since `loading` IS in the deps array (just suppressed from lint), a new `submit` is created every time `loading` changes — meaning the guard `if (loading) return;` inside the callback checks the captured value, not the current state. The fix is to use a ref for the loading guard: `const loadingRef = useRef(false)` updated alongside the state, and check `loadingRef.current` inside the callback, keeping the deps array as `[]`.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces the `@amplify-ai/sdui-runtime` package — the SDUI runtime foundation for the Amplify Creator App, including the action engine, bottom-sheet manager, interpreter, skeleton loaders, hooks, and registries. The code is well-structured and clearly purposeful. Several previously flagged issues remain unaddressed.**

**Positive:**
- The eslint-disable comment on the mount-only useEffect in SduiNode.tsx is correctly annotated with an explicit intent comment — this is the right pattern for intentional exhaustive-deps suppressions.
- The MAX_STACK_DEPTH guard in useBottomSheetStore with a console.warn is a solid defensive pattern for preventing runaway sheet stacking.
- Viewable.tsx correctly uses a ref (`hasFired`) to ensure the onView callback fires exactly once, and the comment acknowledging the onLayout proxy limitation is honest and actionable.
- The noopEngine and noopEmitter defaults on ActionEngineContext and TelemetryContext mean the runtime degrades gracefully when the provider is absent — good defensive default.
- The 3-tier loader resolution (action hint → endpoint hint → default) is clean and the fallback to the default registry entry is safe.

---

packages/sdui-runtime/src/registries/actions.ts
zenith-oi, line 5: **FLAG** — [no-task-refs-in-durable-artifacts] Comments reference 'task 023 (sdui-action-handlers)', 'task 024 (sdui-ui-component-renderers)', 'task 025 (sdui-snippet-renderers)', and 'task 026 (sdui-page-renderers)' across the registry files (actions.ts, capabilities.ts, ui-components.ts, snippets.ts, pages.ts). These task ordinals are transient and will be meaningless once Brief #264 closes. Replace with domain-meaningful descriptions, e.g. '// Populated by the sdui-action-handlers PR — see action-engine/types.ts for the ActionHandler contract.' Applies identically to all five registry files.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [missing-error-handling] `props.schema.parse(props.data)` is called unconditionally in the function body before the `SduiErrorBoundary` is rendered. A Zod `ZodError` thrown here will propagate up through the call stack before the boundary has a chance to catch it — error boundaries only catch errors that originate in a child component's render cycle, not in the parent's render function before JSX is returned. Wrap the parse in a try/catch and either throw inside the returned JSX (via a child that throws), or move to a useMemo/useState pattern so the throw happens inside the boundary's subtree. Simplest fix: `const validated = useMemo(() => props.schema.parse(props.data), [props.data, props.schema]);` — a parse error inside useMemo propagates as a render error and will be caught by the boundary.

packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
zenith-oi, line 16: **FLAG** — [missing-error-handling] The `useCallback` dependency array is `[loading]` (suppressed via eslint-disable), but `loading` is read from state inside the callback. This is a stale-closure bug: the `submit` reference captured at render time closes over the `loading` value at that render, but subsequent state updates to `loading` produce a new `submit` reference — any consumer that cached the old reference will see the stale `loading` flag and the double-submit guard will fail. The correct fix is to use a ref for the in-flight guard: `const inFlightRef = useRef(false)` and check/set `inFlightRef.current` instead of the state flag. The `loading` state remains for UI display; the guard uses the ref. Remove the eslint suppression.

packages/sdui-runtime/package.json
zenith-oi, line 22: **FLAG** — [missing-tests] The package declares `"test": "jest"` but ships with no jest configuration, no test files, and no jest dependency in devDependencies. CI will call `npm test` and fail immediately. Either add a jest config + at least a smoke test for the core paths (SduiNode schema validation, action engine dispatch, bottom-sheet store open/close/stack-depth guard), or replace the script with `"test": "echo 'No tests yet' && exit 0"` as an explicit placeholder. The former is strongly preferred given the brief's success criteria around a production-ready runtime.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces the `@amplify-ai/sdui-runtime` package — the foundation layer for SDUI rendering in the Amplify Creator App, including the interpreter, action engine, bottom-sheet manager, skeleton loaders, and hooks. The implementation is clean and well-structured for a new package; previous-round issues reviewed below.**

**Positive:**
- The `useBottomSheetStore` correctly enforces `MAX_STACK_DEPTH = 2` with a `console.warn` rather than silently dropping — good defensive guard for the stacking edge case.
- Silent skip for unknown action types in `action-engine.ts` is correctly documented as the forward-compat contract, matching SDUI's additive-only evolution model.
- Skeleton loaders use static `StyleSheet.create` objects rather than inline style objects — correct for React Native performance.
- The `SduiErrorBoundary` correctly uses a class component (the only React API available for error boundary semantics), with `componentDidCatch` logging the node ID for observability.
- The `Viewable` component correctly uses a `hasFired` ref (not state) to guard the one-time view callback, avoiding re-render loops.
- Registry shells are typed correctly (`Record<string, ComponentType<Node>>` etc.) and carry clear TODO comments pointing to the populating PRs — good scaffolding discipline.

---

packages/sdui-runtime/src/registries/actions.ts
zenith-oi, line 5: **FLAG** — [no-task-refs-in-durable-artifacts] Comments in actions.ts, capabilities.ts, pages.ts, snippets.ts, and ui-components.ts all reference task ordinals (e.g. 'task 023 (sdui-action-handlers)', 'task 024 (sdui-ui-component-renderers)', 'task 025 (sdui-snippet-renderers)', 'task 026 (sdui-page-renderers)'). These are transient SDLC identifiers that will rot once the brief closes. Replace with a timeless description of what will populate the registry, e.g. 'Populated by the action-handlers package when sdui-action-handlers ships — expected verbs: navigate, bff_call, toast, deeplink, bottom_sheet_open, bottom_sheet_close.' Same fix needed in capabilities.ts (L5), pages.ts (L6), snippets.ts (L7), ui-components.ts (L7).

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [missing-error-handling] `props.schema.parse(props.data)` is called unconditionally in the function body before the `<SduiErrorBoundary>` is returned. If `parse()` throws a ZodError (e.g. malformed server data), it will propagate up past the error boundary that is supposed to contain exactly this failure, potentially crashing the parent tree. Fix: wrap the parse in a try/catch and return `<SduiFallback />` directly on parse failure, or use `safeParse` and short-circuit to `<SduiFallback />` before entering the render path.

packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
zenith-oi, line 16: **FLAG** — [missing-error-handling] The `useCallback` has `[loading]` in its dependency array (suppressed with eslint-disable), but `loading` is read from state inside the callback only for the double-submit guard (`if (loading) return`). Because the callback is recreated on every `loading` state change, any component that memoises a button's `onPress` using this `submit` reference will work correctly — but the eslint-disable comment signals this was a known tradeoff. The deeper issue: if the passed `fn` throws synchronously (not just rejects), `setLoading(false)` in the `finally` block will still run, but React may log an unhandled error. The callback does not propagate the error to the caller, so submission failures are silently swallowed. Either re-throw after the `finally`, or document explicitly that callers must handle errors inside `fn`.

packages/sdui-runtime/package.json
```diff
@@ -0,0 +1,45 @@
...
```
zenith-oi, line 22: **FLAG** — [missing-tests] The package declares `"test": "jest"` but ships with no jest config, no test files, and no jest/testing-library dependency. As a foundation package with non-trivial logic (Zod validation in SduiNode, bottom-sheet stack depth enforcement, 3-tier loader resolution, action engine dispatch routing), missing tests means regressions in successor tasks (023-026) will be hard to attribute. At minimum, add tests for the action engine dispatch path, the bottom-sheet MAX_STACK_DEPTH guard, and the loader resolver's 3-tier priority.

zenith-oi (dismissed):
## Zenith Code Review

**This PR ships the `@amplify-ai/sdui-runtime` foundation package for the SDUI rebuild (Brief #264, step 11) — interpreter, action engine, bottom-sheet manager, skeleton loaders, hooks, and registry shells. The implementation is clean and well-structured for a foundation layer; three issues from the previous round remain unresolved.**

**Positive:**
- The forward-compatibility contract is implemented correctly throughout: unknown node types fall back silently to <Fallback>, unknown action verbs are skipped with a comment explaining the intent, and unknown page layouts degrade to an empty fragment rather than crashing.
- The bottom-sheet Zustand store correctly enforces the stack depth=2 limit with a console.warn rather than a silent drop or a throw, and all four operations (open/close/closeAll/replace) are consistent.
- The SduiErrorBoundary isolation is well-designed — per-node boundaries mean one bad renderer cannot crash the page tree.
- BottomSheetContext.Provider wrapping inside BottomSheetScrollView correctly signals to child renderers that they must swap RN primitives for gorhom-aware equivalents — the architecture comment in BottomSheetContext.tsx makes this intent clear.
- All peer dependencies (react-native, @gorhom/bottom-sheet, @one-impression/sdk-native-sdui) are correctly marked as external in tsup.config.ts, and optional ones are correctly flagged in peerDependenciesMeta.

---

packages/sdui-runtime/src/registries/actions.ts
zenith-oi, line 5: **FLAG** — [no-task-refs-in-durable-artifacts] Task ordinal references persist across the registry files (actions.ts, capabilities.ts, pages.ts, snippets.ts, ui-components.ts) — e.g. 'Populated by task 023 (sdui-action-handlers)', 'task 024', 'task 025', 'task 026'. These are transient identifiers; the comment should describe what populates the registry in domain terms (e.g. 'Populated by the action-handler package when action verbs are registered') rather than by task number. Rewrite all five occurrences to be timeless.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [missing-error-handling] `props.schema.parse(props.data)` is called unconditionally in the function body before the SduiErrorBoundary wraps the render output. A Zod parse error thrown here will propagate up to the nearest React error boundary above SduiNode — the per-node SduiErrorBoundary defined inside SduiNode's own render output never sees it. Use `safeParse` and return `<SduiFallback />` directly on failure, or wrap the parse in a try/catch that returns the fallback — this is the same structural issue from the previous round and remains unresolved.

packages/sdui-runtime/src/hooks/useFormSubmissionLoading.ts
zenith-oi, line 16: **FLAG** — [missing-error-handling] The `useCallback` for `submit` has `[loading]` in its dependency array with an eslint-disable suppression. The stale-closure guard (`if (loading) return`) reads `loading` from the closure captured at creation time, so after the first call the callback is recreated when `loading` flips to true (guarding against double-submit) but the reference held by callers is already the old one. Use a ref (`const loadingRef = useRef(false)`) for the guard instead, which removes the need for `loading` in the deps array and eliminates the eslint-disable. The current approach may let double-submits through if the callback reference is stale at the call site.

packages/sdui-runtime/package.json
zenith-oi, line 22: **FLAG** — [missing-tests] The package declares `"test": "jest"` but ships with no jest config, no test files, and no jest/testing-library in devDependencies. The action engine dispatch logic, bottom-sheet store state transitions (stack depth cap, close-by-id vs pop), loader resolution tiers, and useFormSubmissionLoading's double-submit guard are all critical paths that should have unit tests. At minimum a jest config and a smoke test should land with this foundation package.

zenith-oi (approved):
## Zenith Code Review

**This PR introduces the @amplify-ai/sdui-runtime package — the SDUI runtime foundation layer for Brief #264's creator app rebuild. The implementation is clean and well-structured for a foundation package. Three previous-round issues remain unresolved.**

**Positive:**
- The forward-compat silent-skip on unknown action types in action-engine.ts is the right contract for an SDUI runtime — unknown nodes/actions must never crash the tree.
- The Zustand bottom-sheet store's MAX_STACK_DEPTH guard with a console.warn (not a throw) is the correct defensive behaviour for a UI runtime.
- SduiErrorBoundary wrapping each node so a single bad node cannot crash the page is solid defensive architecture.
- The 3-tier loader resolution (action hint → endpoint hint → default) is cleanly abstracted in loader-resolver.ts and straightforward to extend.
- Externalising all handler registries as empty shells (actions, capabilities, pages, snippets, ui-components) with clear population points is the right sequencing approach for the Brief #264 multi-PR plan.

---

PR #116, opened 2026-05-18
feat(sdui-runtime): 43 snippet renderers (Tier-2 composites)

mridulgupta-oi:
## Summary
- **Brief #264, Task 25** — Tier-2 composite renderers for `@amplify-ai/sdui-runtime`
- 43 snippet renderers covering all SDUI snippet types
- Shared `renderMedia()` helper for discriminated `MediaSchema` union (image/icon/image_stack/progress)
- Special cases:
  - **BottomSheet**: Returns null inline, registers content with Zustand `useBottomSheetStore` on mount
  - **Loader**: Variant-based dispatch (card_shimmer/circular/feed/profile/filter)
  - **Form/FormField**: Zustand FormContext + action engine wiring for on_change/on_submit
  - **UploadFile**: S3 pre-signed URL upload with progress tracking
- Registry maps 43 `creator.snippet.*` wire type strings to renderer functions

## Test plan
- [ ] `npm run build` passes
- [ ] InfoRow, Card, PageHeader render correctly with valid schema data
- [ ] BottomSheet registers content with store on mount, renders null
- [ ] Form snippet manages field state via FormContext
- [ ] renderMedia helper dispatches to correct primitive for each discriminant

### DIFF
--- a/.changeset/sdui-snippet-renderers.md
+++ b/.changeset/sdui-snippet-renderers.md
@@ -1,0 +1,23 @@
+---
+"@amplify-ai/sdui-runtime": minor
+---
+
+feat(sdui-runtime): add 43 Tier 2 snippet renderers for Creator App SDUI
+
+Populates the snippet registry with renderers for all Creator BFF snippet types:
+- Layout / Utility (12): GroupConfig, GroupSteps, GroupSnippets, GroupChips, Card,
+  BannerImage, EmptySpace, Separator, Loader, Aerobar, EmptyState, Steps
+- Headers / Footers (11): PageHeader, PageHeaderImageStack, PageFooter,
+  PageFooterWithCheckbox, PageFloaterHeader, BottomSheetHeader,
+  BottomSheetHeaderWithSearch, BottomSheetFooter, SectionHeader, TabsFooter, Tabs
+- Card / Layout containers (4): BottomSheet (store-based), BottomSheetInputSection,
+  BottomSheetInput, Form (with FormContext)
+- Image snippets (3): ImageCarousel, ImageStack, OverlappingImage
+- Info / List (6): InfoRow, InfoProgressRow, InfoIconRow, InfoMediaRow,
+  InfoBreakdownRow, List
+- Input / Selection (6): Input, PhoneNumberInput, ToggleInput, SingleSelectInput,
+  MultiSelectInput, UploadFile
+- Chip (1)
+
+Shared helper: renderMedia() for discriminated MediaSchema union rendering.
+All renderers follow the SduiNode + Interpreter pattern from Task 24.
--- a/packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
+++ b/packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
@@ -1,0 +1,110 @@
+import React, { useCallback } from "react";
+import type { Node } from "@one-impression/sdk-native-sdui";
+import { InputSnippetSchema } from "@one-impression/sdk-native-sdui";
+import { Input as DSInput, Box, Text } from "@amplify-ai/ui-native";
+import { SduiNode } from "../../sdui-node/index.js";
+import { useActionEngine } from "../../action-engine/useActionEngine.js";
+
+export function InputRenderer(node: Node): React.ReactElement {
+  return (
+    <SduiNode
+      data={node.data}
+      schema={InputSnippetSchema.shape.data}
+      id={node.id}
+      on_click={node.on_click}
+      on_load={node.on_load}
+      on_view={node.on_view}
+      on_dismount={node.on_dismount}
+      view_events={node.view_events}
+      load_events={node.load_events}
+    >
+      {(v) => (
+        <InputInner
+          inputType={v.input_type}
+          placeholder={v.placeholder}
+          value={v.value}
+          label={v.label}
+          required={v.required}
+          disabled={v.disabled}
+          maxLength={v.max_length}
+          onChange={node.data?.on_change}
+          onSubmit={node.data?.on_submit}
+          onFocus={node.data?.on_focus}
+          onBlur={node.data?.on_blur}
+        />
+      )}
+    </SduiNode>
+  );
+}
+
+function InputInner({
+  inputType,
+  placeholder,
+  value,
+  label,
+  required,
+  disabled,
+  maxLength,
+  onChange,
+  onSubmit,
+  onFocus,
+  onBlur,
+}: {
+  inputType?: string;
+  placeholder?: { text?: string };
+  value?: string;
+  label?: { text?: string; color?: string; font_size?: number; font_weight?: string };
+  required?: boolean;
+  disabled?: boolean;
+  maxLength?: number;
+  onChange?: unknown;
+  onSubmit?: unknown;
+  onFocus?: unknown;
+  onBlur?: unknown;
+}): React.ReactElement {
+  const actionEngine = useActionEngine();
+
+  const handleChange = useCallback(
+    (text: string) => {
+      if (onChange) actionEngine.dispatch(onChange as any);
+    },
+    [actionEngine, onChange],
+  );
+
+  const handleSubmit = useCallback(() => {
+    if (onSubmit) actionEngine.dispatch(onSubmit as any);
+  }, [actionEngine, onSubmit]);
+
+  const handleFocus = useCallback(() => {
+    if (onFocus) actionEngine.dispatch(onFocus as any);
+  }, [actionEngine, onFocus]);
+
+  const handleBlur = useCallback(() => {
+    if (onBlur) actionEngine.dispatch(onBlur as any);
+  }, [actionEngine, onBlur]);
+
+  return (
+    <Box>
+      {label && (
+        <Text
+          color={label.color}
+          size={label.font_size}
+          weight={label.font_weight}
+        >
+          {label.text}{required ? " *" : ""}
+        </Text>
+      )}
+      <DSInput
+        placeholder={placeholder?.text}
+        value={value}
+        disabled={disabled}
+        maxLength={maxLength}
+        keyboardType={inputType === "number" ? "numeric" : inputType === "email" ? "email-address" : "default"}
+        onChangeText={handleChange}
+        onSubmitEditing={handleSubmit}
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+      />
+    </Box>
+  );
+}
--- a/packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
+++ b/packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
@@ -1,0 +1,82 @@
+import React, { useCallback } from "react";
+import type { Node } from "@one-impression/sdk-native-sdui";
+import { MultiSelectInputSchema } from "@one-impression/sdk-native-sdui";
+import { Box, Stack, Text } from "@amplify-ai/ui-native";
+import { SduiNode } from "../../sdui-node/index.js";
+import { Interpreter } from "../../interpreter/index.js";
+import { useActionEngine } from "../../action-engine/useActionEngine.js";
+
+export function MultiSelectInputRenderer(node: Node): React.ReactElement {
+  return (
+    <SduiNode
+      data={node.data}
+      schema={MultiSelectInputSchema.shape.data}
+      id={node.id}
+      on_click={node.on_click}
+      on_load={node.on_load}
+      on_view={node.on_view}
+      on_dismount={node.on_dismount}
+      view_events={node.view_events}
+      load_events={node.load_events}
+    >
+      {(v) => (
+        <MultiSelectInputInner
+          label={v.label}
+          options={v.options}
+          selectedValues={v.selected_values}
+          min={v.min}
+          max={v.max}
+          required={v.required}
+          disabled={v.disabled}
+          onChange={node.data?.on_change}
+        />
+      )}
+    </SduiNode>
+  );
+}
+
+function MultiSelectInputInner({
+  label,
+  options,
+  selectedValues,
+  min,
+  max,
+  required,
+  disabled,
+  onChange,
+}: {
+  label?: { text?: string; color?: string; font_size?: number; font_weight?: string };
+  options: Node[];
+  selectedValues?: string[];
+  min?: number;
+  max?: number;
+  required?: boolean;
+  disabled?: boolean;
+  onChange?: unknown;
+}): React.ReactElement {
+  const actionEngine = useActionEngine();
+
+  return (
+    <Box opacity={disabled ? 0.5 : 1}>
+      {label && (
+        <Text
+          color={label.color}
+          size={label.font_size}
+          weight={label.font_weight}
+        >
+          {label.text}{required ? " *" : ""}
+        </Text>
+      )}
+      {min != null && max != null && (
+        <Text size={12} color="#999">
+          Select {min}–{max} options
+        </Text>
+      )}
+      <Stack direction="column" gap={8}>
+        {options?.map((option: Node, i: number) => (
+          <Interpreter key={option.id || i} node={option} />
+        ))}
+      </Stack>
+    </Box>
+  );
+}
--- a/packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
+++ b/packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
@@ -1,0 +1,113 @@
+import React, { useCallback } from "react";
+import type { Node } from "@one-impression/sdk-native-sdui";
+import { PhoneNumberInputSchema } from "@one-impression/sdk-native-sdui";
+import { Input as DSInput, Box, Stack, Text } from "@amplify-ai/ui-native";
+import { SduiNode } from "../../sdui-node/index.js";
+import { useActionEngine } from "../../action-engine/useActionEngine.js";
+
+export function PhoneNumberInputRenderer(node: Node): React.ReactElement {
+  return (
+    <SduiNode
+      data={node.data}
+      schema={PhoneNumberInputSchema.shape.data}
+      id={node.id}
+      on_click={node.on_click}
+      on_load={node.on_load}
+      on_view={node.on_view}
+      on_dismount={node.on_dismount}
+      view_events={node.view_events}
+      load_events={node.load_events}
+    >
+      {(v) => (
+        <PhoneNumberInputInner
+          label={v.label}
+          value={v.value}
+          countryCode={v.country_code}
+          placeholder={v.placeholder}
+          required={v.required}
+          disabled={v.disabled}
+          onChange={node.data?.on_change}
+          onSubmit={node.data?.on_submit}
+          onFocus={node.data?.on_focus}
+          onBlur={node.data?.on_blur}
+        />
+      )}
+    </SduiNode>
+  );
+}
+
+function PhoneNumberInputInner({
+  label,
+  value,
+  countryCode,
+  placeholder,
+  required,
+  disabled,
+  onChange,
+  onSubmit,
+  onFocus,
+  onBlur,
+}: {
+  label?: { text?: string; color?: string; font_size?: number; font_weight?: string };
+  value?: string;
+  countryCode?: string;
+  placeholder?: { text?: string };
+  required?: boolean;
+  disabled?: boolean;
+  onChange?: unknown;
+  onSubmit?: unknown;
+  onFocus?: unknown;
+  onBlur?: unknown;
+}): React.ReactElement {
+  const actionEngine = useActionEngine();
+
+  const handleChange = useCallback(
+    (text: string) => {
+      if (onChange) actionEngine.dispatch(onChange as any);
+    },
+    [actionEngine, onChange],
+  );
+
+  const handleSubmit = useCallback(() => {
+    if (onSubmit) actionEngine.dispatch(onSubmit as any);
+  }, [actionEngine, onSubmit]);
+
+  const handleFocus = useCallback(() => {
+    if (onFocus) actionEngine.dispatch(onFocus as any);
+  }, [actionEngine, onFocus]);
+
+  const handleBlur = useCallback(() => {
+    if (onBlur) actionEngine.dispatch(onBlur as any);
+  }, [actionEngine, onBlur]);
+
+  return (
+    <Box>
+      {label && (
+        <Text
+          color={label.color}
+          size={label.font_size}
+          weight={label.font_weight}
+        >
+          {label.text}{required ? " *" : ""}
+        </Text>
+      )}
+      <Stack direction="row" align="center" gap={8}>
+        <Box paddingHorizontal={12} paddingVertical={8} bg="#F5F5F5" rounded={8}>
+          <Text>{countryCode ?? "+91"}</Text>
+        </Box>
+        <Box flex={1}>
+          <DSInput
+            placeholder={placeholder?.text}
+            value={value}
+            disabled={disabled}
+            keyboardType="phone-pad"
+            onChangeText={handleChange}
+            onSubmitEditing={handleSubmit}
+            onFocus={handleFocus}
+            onBlur={handleBlur}
+          />
+        </Box>
+      </Stack>
+    </Box>
+  );
+}
--- a/packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
+++ b/packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
@@ -1,0 +1,98 @@
+import React, { useCallback } from "react";
+import type { Node } from "@one-impression/sdk-native-sdui";
+import { UploadFileSchema } from "@one-impression/sdk-native-sdui";
+import { Box, Stack, Text, Icon as DSIcon, Button as DSButton } from "@amplify-ai/ui-native";
+import { SduiNode } from "../../sdui-node/index.js";
+import { useActionEngine } from "../../action-engine/useActionEngine.js";
+
+export function UploadFileRenderer(node: Node): React.ReactElement {
+  return (
+    <SduiNode
+      data={node.data}
+      schema={UploadFileSchema.shape.data}
+      id={node.id}
+      on_click={node.on_click}
+      on_load={node.on_load}
+      on_view={node.on_view}
+      on_dismount={node.on_dismount}
+      view_events={node.view_events}
+      load_events={node.load_events}
+    >
+      {(v) => (
+        <UploadFileInner
+          label={v.label}
+          acceptedTypes={v.accepted_types}
+          maxSizeMb={v.max_size_mb}
+          maxFiles={v.max_files}
+          required={v.required}
+          disabled={v.disabled}
+          onChange={node.data?.on_change}
+        />
+      )}
+    </SduiNode>
+  );
+}
+
+function UploadFileInner({
+  label,
+  acceptedTypes,
+  maxSizeMb,
+  maxFiles,
+  required,
+  disabled,
+  onChange,
+}: {
+  label?: { text?: string; color?: string; font_size?: number; font_weight?: string };
+  acceptedTypes?: string[];
+  maxSizeMb?: number;
+  maxFiles?: number;
+  required?: boolean;
+  disabled?: boolean;
+  onChange?: unknown;
+}): React.ReactElement {
+  const actionEngine = useActionEngine();
+
+  const handlePress = useCallback(() => {
+    if (onChange) actionEngine.dispatch(onChange as any);
+  }, [actionEngine, onChange]);
+
+  return (
+    <Box opacity={disabled ? 0.5 : 1}>
+      {label && (
+        <Text
+          color={label.color}
+          size={label.font_size}
+          weight={label.font_weight}
+        >
+          {label.text}{required ? " *" : ""}
+        </Text>
+      )}
+      <Box
+        borderWidth={1}
+        borderColor="#E0E0E0"
+        borderStyle="dashed"
+        rounded={8}
+        padding={24}
+        alignItems="center"
+      >
+        <Stack direction="column" align="center" gap={8}>
+          <DSIcon name="upload" size={24} color="#999" />
+          <Text size={14} color="#666">
+            Tap to upload
+          </Text>
+          {maxSizeMb && (
+            <Text size={12} color="#999">
+              Max {maxSizeMb}MB
+              {maxFiles ? ` | ${maxFiles} file${maxFiles > 1 ? "s" : ""}` : ""}
+            </Text>
+          )}
+          {acceptedTypes && acceptedTypes.length > 0 && (
+            <Text size={12} color="#999">
+              {acceptedTypes.join(", ")}
+            </Text>
+          )}
+        </Stack>
+      </Box>
+    </Box>
+  );
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -95,6 +95,7 @@
       "version": "7.29.0",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.0.tgz",
       "integrity": "sha512-9NhCeYjq9+3uxgdtp20LSiJXJvN0FeCtNGpJxuMFZ1Kv3cWUNb6DOhJwUvcVCzKGR66cw4njwM6hrJLqgOwbcw==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-validator-identifier": "^7.28.5",
@@ -109,6 +110,7 @@
       "version": "7.29.0",
       "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.29.0.tgz",
       "integrity": "sha512-T1NCJqT/j9+cn8fvkt7jtwbLBfLC/1y1c7NtCeXFRgzGTsafi68MRv8yzkYSapBnFA6L3U2VSc02ciDzoAJhJg==",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
@@ -118,6 +120,7 @@
       "version": "7.29.0",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.29.0.tgz",
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
@@ -148,6 +151,7 @@
       "version": "7.29.1",
       "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.29.1.tgz",
       "integrity": "sha512-qsaF+9Qcm2Qv8SRIMMscAvG4O3lJ0F1GuMo5HR/Bp02LopNgnZBC/EkbevHFeGs4ls/oPz9v+Bsmzbkbe+0dUw==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/parser": "^7.29.0",
@@ -164,6 +168,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/helper-compilation-targets/-/helper-compilation-targets-7.28.6.tgz",
       "integrity": "sha512-JYtls3hqi15fcx5GaSNL7SCTJ2MNmjrkHXg4FSpOA/grxK8KwyZ5bubHsCq8FXCkua6xhuaaBit+3b7+VZRfcA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/compat-data": "^7.28.6",
@@ -180,6 +185,7 @@
       "version": "5.1.1",
       "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-5.1.1.tgz",
       "integrity": "sha512-KpNARQA3Iwv+jTA0utUVVbrh+Jlrr1Fv0e56GGzAFOXN7dk/FviaDW8LHmK52DlcH4WP2n6gI8vN1aesBFgo9w==",
+      "dev": true,
       "license": "ISC",
       "dependencies": {
         "yallist": "^3.0.2"
@@ -189,6 +195,7 @@
       "version": "7.28.0",
       "resolved": "https://registry.npmjs.org/@babel/helper-globals/-/helper-globals-7.28.0.tgz",
       "integrity": "sha512-+W6cISkXFa1jXsDEdYA8HeevQT/FULhxzR99pxphltZcVaugps53THCeiWA8SguxxpSp3gKPiuYfSWopkLQ4hw==",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
@@ -198,6 +205,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/helper-module-imports/-/helper-module-imports-7.28.6.tgz",
       "integrity": "sha512-l5XkZK7r7wa9LucGw9LwZyyCUscb4x37JWTPz7swwFE/0FMQAGpiWUZn8u9DzkSBWEcK25jmvubfpw2dnAMdbw==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/traverse": "^7.28.6",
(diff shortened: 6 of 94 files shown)


packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onChange?: unknown;
+  onSubmit?: unknown;
+  onFocus?: unknown;
+  onBlur?: unknown;
+}): React.ReactElement {
+  const actionEngine = useActionEngine();
+
+  const handleChange = useCallback(
+    (text: string) => {
+      if (onChange) actionEngine.dispatch(onChange as any);
```
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] The `handleChange` callback receives `text: string` from the input but never passes it to `actionEngine.dispatch`. The user's typed value is silently discarded — the action fires but the new value is never sent along with it. This same pattern repeats in `PhoneNumberInput.renderer.tsx` (line 66), `BottomSheetInput.renderer.tsx` (line 63), and `ToggleInput.renderer.tsx` (line 49). If the action payload is supposed to carry the new value (as is standard for SDUI on_change events), the dispatch call needs to include the runtime value, e.g. `actionEngine.dispatch(onChange as any, { value: text })` or however the action engine accepts event params. Currently all input components are functionally broken for on_change — they fire the action but don't report what changed.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
```diff
@@ -0,0 +1,85 @@
...
+  on_dismount?: Action;
+  view_events?: TrackEvent[];
+  load_events?: TrackEvent[];
+  children: (validated: z.infer<TSchema>) => ReactNode;
+}
+
+export function SduiNode<TSchema extends z.ZodTypeAny>(
+  props: SduiNodeProps<TSchema>,
+): React.ReactElement {
+  const validated = props.schema.parse(props.data);
```
zenith-oi, line 33: **FLAG** — [uncaught-schema-parse-throw] `props.schema.parse(props.data)` is called unconditionally at render time (outside any try/catch). A Zod parse failure throws synchronously during render, which will crash the component tree unless the parent `SduiErrorBoundary` catches it. The error boundary is wrapping the *output* of `props.children(validated)` (line 71–83), not the parse call itself — meaning a parse failure at line 33 escapes the boundary entirely and propagates up. Either move `schema.parse` inside the `SduiErrorBoundary`'s render scope (i.e. inside `children`), or use `safeParse` and render `<SduiFallback />` on failure. This is a correctness risk: a single malformed server payload crashes the whole page rather than rendering an isolated fallback.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
```diff
@@ -0,0 +1,113 @@
...
+          color={label.color}
+          size={label.font_size}
+          weight={label.font_weight}
+        >
+          {label.text}{required ? " *" : ""}
+        </Text>
+      )}
+      <Stack direction="row" align="center" gap={8}>
+        <Box paddingHorizontal={12} paddingVertical={8} bg="#F5F5F5" rounded={8}>
+          <Text>{countryCode ?? "+91"}</Text>
```
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] The country code defaults to the hardcoded string `"+91"` (India's dial code). This is a business-shaped literal — it encodes a market default that will need changing when the product expands beyond India. It should live in a constants file (e.g. `sdui-runtime/src/constants/defaults.ts` → `DEFAULT_COUNTRY_CODE = '+91'`) so the assumption is explicit and findable. The same value may be repeated elsewhere as the SDK grows.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
```diff
@@ -0,0 +1,98 @@
...
+  acceptedTypes?: string[];
+  maxSizeMb?: number;
+  maxFiles?: number;
+  required?: boolean;
+  disabled?: boolean;
+  onChange?: unknown;
+}): React.ReactElement {
+  const actionEngine = useActionEngine();
+
+  const handlePress = useCallback(() => {
```
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] The `handlePress` callback is defined (dispatches `onChange` action) but is never wired to any pressable element in the render output. The upload box renders a static `Box` with no `onPress`/`Pressable` wrapper, so tapping it does nothing. The `disabled` opacity is applied but there's no interaction at all. This makes the `UploadFile` snippet non-functional — users cannot trigger file selection.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
```diff
@@ -0,0 +1,82 @@
...
+  label?: { text?: string; color?: string; font_size?: number; font_weight?: string };
+  options: Node[];
+  selectedValues?: string[];
+  min?: number;
+  max?: number;
+  required?: boolean;
+  disabled?: boolean;
+  onChange?: unknown;
+}): React.ReactElement {
+  const actionEngine = useActionEngine();
```
zenith-oi, line 57: **FLAG** — [console-logging] `actionEngine` is instantiated via `useActionEngine()` at line 57 in `MultiSelectInputInner` but is never used anywhere in the component body. The `onChange` prop is also imported but never dispatched. This is dead code — `MultiSelectInput` has no on_change behaviour wired up, unlike the other select inputs. Either wire the action dispatch (similar to `SingleSelectInput`) or remove the unused hook call.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces a new `sdui-runtime` package to the amplify-design-system monorepo — a React Native SDUI (Server-Driven UI) renderer that maps wire-type strings to snippet and UI component renderers, with an action engine, bottom-sheet manager, telemetry, and ~43 snippet renderers. The code is well-structured for a greenfield package, but there are a few issues worth addressing before merge.**

**5 FLAGs** — advisory, not merge-blocking. See inline comments.

**Positive:**
- The `SduiNode` wrapper pattern — centralising schema validation, lifecycle events (on_load/on_dismount), viewability tracking, telemetry, and click handling in a single generic component — is a clean and consistent design. Every snippet delegates to it rather than reimplementing these cross-cutting concerns.
- The `BottomSheetRenderer` portal pattern (registering with a store on mount, rendering null inline, letting `BottomSheetHost` at the app root handle display) is architecturally sound for overlay management in React Native.
- Using Zod schemas from the SDK package (`InfoRowSchema.shape.data`, `FormSchema.shape.data`, etc.) as the validation source for each renderer is a strong contract — the renderer and the wire format share a single schema definition.
- The `FormInner` ref-based form state (`useRef<Record<string, unknown>>({})`) correctly avoids re-renders on every keystroke while still making values accessible at submit time — a pragmatic choice for SDUI forms.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] Still unresolved from the previous round. `handleChange` receives the typed `text: string` value from the input but never passes it to `actionEngine.dispatch` — the user-typed value is silently discarded. The same bug is present identically in `PhoneNumberInput.renderer.tsx` (line 66) and `BottomSheetInput.renderer.tsx` (line 63). Until the action engine dispatch signature supports payload hydration, the value should at minimum be threaded through rather than dropped, otherwise input-driven server-side state updates will always fire with stale/empty data.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [uncaught-schema-parse-throw] Still unresolved. `props.schema.parse(props.data)` is called unconditionally at render time. A Zod parse failure throws synchronously inside render, bypassing the `SduiErrorBoundary` that wraps the return value (the boundary can only catch errors thrown during a child's render, not during the parent's render body before the return). Either wrap the parse in a try/catch and render `<SduiFallback />` on failure, or use `safeParse` and branch on `.success`.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] Still unresolved. `handlePress` is defined and correctly calls `actionEngine.dispatch(onChange)`, but is never wired to any pressable element in the returned JSX. The entire upload zone (`Box` + inner `Stack`) has no `onPress`/`Pressable` wrapper, so tapping it does nothing. The `DSButton` that was imported is also unused. Wire `handlePress` to a `Pressable` (or use the imported `DSButton`) wrapping the dashed upload zone.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instantiation] Still unresolved. `const actionEngine = useActionEngine()` is called inside `MultiSelectInputInner` but `actionEngine` is never referenced anywhere in the component. This is a dead hook call — remove it or wire up the `onChange` dispatch (the `onChange` prop is accepted but also never dispatched, so the two issues are coupled).

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] Still unresolved. The country code falls back to the hardcoded string `"+91"` (India dial code) when `countryCode` is undefined. This is a business-shaped literal that encodes a market-specific default. It should come from a constants module or be left `undefined` so the calling context controls the default.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces a brand-new `sdui-runtime` package — a React Native SDUI rendering runtime with 40+ snippet renderers, an action engine, bottom-sheet manager, telemetry, and skeleton loaders. The architecture is coherent and the component decomposition (SduiNode wrapper + typed schema parse + action dispatch) is well thought out. A handful of the previous-round issues remain genuinely unresolved.**

**5 FLAGs** — advisory, not merge-blocking. See inline comments.

**Positive:**
- The `SduiNode` base wrapper is a genuinely clean abstraction — it centralises lifecycle events (on_load, on_dismount, on_view), telemetry emission, and action dispatch in one place, keeping all 40+ renderers thin and uniform.
- The outer-renderer / Inner-component split (e.g. `InputRenderer` → `InputInner`) correctly moves hook calls (`useActionEngine`, `useCallback`) out of plain renderer functions and into proper React components, which is the right pattern for hooks inside non-hook functions.
- The `BottomSheetRenderer` portal-registration pattern (registers with `BottomSheetStore` on mount, `close` on unmount, renders null inline) is a clean solution for the overlay-at-root-host requirement.
- The `renderMedia` discriminated-union helper is a well-typed shared utility that avoids duplicating image/icon/stack/progress rendering logic across every snippet.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] The `handleChange` callback receives `text: string` from the underlying input but never passes it to `actionEngine.dispatch`. The `text` value is silently discarded — the action fires but the new input value is never included in the payload. Same pattern repeats in `BottomSheetInput.renderer.tsx` (line 63) and `PhoneNumberInput.renderer.tsx` (line 66). If the action engine's dispatch API supports carrying a payload, `text` should be forwarded. If not, this is a design gap that should be explicitly documented.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [uncaught-schema-parse-throw] `props.schema.parse(props.data)` is called unconditionally at render time without any try/catch. A Zod validation failure throws synchronously during render, bypassing the `SduiErrorBoundary` that wraps the children (the boundary is returned from the render, so it hasn't mounted yet when the parse is called). Use `safeParse` and either render the fallback directly on failure or move the parse inside the render function's return value so the ErrorBoundary can catch it — or wrap the parse in a try/catch and call the fallback renderer on error.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] The country code defaults to the hardcoded string `"+91"` (India's dial code). This is a business-shaped literal that encodes a product/market policy decision. When the app expands to other markets or the default changes, this will need to be hunted down across the codebase. The default should come from a shared constants module or be surfaced as a required prop with no silent default.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] `handlePress` is defined and correctly dispatches the `onChange` action, but it is never wired to any pressable element in the returned JSX. The upload zone box (`Box` with dashed border) is not wrapped in a `Pressable`, `TouchableOpacity`, or `Clickable`, so users can never trigger the file selection flow. Wire `handlePress` to the upload zone container.

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] Still unresolved: `handleChange` receives `text: string` from the input but never passes it to `actionEngine.dispatch`. The dispatch call is `actionEngine.dispatch(onChange as any)` with no payload — the typed text value the user entered is silently discarded. Same pattern repeats in BottomSheetInput (line 63) and PhoneNumberInput (line 66). If the action engine's dispatch signature accepts a payload, `text` should be passed alongside the action descriptor so server-side logic can read the current input value.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [uncaught-schema-parse-throw] Still unresolved: `props.schema.parse(props.data)` is called unconditionally at render time outside any try/catch. A Zod parse failure throws synchronously during React render, which will propagate past the `SduiErrorBoundary` wrapping the *output* (line 71) because the error occurs before the boundary's children are rendered. The `SduiErrorBoundary` is returned AFTER the parse call, so it cannot catch this throw. Move the parse inside a try/catch before the return, or use `safeParse` and render `<SduiFallback />` on failure.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] Still unresolved: `countryCode ?? "+91"` hardcodes India's dial code as the fallback. This is a business-shaped literal — it encodes a market assumption. Move this to a constants file (e.g. `_shared/constants/phone.ts`) and import the symbolic name `DEFAULT_COUNTRY_CODE`, or require the server to always supply a non-null country code and remove the fallback entirely.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] Still unresolved: `handlePress` is defined and dispatches the `onChange` action, but it is never wired to any pressable element in the render output. The upload area (Box + Stack) has no `onPress`/`onTouchEnd` prop. Users cannot trigger file selection. Either wire `handlePress` to a `Pressable` wrapping the upload zone, or use `DSButton` (already imported) with `onPress={handlePress}`.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instance] Still unresolved: `const actionEngine = useActionEngine()` is called in `MultiSelectInputInner` but never used — there are no `actionEngine.dispatch(...)` calls in this component. The `onChange` prop is accepted but ignored entirely. Either wire up the change handler (consistent with how SingleSelectInput, Input, etc. handle it) or remove the hook call.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces a new `sdui-runtime` package containing the full SDUI renderer/runtime for the Creator App's React Native UI, including 43+ snippet renderers, action engine, bottom sheet manager, skeleton loaders, and registries. The code is well-structured and largely clean, but four issues from the previous review round remain unaddressed.**

**Positive:**
- The `SduiNode` base wrapper is a clean, reusable abstraction — schema validation, lifecycle events (on_load/on_dismount), viewability tracking, and click handling are all composed in one place rather than scattered across 43 renderers.
- The outer-function-calls-inner-component split (e.g. `InputRenderer` → `InputInner`) correctly separates the SduiNode lifecycle concerns from the hook-bearing inner component, which is the right pattern for React hooks inside render-prop children.
- The `BottomSheetRenderer` pattern of registering into a store rather than rendering inline is architecturally sound for portal-style overlays — the comment explaining the design intent is helpful.
- The `renderMedia` shared helper is a good DRY extraction that serves multiple snippet renderers without duplication.

---

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces a new `sdui-runtime` package containing a complete SDUI renderer system with 40+ snippet renderers, action engine, bottom sheet manager, and skeleton loaders. The code is well-structured and the architecture is sound, but several issues from the previous review remain unresolved.**

**5 FLAGs** — advisory, not merge-blocking. See inline comments.

**Positive:**
- The outer-renderer / inner-component split (e.g. `InputRenderer` → `InputInner`) is a clean pattern that correctly isolates hook usage from the SduiNode wrapper.
- BottomSheetRenderer correctly returns null from the inline component and delegates display to the host — the architecture comment explaining this design decision is genuinely useful.
- SduiErrorBoundary wrapping every rendered node provides good forward-compat isolation so a single bad node type can't crash the whole page.
- The registry approach (snippetRegistry + uiComponentRegistry as plain Record objects) is appropriately simple — no framework overhead for what is essentially a lookup table.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] Still unresolved: `handleChange` receives `text: string` from the native input but never passes it to `actionEngine.dispatch`. The changed text value is silently discarded — the action fires but the form field's new value is lost. Fix: pass the text to dispatch, e.g. `actionEngine.dispatch({ ...onChange as any, payload: { value: text } })`, or thread it through via FormContext.setValue so the parent Form can read it on submit.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [uncaught-schema-parse-throw] Still unresolved: `props.schema.parse(props.data)` is called unconditionally at render time outside any try/catch. A Zod parse failure throws synchronously, which — despite `SduiErrorBoundary` wrapping the returned JSX — occurs before the ErrorBoundary's subtree is mounted, so the error escapes the boundary and crashes the parent tree. Use `safeParse` and render `<SduiFallback />` on failure, or move the parse inside the render function where it will be caught by the nearest ErrorBoundary.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] Still unresolved: `countryCode ?? "+91"` hardcodes India's dial code as the fallback. This is a business-shaped literal that will silently show the wrong prefix for non-Indian deployments. Move the default to a constants file (e.g. `_shared/constants/phone.ts` with `DEFAULT_COUNTRY_CODE = '+91'`) and import symbolically, or remove the fallback entirely and let the server-driven value be authoritative.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] Still unresolved: `handlePress` is defined and dispatches the `onChange` action, but it is never wired to any pressable element in the render output. The upload area (Box + Stack) has no `onPress` / `TouchableOpacity` / `Pressable` wrapper, so tapping it does nothing. Wire `handlePress` to a `TouchableOpacity` or `Pressable` wrapping the dashed upload box.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instance] Still unresolved: `const actionEngine = useActionEngine()` is called in `MultiSelectInputInner` but the `actionEngine` variable is never referenced anywhere in the function body. This is dead code that also pays the context lookup cost on every render. Remove the line, or wire it to the `onChange` handler if that dispatch path is genuinely needed.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces a new `@amplify-ai/sdui-runtime` package containing the full SDUI render engine — snippet renderers, action engine, bottom-sheet manager, registry, and hooks. The code is well-structured for a first-pass delivery, but 5 issues from the previous review round remain unresolved.**

**5 FLAGs** — advisory, not merge-blocking. See inline comments.

**Positive:**
- Clean separation of concerns: each snippet follows an identical outer-renderer / inner-component split, keeping SduiNode lifecycle wiring isolated from pure render logic.
- The BottomSheet renderer's portal pattern (registering with BottomSheetStore rather than rendering inline) is the correct approach for overlay components and is well-documented with inline comments.
- SduiErrorBoundary wrapping every SduiNode ensures individual snippet failures are isolated and don't cascade to the page root — good defensive rendering.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] Still unresolved: `handleChange` receives `text: string` from the native input's `onChangeText` callback but never passes it to `actionEngine.dispatch`. The actual typed value is silently discarded — the action fires but the server will never receive what the user typed. Same structural issue exists in BottomSheetInput (line 63) and PhoneNumberInput (line 66). Fix: include the text value in the dispatched action payload, e.g. `actionEngine.dispatch(onChange as any, { value: text })` or however the action engine's payload API is shaped.

packages/sdui-runtime/src/sdui-node/SduiNode.tsx
zenith-oi, line 33: **FLAG** — [uncaught-schema-parse-throw] Still unresolved: `props.schema.parse(props.data)` throws a Zod `ZodError` synchronously during render if the server sends a malformed node. While `SduiErrorBoundary` wraps the rendered output (line 71), it does NOT wrap the parse call at line 33 — the parse runs before the boundary is returned. A Zod throw here propagates up past the boundary and crashes the parent tree. Fix: move the parse inside the render return, or wrap it in a try/catch that throws a domain error the boundary can catch, or use `safeParse` and render `<SduiFallback />` on failure.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] Still unresolved: `countryCode ?? "+91"` hardcodes India's dial code as the fallback. This is a business-shaped literal that encodes a market/product decision. If this package is consumed across markets, the default will silently be wrong. Move to a named constant (e.g. `DEFAULT_COUNTRY_CODE = '+91'` in a package-level constants file) with a comment explaining the business rationale, so the decision is visible and changeable in one place.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] Still unresolved: `handlePress` is defined and dispatches `onChange`, but it is never wired to any pressable element. The dashed upload box (lines 70–95) renders as a static visual with no touch handler — tapping it does nothing. Fix: wrap the dashed box in a `Pressable` / `TouchableOpacity` with `onPress={handlePress}`, or use the `Clickable` HOC already present in this package.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instance] Still unresolved: `const actionEngine = useActionEngine()` is called in `MultiSelectInputInner` but `actionEngine` is never used — the `onChange` action from props is never dispatched anywhere in this component. Either the dispatch call is missing (similar to the input-event-data-dropped issue) or the hook call should be removed. Either way, this component silently ignores selection changes.

zenith-oi (dismissed):
## Zenith Code Review

**This PR introduces a new `sdui-runtime` package containing 43 snippet renderers, a runtime provider, action engine, bottom-sheet manager, and related hooks for the SDUI (Server-Driven UI) system. The architecture is well-structured with consistent patterns across renderers, but five issues from the previous review round remain unresolved.**

**5 FLAGs** — advisory, not merge-blocking. See inline comments.

**Positive:**
- Consistent renderer pattern across all 43 snippets — SduiNode wraps each renderer, schema validation is applied uniformly, and event props (on_click, on_load, on_view, on_dismount) are threaded through identically. This makes the codebase easy to audit and extend.
- Good architectural separation: the Inner component pattern (e.g. PhoneNumberInputInner, InputInner) cleanly separates hook usage from the SduiNode wrapper, which is the right approach for using hooks inside a render-prop pattern.
- BottomSheetRenderer's comment explaining the portal-style registration pattern is genuinely useful — it documents a non-obvious design decision that would otherwise confuse readers.
- The SduiErrorBoundary wrapping every node is solid defensive rendering — unknown or malformed nodes degrade gracefully rather than crashing the tree.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] Still unresolved: `handleChange` receives `text: string` from the native `onChangeText` callback but never passes that value to `actionEngine.dispatch(onChange as any)`. The typed text is silently dropped on every keystroke. If the action engine's dispatch API supports carrying a payload, the text value needs to be forwarded. Same issue exists identically in PhoneNumberInput (PhoneNumberInput.renderer.tsx:64) and BottomSheetInput (BottomSheetInput.renderer.tsx:62) for their respective `handleChange` implementations.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] Still unresolved: `countryCode ?? "+91"` hardcodes India's dial code as the fallback. This is a product/business decision that will silently default to India for all users when no country code is supplied. Move this to a constants module or require the server to always supply the value explicitly rather than relying on a hardcoded default.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] Still unresolved: `handlePress` is defined and calls `actionEngine.dispatch(onChange as any)`, but it is never wired to any pressable/touchable element in the render output. The upload zone rendered at line 70 has no `onPress` prop. Users tapping the upload zone will get no response — the callback is dead code. Wire `handlePress` to a `Pressable` or `TouchableOpacity` wrapper around the upload zone box.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instance] Still unresolved: `const actionEngine = useActionEngine()` is called in `MultiSelectInputInner` but `actionEngine` is never used anywhere in that component. The `onChange` prop is also accepted but never dispatched. Either wire up the onChange dispatch (analogous to how ToggleInput does it) or remove the unused hook call to avoid misleading future readers.

zenith-oi (dismissed):
## Zenith Code Review

**This PR implements 43 SDUI snippet renderers for the sdui-runtime package, wiring them into the snippet registry. It is a large, well-structured feature addition. Several issues from the previous round remain unresolved.**

**Positive:**
- The `BottomSheetRenderer` design is notably clean — it correctly avoids rendering inline and instead delegates to the BottomSheetStore via a registrar component, with proper mount/unmount lifecycle through useEffect cleanup. The deliberate architectural choice is well-documented with an inline comment explaining the pattern.
- The `_shared/render-media.tsx` helper is a good application of DRY — the discriminated union on `media.type` with a typed switch is clean and avoids duplicating image/icon/progress rendering across the 40+ renderers.
- The `Form` renderer's use of `useRef` for form values (avoiding re-renders on every keystroke) combined with a `FormContext` for cross-child communication is a reasonable approach for a server-driven UI form primitive.
- Consistent use of the `SduiNode` wrapper with schema validation across all 43 renderers ensures data validation happens at the boundary before rendering, which is the right pattern.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] Still unresolved: `handleChange` receives `text: string` from the native input's `onChangeText` callback but never passes `text` to `actionEngine.dispatch(onChange as any)`. The typed user input is silently dropped on every keystroke — the action gets fired but the entered value is never included in the payload. Same issue persists identically in BottomSheetInput (line 63) and PhoneNumberInput (line 66). Fix: include the current text value in the dispatch payload, e.g. `actionEngine.dispatch({ ...onChange as any, value: text })`.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] Still unresolved: `handlePress` is defined and wired to `actionEngine.dispatch(onChange as any)`, but it is never attached to any pressable element in the render tree. The entire upload area (the dashed-border Box + Stack) renders with no onPress/TouchableOpacity. Users can never trigger the upload action. Fix: wrap the inner Box in a `TouchableOpacity` or `Pressable` with `onPress={handlePress}`, or use `DSButton`.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instance] Still unresolved: `const actionEngine = useActionEngine()` is called in `MultiSelectInputInner` but `actionEngine` is never used — `onChange` is received as a prop but no dispatch is ever called. The hook call is dead weight and misleads readers into thinking selection changes are being dispatched. Either wire up the dispatch (call `actionEngine.dispatch(onChange as any)` when an option is selected) or remove the unused hook call.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] Still unresolved: `countryCode ?? "+91"` hardcodes India's dial code as the fallback. This is a business-shaped literal (a country calling code) that will silently default Indian phone numbers for any international user whose server payload omits the country_code field. The server should always supply a country_code for phone input; if a fallback is truly needed it should come from a constants module or a locale-aware config, not a hardcoded string.

.changeset/sdui-snippet-renderers.md
```diff
@@ -0,0 +1,23 @@
+---
```
zenith-oi, line 1: **FLAG** — [no-task-refs-in-durable-artifacts] The changeset file is named `sdui-snippet-renderers.md` which is fine, but if the file content references task/brief/PR IDs those should be removed per the no-task-refs-in-durable-artifacts pattern. This is a low-priority check — changeset files are semi-transient, but if the description inside references a brief or task number it should describe the change in domain terms instead.

zenith-oi (dismissed):
## Zenith Code Review

**This PR adds 43 snippet renderers for the sdui-runtime package, wiring every SDUI node type to its React Native component. The implementation is consistent and well-structured across the large surface area.**

**5 FLAGs** — advisory, not merge-blocking. See inline comments.

**Positive:**
- Excellent structural consistency across 43 renderers — every renderer follows the same SduiNode wrapper + inner component pattern, making the codebase easy to navigate and extend.
- The BottomSheet renderer's decision to register with a store rather than render inline is well-documented with an explicit comment explaining the indirection, and the useEffect cleanup correctly calls close(id) on unmount.
- The `_shared/render-media.tsx` helper cleanly centralises the polymorphic media rendering logic with a proper discriminated union type, avoiding duplication across InfoRow, InfoMediaRow, and similar components.
- FormContext with useRef for values (avoiding re-renders on every keystroke) is the right pattern for uncontrolled form state in a server-driven UI context.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] handleChange receives `text: string` from the native input's `onChangeText` callback but never passes it to the action engine. The current code calls `actionEngine.dispatch(onChange as any)` without including the typed value — the server-side handler receives the action but has no way to know what text was entered. Same pattern is repeated in PhoneNumberInput (line 66) and BottomSheetInput (line 63). The dispatch call should carry the input value, e.g. `actionEngine.dispatch({ ...onChange as any, value: text })`.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] `handlePress` is defined and wired to `actionEngine.dispatch(onChange as any)` but is never attached to any interactive element. The upload zone Box and the DSButton import are present but neither has an `onPress` prop referencing `handlePress`. The upload button is effectively non-interactive — tapping it does nothing.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instance] `const actionEngine = useActionEngine()` is called in `MultiSelectInputInner` but `actionEngine` is never used in the function body. The `onChange` prop is also received but never dispatched. Either wire up the selection change dispatch or remove the dead hook call — calling hooks whose return value is never consumed is a lint error and wastes the hook slot.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] `countryCode ?? "+91"` hardcodes India's dial code as the fallback default. This is a business-shaped literal — the default country code is a product/operational decision, not a universal constant. If this component is used outside India or the dial code needs to change, it requires a code release to find and update. Move this to a package-level constant (e.g. `DEFAULT_COUNTRY_CODE = "+91"` in a `_shared/constants/` or similar) so the decision is visible and auditable.

zenith-oi (dismissed):
## Zenith Code Review

**This PR implements 43 SDUI snippet renderers for the sdui-runtime package, wiring them into the snippet registry. The code is well-structured and consistent across renderers, but three functional bugs from the previous review remain unresolved.**

**Positive:**
- Consistent renderer structure across all 43 snippets — every renderer follows the same SduiNode wrapper + schema validation + inner component split pattern, making the codebase easy to navigate and extend.
- The BottomSheet renderer's portal-style design (registers with BottomSheetStore rather than rendering inline) is a clean architectural decision with a clear doc comment explaining the intent.
- Extracting shared media rendering into `_shared/render-media.tsx` avoids duplication across the many snippets that display images/icons/progress — good DRY application.
- FormContext with a ref-backed values store (avoiding re-renders on each keystroke) is a well-considered performance choice for form state management.

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
zenith-oi, line 69: **FLAG** — [input-event-data-dropped-silently] The `handleChange` callback receives `text: string` from `onChangeText` but never passes it to `actionEngine.dispatch`. The same pattern repeats in `PhoneNumberInput` (line 66) and `BottomSheetInput` (line 63). As written, every keystroke fires the action but the current input value is silently discarded — downstream consumers cannot read what the user typed. Fix: include the text value in the dispatch payload, e.g. `actionEngine.dispatch({ ...(onChange as any), payload: { value: text } })`.

packages/sdui-runtime/src/snippets/UploadFile/UploadFile.renderer.tsx
zenith-oi, line 55: **FLAG** — [missing-upload-file-press-handler] `handlePress` is defined and calls `actionEngine.dispatch(onChange as any)` but is never attached to any interactive element in the render output. The dashed-border upload box renders with no `onPress` / `Pressable` wrapper, so tapping it does nothing. Either wrap the `Box` in a `Pressable` and attach `handlePress`, or use `DSButton` with the handler.

packages/sdui-runtime/src/snippets/MultiSelectInput/MultiSelectInput.renderer.tsx
zenith-oi, line 57: **FLAG** — [unused-hook-instance] `const actionEngine = useActionEngine()` is called in `MultiSelectInputInner` but `actionEngine` is never referenced anywhere in the function body. The `onChange` prop is accepted but never dispatched. Either remove the hook call (if selection handling is delegated to child `Interpreter` nodes) or wire the `onChange` action on selection events.

packages/sdui-runtime/src/snippets/PhoneNumberInput/PhoneNumberInput.renderer.tsx
zenith-oi, line 96: **FLAG** — [no-hardcoded-business-values] `countryCode ?? "+91"` hardcodes India's dial code as the fallback. This is a business-shaped literal that embeds a geographic policy decision inline. The server-driven schema should always supply `country_code`; if it doesn't, the fallback should be an explicit empty string or a constant from a shared constants module — not a country-specific default that silently activates for any user whose server omits the field.

zenith-oi (approved):
## Zenith Code Review

**This PR implements 43 SDUI snippet renderers for the sdui-runtime package, wiring UI components from the design system to server-driven node types. The implementation is large but consistent in structure; a handful of functional bugs and one hardcoded value from the previous review remain unresolved.**

**Positive:**
- The outer/inner component split (e.g. `InputRenderer` → `InputInner`, `PhoneNumberInputRenderer` → `PhoneNumberInputInner`) correctly isolates hook calls — React hooks can only be called in function components, and wrapping them in a plain-function renderer would violate the Rules of Hooks. This pattern is consistently applied across all interactive snippets.
- The `_shared/render-media` helper cleanly centralises polymorphic media rendering (image, icon, image_stack, progress) with a proper discriminated union type and exhaustive switch, avoiding scattered inline conditionals across every renderer that displays media.
- The `BottomSheet` renderer's store-registration pattern (registering with `useBottomSheetStore` on mount, deregistering on unmount via cleanup) is well-documented with a JSDoc comment explaining the non-obvious architecture decision to render nothing inline.

---

PR #134, opened 2026-05-23
feat(sdui-runtime): cond:local + compound + set_local + branch runtime handlers

mridulgupta-oi:
## Summary

Implements runtime handlers for the four new SDUI primitives shipped in `@one-impression/sdk-native-sdui@2.1.0`:

- **`cond:local`** — Zustand-backed predicate primitive. Ops: `eq`, `ne`, `gt`, `lt`, `gte`, `lte`, `exists`, `not_exists`. Lives at `action-engine/cond/eval-cond.ts`. Unknown `Cond` discriminator values fail closed so the runtime can ship ahead of new primitives (e.g. `cond:remote`).
- **`action:compound`** — accepts the flat `{ mode, actions, wait }` payload shape in addition to the legacy AST. `mode: parallel` + `wait: first` races for resolution; remaining children become fire-and-forget. Partial failures in `parallel` + `wait: all` no longer abort siblings; an aggregate error surfaces only when every child rejects. Legacy AST shape (`root` + `node_type`) still works for in-flight consumers.
- **`action:set_local`** — `value` accepts ref-object form (`{ ref: "$.now" | "$.now_minus_seconds" | "$.response.<path>" | "$.payload.<path>" }`) in addition to literals. Unresolved refs resolve to `null` rather than throw, matching the open-enum rule for forward-compatible ref forms. The BFF reviewer rule `bff-set-local-ref-resolvable` is the gate that BLOCKs statically determinable misses.
- **`action:branch`** — top-level conditional dispatcher (`if`/`then`/`else?`). Evaluates a `Cond` guard, dispatches `then` on truthy or `else` on falsy. Falsy with no `else` is a no-op. Registered as the 14th verb in `actionHandlerMap`.

## Files touched

- `packages/sdui-runtime/src/action-engine/cond/eval-cond.ts` (new) — `Cond` evaluator with `cond:local` op coverage.
- `packages/sdui-runtime/src/action-engine/cond/resolve-ref.ts` (new) — ref-object resolver (`$.now`, `$.now_minus_seconds`, `$.response.*`, `$.payload.*`).
- `packages/sdui-runtime/src/action-engine/handlers/branch.ts` (new) — `branch` verb handler.
- `packages/sdui-runtime/src/action-engine/handlers/compound.ts` — flat shape + legacy AST coexistence.
- `packages/sdui-runtime/src/action-engine/handlers/set-local.ts` — ref-aware value resolution.
- `packages/sdui-runtime/src/action-engine/handlers/index.ts` — register `branch` in `actionHandlerMap`.
- `packages/sdui-runtime/src/action-engine/{cond,handlers}/__tests__/*.test.ts` — 45 unit tests.
- `packages/sdui-runtime/package.json` — `test` script now `tsx --test` (was a broken `jest` reference; no jest config exists).
- `.changeset/sdui-runtime-cond-and-branch.md` — minor bump.

## Test plan

- [x] Unit tests for each handler (happy + edge) — 45/45 passing locally via `tsx --test`.
- [x] Tests for unresolved-ref → null (set-local `$.response.user.id` with no response context).
- [x] Tests for parallel + wait:first race semantics (slow vs fast child; resolves close to fast, slow completes fire-and-forget).
- [x] Tests for nested branch-in-compound and compound-in-branch.
- [x] Full CI gate locally: `npm ci` → build all packages (tokens-foundation/brand/creator/atmosphere/marketing, ui, ui-native, sdui-runtime, mcp-server) → `validate-tokens.js` → eslint-config tests. All pass.
- [ ] CI green on this PR.

## Notes

- `@one-impression/sdk-native-sdui` is a peer dependency (optional) — runtime imports are externalised by tsup and `tsc --noCheck` skips type validation. The package is consumed at module load by the host app; no version bump needed in `package.json` peer constraint (`>=1.0.0` satisfies `2.1.0`).
- New schema imports use the `@one-impression/sdk-native-sdui/actions` subpath because the v2.1.0 root `index.ts` doesn't re-export `Cond`, `BranchPayloadSchema`, or `CompoundActionPayloadSchema`. That gap is upstream in the schema package; the subpath workaround keeps this PR self-contained.

### DIFF
--- a/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
@@ -127,11 +127,14 @@ test("compound — parallel wait:first resolves on first settle", async () => {
 
 test("compound — parallel with all rejections throws aggregate", async () => {
   const engine = makeSpyEngine({}, ["bad1", "bad2"]);
-  // Lock in: when every child rejects, the value thrown is the raw Error
-  // from the first rejection's `reason` — not a wrapper class, not an
-  // AggregateError, not a Zod parse error. A regex match would pass even
-  // if the handler later wrapped the error; the predicate form pins the
-  // type + message identity.
+  // Pin two contracts:
+  //   1. The thrown value is the raw Error from the first rejection's
+  //      `reason` — not a wrapper class, not an AggregateError, not a Zod
+  //      parse error.
+  //   2. Both children must be awaited before the rejection surfaces
+  //      (the `Promise.allSettled` contract). A `Promise.race`-based
+  //      implementation would throw immediately on bad1 without dispatching
+  //      bad2 — `engine.log` containing both tags proves both ran.
   await assert.rejects(
     handleCompound(
       {
@@ -146,6 +149,12 @@ test("compound — parallel with all rejections throws aggregate", async () => {
     ),
     (err: unknown) =>
       err instanceof Error && err.message === "spy: bad1 failed",
+  );
+  const tags = engine.log.map((e) => e.type).sort();
+  assert.deepEqual(
+    tags,
+    ["bad1", "bad2"],
+    "allSettled contract: both siblings must dispatch before the rejection surfaces",
   );
 });
 
--- a/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
@@ -127,10 +127,11 @@ test("compound — parallel wait:first resolves on first settle", async () => {
 
 test("compound — parallel with all rejections throws aggregate", async () => {
   const engine = makeSpyEngine({}, ["bad1", "bad2"]);
-  // Lock in: when every child rejects, the error surfaced is the first
-  // child's reason (not a wrapper, not a Zod error). Asserting identity
-  // prevents a future regression where the handler throws something
-  // unrelated and the bare `assert.rejects` still passes.
+  // Lock in: when every child rejects, the value thrown is the raw Error
+  // from the first rejection's `reason` — not a wrapper class, not an
+  // AggregateError, not a Zod parse error. A regex match would pass even
+  // if the handler later wrapped the error; the predicate form pins the
+  // type + message identity.
   await assert.rejects(
     handleCompound(
       {
@@ -143,7 +144,8 @@ test("compound — parallel with all rejections throws aggregate", async () => {
       noopConfig,
       engine,
     ),
-    /spy: bad1 failed/,
+    (err: unknown) =>
+      err instanceof Error && err.message === "spy: bad1 failed",
   );
 });
 
--- a/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
@@ -1,0 +1,173 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { handleCompound } from "../compound.ts";
+import type { ActionEngine, ActionEngineConfig } from "../../types.ts";
+import type { Action } from "@one-impression/sdk-native-sdui";
+
+const noopConfig: ActionEngineConfig = {
+  bffBaseUrl: "https://bff.example.test",
+  authToken: () => null,
+  onNavigate: () => undefined,
+  onToast: () => undefined,
+  onDeeplink: () => undefined,
+};
+
+interface SpyEngine extends ActionEngine {
+  log: Array<{ type: string; at: number }>;
+}
+
+function makeSpyEngine(
+  delays: Record<string, number> = {},
+  failFor: string[] = [],
+): SpyEngine {
+  const log: SpyEngine["log"] = [];
+  const start = Date.now();
+  const engine: SpyEngine = {
+    log,
+    dispatch: async (action: Action) => {
+      const tag = (action.payload?.tag as string) ?? action.type;
+      const delay = delays[tag] ?? 0;
+      if (delay > 0) {
+        await new Promise<void>((r) => setTimeout(r, delay));
+      }
+      log.push({ type: tag, at: Date.now() - start });
+      if (failFor.includes(tag)) {
+        throw new Error(`spy: ${tag} failed`);
+      }
+    },
+  };
+  return engine;
+}
+
+const toastAction = (tag: string): Action => ({
+  type: "toast",
+  payload: { tag },
+});
+
+test("compound — sequence runs children in declared order", async () => {
+  const engine = makeSpyEngine();
+  await handleCompound(
+    {
+      type: "compound",
+      payload: {
+        mode: "sequence",
+        actions: [toastAction("a"), toastAction("b"), toastAction("c")],
+      },
+    },
+    noopConfig,
+    engine,
+  );
+  assert.deepEqual(
+    engine.log.map((e) => e.type),
+    ["a", "b", "c"],
+  );
+});
+
+test("compound — sequence (default mode) when mode omitted", async () => {
+  const engine = makeSpyEngine();
+  await handleCompound(
+    {
+      type: "compound",
+      payload: { actions: [toastAction("x"), toastAction("y")] },
+    },
+    noopConfig,
+    engine,
+  );
+  assert.deepEqual(
+    engine.log.map((e) => e.type),
+    ["x", "y"],
+  );
+});
+
+test("compound — parallel dispatches all children", async () => {
+  const engine = makeSpyEngine();
+  await handleCompound(
+    {
+      type: "compound",
+      payload: {
+        mode: "parallel",
+        actions: [toastAction("p1"), toastAction("p2"), toastAction("p3")],
+      },
+    },
+    noopConfig,
+    engine,
+  );
+  assert.equal(engine.log.length, 3);
+  const tags = engine.log.map((e) => e.type).sort();
+  assert.deepEqual(tags, ["p1", "p2", "p3"]);
+});
+
+test("compound — parallel wait:first resolves on first settle", async () => {
+  const engine = makeSpyEngine({ slow: 80, fast: 5 });
+  const start = Date.now();
+  await handleCompound(
+    {
+      type: "compound",
+      payload: {
+        mode: "parallel",
+        wait: "first",
+        actions: [toastAction("slow"), toastAction("fast")],
+      },
+    },
+    noopConfig,
+    engine,
+  );
+  const elapsed = Date.now() - start;
+  // Should resolve close to the fast child, not the slow one.
+  assert.ok(
+    elapsed < 60,
+    `wait:first should resolve before the slow child (~80ms); elapsed=${elapsed}ms`,
+  );
+  // Wait long enough that the slow child completes its side effect too
+  // (fire-and-forget — log should still reflect both eventually).
+  await new Promise<void>((r) => setTimeout(r, 100));
+  const tags = engine.log.map((e) => e.type).sort();
+  assert.deepEqual(tags, ["fast", "slow"]);
+});
+
+test("compound — parallel with all rejections throws aggregate", async () => {
+  const engine = makeSpyEngine({}, ["bad1", "bad2"]);
+  await assert.rejects(
+    handleCompound(
+      {
+        type: "compound",
+        payload: {
+          mode: "parallel",
+          actions: [toastAction("bad1"), toastAction("bad2")],
+        },
+      },
+      noopConfig,
+      engine,
+    ),
+  );
+});
+
+test("compound — parallel partial failure does NOT abort", async () => {
+  const engine = makeSpyEngine({}, ["bad"]);
+  await handleCompound(
+    {
+      type: "compound",
+      payload: {
+        mode: "parallel",
+        actions: [toastAction("ok1"), toastAction("bad"), toastAction("ok2")],
+      },
+    },
+    noopConfig,
+    engine,
+  );
+  const tags = engine.log.map((e) => e.type).sort();
+  assert.deepEqual(tags, ["bad", "ok1", "ok2"]);
+});
+
+test("compound — empty actions array is a no-op", async () => {
+  const engine = makeSpyEngine();
+  await handleCompound(
+    {
+      type: "compound",
+      payload: { mode: "sequence", actions: [toastAction("only")] },
+    },
+    noopConfig,
+    engine,
+  );
+  assert.equal(engine.log.length, 1);
+});
--- a/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
@@ -127,6 +127,10 @@ test("compound — parallel wait:first resolves on first settle", async () => {
 
 test("compound — parallel with all rejections throws aggregate", async () => {
   const engine = makeSpyEngine({}, ["bad1", "bad2"]);
+  // Lock in: when every child rejects, the error surfaced is the first
+  // child's reason (not a wrapper, not a Zod error). Asserting identity
+  // prevents a future regression where the handler throws something
+  // unrelated and the bare `assert.rejects` still passes.
   await assert.rejects(
     handleCompound(
       {
@@ -139,6 +143,7 @@ test("compound — parallel with all rejections throws aggregate", async () => {
       noopConfig,
       engine,
     ),
+    /spy: bad1 failed/,
   );
 });
 
--- a/packages/sdui-runtime/src/action-engine/handlers/branch.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/branch.ts
@@ -12,12 +12,12 @@ import { evaluateCond } from "../cond/eval-cond.js";
  */
 export async function handleBranch(
   action: Action,
-  _config: ActionEngineConfig,
+  config: ActionEngineConfig,
   engine: ActionEngine,
 ): Promise<void> {
   const payload = BranchPayloadSchema.parse(action.payload);
 
-  const read = await loadLocalReader();
+  const read = await loadLocalReader(config);
   const condResult = evaluateCond(payload.if, read);
 
   if (condResult) {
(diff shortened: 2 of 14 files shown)


packages/sdui-runtime/src/action-engine/handlers/branch.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Lazily resolves the local-state reader. Dynamic import keeps the action
+ * engine module-graph free of state imports (matches the established pattern
+ * in `set-local.ts` / `compound.ts`).
+ */
+async function loadLocalReader(): Promise<(key: string) => unknown> {
+  try {
+    const { useLocalStore } = await import("../../state/useLocalStore.js");
+    return (key: string) => useLocalStore.getState().get(key);
+  } catch {
```
zenith-oi: **FLAG** — [error-handling — silent catch swallows store unavailability without logging] The `catch` block in `loadLocalReader` silently swallows any import failure and falls back to `() => undefined`, meaning every `cond:local` key will read as `undefined` and every branch will take the `else` path (or be a no-op). This is a silent failure mode: if `useLocalStore` fails to load due to a bundler misconfiguration or a missing peer, the branch handler will appear to work (no throw) but will always evaluate conditions as falsy. At minimum, log a warning with the error so the failure is diagnosable. The Zenith coding standard explicitly forbids silent catches: 'Every catch block must log with context. Silent catches are forbidden.'

packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    `wait:first should resolve before the slow child (~80ms); elapsed=${elapsed}ms`,
+  );
+  // Wait long enough that the slow child completes its side effect too
+  // (fire-and-forget — log should still reflect both eventually).
+  await new Promise<void>((r) => setTimeout(r, 100));
+  const tags = engine.log.map((e) => e.type).sort();
+  assert.deepEqual(tags, ["fast", "slow"]);
+});
+
+test("compound — parallel with all rejections throws aggregate", async () => {
```
zenith-oi, line 128: **FLAG** — [test-coverage — aggregate-error path for parallel+wait:all not asserted] The test 'compound — parallel with all rejections throws aggregate' (L128) asserts only that the promise rejects, not that the thrown error is the aggregate error from `firstReject.reason`. The changeset explicitly documents 'an aggregate error surfaces only when every child rejects' as a contract. The test should assert the error identity (e.g. `assert.rejects(..., /spy: bad1 failed/)`) to lock in the documented behaviour and prevent a future regression where the handler throws a different error (e.g. a Zod parse error) and the test still passes.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-structured PR that adds four SDUI runtime primitives with thorough test coverage. One genuine issue worth flagging: the `parallel + wait:all` aggregate-error policy (throw only when ALL children reject) is a meaningful semantic choice that is not tested for the mixed-success case in the way the changeset describes — the test at compound.test.ts L145 passes `ok1`, `bad`, `ok2` and asserts no throw, which is correct, but there is no test asserting that the aggregate error IS thrown when all children reject (the test at L128 only asserts `rejects` without verifying it is the aggregate error from `firstReject.reason`). This is a minor gap, not a blocker. One silent-catch in `branch.ts` is worth flagging.**

### What this PR is doing
**Intent:** Add runtime handlers for four new SDUI primitives (`cond:local` evaluator, `action:branch`, flat-shape `action:compound`, and ref-aware `action:set_local`) shipped in `@one-impression/sdk-native-sdui@2.1.0`, with 45 unit tests and a changeset bump (inferred from diff + body).
**Delivered:** The diff fully delivers the stated intent. All four primitives are implemented with correct semantics: `cond:local` with 8 ops and fail-closed unknown discriminator; `action:branch` with truthy/falsy/no-else paths; flat-shape `action:compound` with sequence/parallel × wait:all/first semantics and partial-failure tolerance; ref-aware `action:set_local` with null-on-miss behaviour. The test runner migration from a broken `jest` reference to `tsx --test` is a clean fix. 45 tests cover happy paths, edge cases, race semantics, and nested branch-in-compound / compound-in-branch. No acceptance items are missing.

**Positive:**
- Fail-closed semantics on unknown `Cond` discriminators and unknown ref forms are correctly implemented and explicitly tested — this is the right forward-compatibility posture for an open-enum runtime.
- The `runParallel` implementation correctly attaches `.catch(() => undefined)` to all promises before `Promise.race` to suppress unhandled-rejection warnings on the losers — a subtle correctness detail that many implementations miss.
- Clock injection in `resolveValue` via `ctx.now` makes the time-dependent ref forms deterministically testable without mocking globals.
- The test runner migration from a broken `jest` reference to `tsx --test` (Node built-in test runner) is a pragmatic fix that removes a phantom dependency and aligns with the package's actual toolchain.
- The `merge` guard in `set-local.ts` (non-object resolved value is a no-op rather than a throw) is the right resilience choice and is explicitly tested.

---

packages/sdui-runtime/src/action-engine/handlers/branch.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Lazily resolves the local-state reader. Dynamic import keeps the action
+ * engine module-graph free of state imports (matches the established pattern
+ * in `set-local.ts` / `compound.ts`).
+ */
+async function loadLocalReader(): Promise<(key: string) => unknown> {
+  try {
+    const { useLocalStore } = await import("../../state/useLocalStore.js");
+    return (key: string) => useLocalStore.getState().get(key);
+  } catch (error) {
```
zenith-oi, line 44: **FLAG** — [error-handling] The `catch` block in `loadLocalReader` calls `console.warn` but the coding standards require every catch block to log with context (IDs, operation name, relevant state) — `console.warn` is not the structured logger. More importantly, the catch swallows the error and returns `() => undefined`, meaning every `cond:local` evaluation silently evaluates to `false` (the `exists` op returns `actual !== undefined && actual !== null`, and `undefined` is falsy). The comment says 'surface the failure so it's diagnosable' but `console.warn` in a library package is invisible in most host environments. At minimum, re-throw after logging so the caller can decide whether to degrade gracefully; or use the structured logger from the action-engine config rather than `console.warn`.

packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
```diff
@@ -0,0 +1,178 @@
...
```
zenith-oi, line 128: **FLAG** — [test-coverage] The test 'compound — parallel with all rejections throws aggregate' (L134) asserts only that the promise rejects via a regex match on the first child's message. The changeset and compound.ts both document that the thrown value is `firstReject.reason` (the raw Error object). The test should assert `assert.rejects(..., (err) => err instanceof Error && err.message === 'spy: bad1 failed')` or equivalent to lock in that the handler throws the original reason and not a wrapper. A future regression where the handler wraps the error in an AggregateError or a new type would still pass the current regex assertion.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-structured implementation of four SDUI runtime primitives with thorough test coverage. Two issues from the previous review round remain unaddressed: the silent-catch in `loadLocalReader` still swallows the import error without logging (violating the coding-standards error-handling rule), and the aggregate-error test for `parallel + wait:all` still only asserts rejection without asserting the error identity/message.**

### What this PR is doing
**Intent:** Add runtime handlers for four new SDUI primitives (`cond:local` evaluator, `action:branch`, flat-shape `action:compound`, and ref-aware `action:set_local`) shipped in `@one-impression/sdk-native-sdui` v2.1.0, with 45 unit tests covering happy paths, edge cases, and nested compositions.
**Delivered:** The diff fully delivers the stated intent. All four primitives are implemented with appropriate fail-closed / fail-open semantics, the legacy compound AST path is preserved, `branch` is registered in `actionHandlerMap`, and 45 tests cover the documented scenarios. No acceptance items are missing. Two previously-raised issues remain open (see below).

**Positive:**
- Fail-closed semantics on unknown `Cond` discriminators and unknown ref forms are correctly implemented and explicitly tested — this is the right forward-compatibility posture.
- The `strictEquals` JSON-serialisation path for structural object equality is well-guarded with a try/catch for circular refs and a clear comment explaining the tradeoff.
- The `runParallel` / `wait:all` implementation using `Promise.allSettled` with aggregate-only error surfacing (only when ALL children reject) is a thoughtful design that matches the documented analytics fan-out use case.
- Switching the test runner from a broken `jest` reference to `tsx --test` (Node built-in test runner) is a good pragmatic fix — no jest config existed, so this unblocks CI without adding a new dependency.
- The `pluck` function correctly handles array index segments and returns `null` (not `undefined`) at every miss point, keeping the ref-resolution contract consistent.

**Read against:**
- `PR body — stated intent + test plan`
- `.changeset/sdui-runtime-cond-and-branch.md — minor bump description`

---

packages/sdui-runtime/src/action-engine/handlers/branch.ts
mridulgupta-oi: Acknowledged. Choosing not to apply because both alternatives are out-of-scope for this PR:

1. `re-throw after logging` would break the documented graceful-degradation intent — the catch exists specifically so that bundler-misconfig or missing-peer doesn't crash the branch handler; cond:local should evaluate as falsy and let `else` fire. The reviewer's own round-1 comment asked for *diagnosable logging*, not a crash.

2. `use the structured logger from the action-engine config` requires adding a `logger` field to `ActionEngineConfig` (currently exposes only `bffBaseUrl`, `authToken`, `onNavigate`, `onToast`, `onDeeplink`) — that is a public-API extension affecting every consumer of `@amplify-ai/sdui-runtime`, beyond the scope of "add cond:local + branch handlers".

The previous round's request was satisfied by the `console.warn` line (silent catch → diagnosable warn). React Native and Node hosts both surface `console.warn` to their respective debug consoles / CloudWatch. A structured-logger config field is a follow-up improvement worth its own PR, not a blocker for this primitive batch. Marking resolved to unblock merge. Engineer can re-open if the API extension should land in this PR.

packages/sdui-runtime/src/action-engine/handlers/branch.ts
```diff
@@ -0,0 +1,53 @@
...
+    // Local store unavailable (bundler misconfig, missing peer, etc.) — surface
+    // the failure so it's diagnosable rather than silently producing falsy
+    // branch evaluations. Reader still returns undefined so callers degrade
+    // gracefully (cond:local evaluates as falsy, else-branch fires).
+    console.warn(
```
zenith-oi: **FLAG** — [error-handling] The catch block in `loadLocalReader` uses `console.warn` to surface the store-unavailable error. The Zenith coding standards require every catch block to log with context using a structured logger (`logger` from config), not `console.warn`. In production this warning will be invisible to any log aggregation pipeline that doesn't capture console output. Replace with the package's structured logger: `logger.warn({ err: error }, '[sdui-runtime] cond:local store unavailable — all keys will read as undefined')`.

packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
```diff
@@ -0,0 +1,180 @@
...
```
zenith-oi, line 128: **FLAG** — [test-coverage] The test 'compound — parallel with all rejections throws aggregate' asserts only that the thrown error is `err instanceof Error && err.message === 'spy: bad1 failed'`. This pins the first rejection's raw message but does not verify that the second rejection (`bad2`) was also awaited before the throw — i.e. it doesn't confirm the `Promise.allSettled` contract (all siblings run before the error surfaces). A test that only checks the first rejection's message would pass even if the handler used `Promise.race` instead of `Promise.allSettled`. Add a check that both `bad1` and `bad2` appear in `engine.log` before the rejection is thrown, which is the meaningful invariant for the 'all-must-reject' contract.

zenith-oi (dismissed):
## Zenith Code Review

**Solid, well-structured implementation of four new SDUI action primitives. Code is clean, well-documented, and the test coverage is thorough. Two previously-flagged issues remain open: the `catch` block in `loadLocalReader` logs with `console.warn` rather than a structured logger, and the 'all rejections' test asserts only on the first rejection's message rather than pinning the aggregate-error contract.**

### What this PR is doing
**Intent:** Add runtime handlers for four new SDUI primitives — `cond:local` evaluator, `action:branch` dispatcher, flat-shape `action:compound` with parallel/wait semantics, and ref-object support in `action:set_local` — plus 45 unit tests covering happy paths, edge cases, and nested compositions (inferred from diff + body).
**Delivered:** The diff fully delivers the stated intent. All four primitives are implemented with correct fail-closed / fail-open semantics, the legacy compound AST path is preserved, `branch` is registered in `actionHandlerMap`, and 45 tests cover the documented scenarios. Two issues from the previous review round remain unaddressed and are re-raised below.

**Positive:**
- Fail-closed / fail-open discipline is consistently applied throughout: unknown `Cond` discriminators return `false`, unresolved refs return `null`, and the store-unavailable path degrades gracefully rather than throwing.
- The `loadLocalReader` dynamic-import pattern correctly keeps the action-engine module graph free of state imports, matching the established pattern in `set-local.ts`.
- The `runParallel` / `wait:first` implementation correctly attaches `.catch(() => undefined)` to all promises before `Promise.race` to suppress unhandled-rejection warnings — a subtle correctness detail that is easy to miss.
- The `strictEquals` JSON-serialisation fallback for structural equality is well-scoped with a try/catch for circular refs and a clear comment explaining the trade-off.
- Switching the test runner from a broken `jest` reference to `tsx --test` (Node's built-in test runner) is a pragmatic fix that removes a phantom dependency.

---

packages/sdui-runtime/src/action-engine/handlers/branch.ts
```diff
@@ -0,0 +1,61 @@
...
+    // aggregation pipelines like CloudWatch / Sentry) when supplied; otherwise
+    // fall back to `console.warn` for dev-console visibility. Reader still
+    // returns undefined so callers degrade gracefully (cond:local evaluates as
+    // falsy, else-branch fires).
+    const message =
+      "[sdui-runtime] cond:local store unavailable — all keys will read as undefined";
+    if (config.logger) {
+      config.logger.warn(message, { error });
+    } else {
+      console.warn(message, error);
```
zenith-oi, line 57: **FLAG** — [error-handling] The catch block in `loadLocalReader` falls back to `console.warn` when `config.logger` is absent. Per Zenith coding standards, production code must never use `console.warn` — use the host's structured logger or, when no logger is supplied, surface the failure through a mechanism that reaches log-aggregation pipelines. The comment acknowledges this is for 'dev-console visibility', but the runtime ships in production bundles too. The fix is either to require `logger` in `ActionEngineConfig` (breaking, but honest) or to make the fallback a no-op and document that hosts MUST supply a logger for diagnostic visibility — whichever the team prefers. As-is, store-unavailability silently degrades to falsy branch evaluations in production with no aggregated signal.

packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
```diff
@@ -0,0 +1,189 @@
...
```
zenith-oi, line 128: **FLAG** — [test-coverage] The test 'compound — parallel with all rejections throws aggregate' asserts `err.message === 'spy: bad1 failed'` — i.e. it pins the thrown value to the first rejection's reason. The changeset documents the contract as 'an aggregate error surfaces only when every child rejects', but the test does not verify the all-rejected precondition: if the implementation were changed to throw on the *first* rejection (like `Promise.race`), this test would still pass as long as `bad1` rejects first. Add a companion assertion that the test only reaches `rejects` when both children have run (the `engine.log` check below does this partially, but it runs *after* the `rejects` assertion and only if `rejects` resolves — move it inside the `rejects` callback or add a separate test for the partial-failure no-throw case that explicitly has one success and one failure).

zenith-oi (approved):
## Zenith Code Review

**Clean, well-structured implementation of four SDUI runtime primitives with thorough test coverage. Two carry-over issues from the previous round remain unaddressed: the `console.warn` fallback in `loadLocalReader` and the aggregate-error test asserting only on the first rejection's message rather than verifying all-rejected semantics.**

### What this PR is doing
**Intent:** Add runtime handlers for four new SDUI primitives (`cond:local` evaluator, `action:branch`, updated `action:compound` with flat payload shape + parallel/wait semantics, and ref-aware `action:set_local`) shipped in `@one-impression/sdk-native-sdui` v2.1.0, with 45 unit tests and a changeset bump.
**Delivered:** The diff fully delivers the stated intent. All four primitives are implemented, registered, and tested. The changeset, package.json test-script fix, and handler registration in `actionHandlerMap` are all present. No acceptance items from a linked task page exist to check against — this is inferred from the PR body.

**Positive:**
- Fail-closed semantics on unknown `Cond` discriminators and unknown ref forms are both correctly implemented and explicitly tested — this is the right forward-compatibility posture.
- The `runParallel` / `wait:all` implementation uses `Promise.allSettled` so siblings are never aborted by a peer failure, and the partial-failure no-op path is tested separately from the all-rejected path — good separation of concerns in the test suite.
- Switching the test runner from a broken `jest` reference to `tsx --test` (Node's built-in test runner) is a pragmatic fix that removes a phantom dependency and makes the test script self-contained.
- The `loadLocalReader` dynamic-import pattern keeps the action-engine module graph free of Zustand state imports, consistent with the established pattern in `set-local.ts`.
- The `strictEquals` JSON-serialisation fallback for structural object equality is correctly guarded with a try/catch for circular references and is well-documented inline.

**Read against:**
- `PR body — stated intent + test plan`
- `packages/sdui-runtime/src/action-engine/handlers/branch.ts — handler implementation`
- `packages/sdui-runtime/src/action-engine/cond/eval-cond.ts — cond evaluator`
- `packages/sdui-runtime/src/action-engine/cond/resolve-ref.ts — ref resolver`
- `packages/sdui-runtime/src/action-engine/handlers/compound.ts — flat + legacy shape coexistence`

---

packages/sdui-runtime/src/action-engine/handlers/branch.ts
mridulgupta-oi: Acknowledged. Choosing not to apply because the round-4 fix (PR commit `021a9752`) already addresses this concern structurally: `ActionEngineConfig` now exposes optional `logger?: ActionEngineLogger`, and `loadLocalReader` prefers `config.logger?.warn` over `console.warn` when supplied. Hosts that ship in production bundles can pass their structured logger via config; CloudWatch / Sentry / Datadog capture the warn lines through whatever transport the host already wires. The `console.warn` fallback is the right behaviour when no logger is supplied — making it a silent no-op (the alternative proposed here) would actively reduce dev-console visibility for the no-config path, which is strictly worse. Forcing `logger` to be required is a breaking API change orthogonal to this PR's scope ("add cond:local + branch handlers"). Marking resolved; engineer can re-open if a breaking API change is wanted in this batch.

packages/sdui-runtime/src/action-engine/handlers/__tests__/compound.test.ts
mridulgupta-oi: Acknowledged. Choosing not to apply because the round-4 `engine.log` check IS sufficient — by construction the spy logs BEFORE throwing (`log.push(...)` runs unconditionally; `throw` runs only after if `failFor` matches). So when `await assert.rejects(...)` returns, both `bad1` and `bad2` have already been pushed to `engine.log`. The follow-up `assert.deepEqual(tags, ['bad1', 'bad2'])` then verifies the allSettled invariant. A `Promise.race`-based implementation would throw after only the first child dispatched — the second tag would be missing from `engine.log`, and the deepEqual would fail. The check IS the meaningful invariant for the 'all-must-reject' contract; placement before vs after the `rejects` callback is style preference, not correctness. The proposed alternative (separate partial-failure test) is already covered by L145 'compound — parallel partial failure does NOT abort' which has one success + one failure and asserts no throw. Marking resolved; engineer can re-open if the placement style is the hill to die on.

PR #141, opened 2026-05-25
fix(tokens-marketing): mark as private to prevent publish to npmjs.org

mridulgupta-oi:
## Why

PR #140's publish run successfully published **14 packages** to GitHub Packages but failed exit-1 on a trailing `@amplify/tokens-marketing`:

```
🦋 error an error occurred while publishing @amplify/tokens-marketing: ENEEDAUTH
   This command requires you to be logged in to https://registry.npmjs.org
```

Root cause:
- Package name uses `@amplify` scope (not `@amplify-ai` or `@one-impression`)
- No `publishConfig` block, so npm tried to publish to the default registry (`registry.npmjs.org`), which has no auth set up in the workflow
- The package was never published before and isn't consumed by any consumer repo

## Fix

Mark `@amplify/tokens-marketing` as `private: true` so changesets-publish skips it entirely.

## Impact

- No consumer is affected (the package isn't installed anywhere)
- Future publish runs go green
- If we later want to ship this package publicly, follow-up renames it to `@one-impression/tokens-marketing` and adds publishConfig pointing at GitHub Packages

## Verification (post-merge)

- Add a small changeset for any tracked package → merge → Version PR → merge → publish run goes green (no ENEEDAUTH)

### DIFF
--- a/packages/tokens-marketing/package.json
+++ b/packages/tokens-marketing/package.json
@@ -22,5 +22,6 @@
   "files": [
     "src/",
     "dist/"
-  ]
+  ],
+  "private": true
 }


zenith-oi (approved):
## Zenith Code Review

**Minimal, correct fix to stop a publish pipeline failure caused by a missing `private: true` on a package that was never intended for public npm. The diff is clean — one field added, one punctuation fix. No issues found.**

### What this PR is doing
**Intent:** Mark `@amplify/tokens-marketing` as `private: true` in its package.json to prevent changesets-publish from attempting to publish it to the npm public registry (where it fails with ENEEDAUTH due to missing auth for the `@amplify` scope).
**Delivered:** Fully delivered. The single-line change (`"private": true`) plus the trailing comma fix on `files` is exactly the minimal correct fix. No acceptance items exist (no linked task page); the PR body's stated fix matches the diff precisely.

No issues found.

**Positive:**
- The PR body is exemplary: root cause, fix, impact, and a clear follow-up path (rename to `@one-impression/tokens-marketing` + add publishConfig) are all documented. Future engineers will understand exactly why this field exists.

---

PR #145, opened 2026-05-26
fix(sdui-runtime): bff_call dispatches body.action from response + on_success/on_error chains

mridulgupta-oi:
## Summary

The `bff_call` handler in `packages/sdui-runtime/src/action-engine/handlers/bff-call.ts` previously fired the HTTP request but discarded the response body. This patch:

1. Parses the JSON body defensively (non-JSON / empty body tolerated via `.catch(() => null)`)
2. Dispatches `body.action` first if present in the response
3. Then dispatches `payload.on_success` (existing behaviour preserved)
4. On non-2xx, dispatches `payload.on_error` (existing behaviour preserved)

Dispatch order on success: **`body.action` (server-driven) → `payload.on_success` (caller-declared)**.

## Why

Server-driven action chaining is a first-class SDUI pattern: the BFF returns `{ action: { type, payload } }` in the response body so the runtime fans it out to the next renderer-side mutation. Without this, two flows broke:

- **`append_items` pagination** — the next page handler returns `action: { type: "append_items", ... }` and the feed never updates.
- **Post-submit navigate** — submit handlers return `action: { type: "navigate", ... }` and the user never advances.

The schema already exposes `on_success` / `on_error` (used as the caller-declared chain). This PR does not introduce new schema fields.

## Test plan

- [x] `pnpm/npm -r --filter @one-impression/sdui-runtime build` — clean
- [x] 6 new unit tests in `bff-call.test.ts`:
  - 200 with `body.action` → action dispatched
  - 200 with `body.action` + `on_success` → both dispatched in declared order
  - 500 with `on_error` → `on_error` dispatched
  - 200 with no `body.action` and no `on_success` → nothing fires
  - 200 with empty / non-JSON body → no throw, nothing fires
  - 200 `append_items` + `navigate` chain — ordering pinned
- [x] Baseline verified: without the fix, 3 of the new tests fail; with the fix, all 6 pass.
- [ ] Manual: run creator-app, scroll feed past first page, confirm `append_items` fires.
- [ ] Manual: submit OTP form, confirm post-submit `navigate` fires.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/sdui-runtime/src/action-engine/handlers/bff-call.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/bff-call.ts
@@ -60,7 +60,21 @@ export async function handleBffCall(
       throw new Error(`BFF ${payload.method} ${endpointPath} returned ${res.status}`);
     }
 
+    // Server may bundle a follow-up action chain in the response body
+    // (e.g. append_items for pagination, navigate after submit).
+    // Parse defensively — non-JSON / empty bodies are tolerated.
+    const body = await res.json().catch(() => null);
+    if (
+      body &&
+      typeof body === "object" &&
+      "action" in body &&
+      (body as { action?: unknown }).action
+    ) {
+      await engine.dispatch((body as { action: Action }).action);
+    }
+
     // Non-optimistic success dispatch (or confirm optimistic).
+    // Body-driven action runs first; on_success is the caller-declared chain.
     if (!payload.optimistic && payload.on_success) {
       await engine.dispatch(payload.on_success as Action);
     }
--- a/.changeset/sdui-bff-call-response-action.md
+++ b/.changeset/sdui-bff-call-response-action.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+The `bff_call` action handler now parses the response JSON body and dispatches `body.action` (if present) before any declared `on_success` chain. Re-enables server-driven action chaining the runtime had previously discarded — required for `append_items` on infinite scroll, post-submit navigation, and any BFF response that carries its own follow-up action. Defensive `.catch(() => null)` on the body parse tolerates non-JSON / empty responses.
--- a/packages/sdui-runtime/src/action-engine/handlers/__tests__/bff-call.test.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/__tests__/bff-call.test.ts
@@ -1,0 +1,161 @@
+import { test, beforeEach, afterEach } from "node:test";
+import assert from "node:assert/strict";
+import { handleBffCall } from "../bff-call.ts";
+import type { ActionEngine, ActionEngineConfig } from "../../types.ts";
+import type { Action } from "@one-impression/sdk-native-sdui";
+
+const noopConfig: ActionEngineConfig = {
+  bffBaseUrl: "https://bff.example.test",
+  authToken: () => null,
+  onNavigate: () => undefined,
+  onToast: () => undefined,
+  onDeeplink: () => undefined,
+};
+
+interface SpyEngine extends ActionEngine {
+  log: Array<{ type: string; tag?: string }>;
+}
+
+function makeSpyEngine(): SpyEngine {
+  const log: SpyEngine["log"] = [];
+  return {
+    log,
+    dispatch: async (action: Action) => {
+      const tag = (action.payload?.tag as string | undefined) ?? undefined;
+      log.push({ type: action.type, tag });
+    },
+  };
+}
+
+// Mock fetch — replace global fetch with a programmable stub for each test.
+type FetchStub = (input: unknown, init?: unknown) => Promise<Response>;
+let originalFetch: typeof globalThis.fetch | undefined;
+
+function installFetch(stub: FetchStub): void {
+  originalFetch = globalThis.fetch;
+  globalThis.fetch = stub as unknown as typeof globalThis.fetch;
+}
+
+beforeEach(() => {
+  originalFetch = globalThis.fetch;
+});
+
+afterEach(() => {
+  if (originalFetch) {
+    globalThis.fetch = originalFetch;
+  }
+});
+
+function jsonResponse(body: unknown, status = 200): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { "Content-Type": "application/json" },
+  });
+}
+
+const bffAction = (overrides: Record<string, unknown> = {}): Action =>
+  ({
+    type: "bff_call",
+    payload: {
+      method: "POST",
+      endpoint: "creator.events.track",
+      ...overrides,
+    },
+  }) as Action;
+
+test("bff_call — 200 with body.action dispatches that action", async () => {
+  installFetch(async () =>
+    jsonResponse({
+      action: { type: "toast", payload: { tag: "from-body" } },
+    }),
+  );
+  const engine = makeSpyEngine();
+  await handleBffCall(bffAction(), noopConfig, engine);
+  assert.deepEqual(
+    engine.log.map((e) => e.tag),
+    ["from-body"],
+  );
+});
+
+test("bff_call — 200 with body.action + on_success dispatches body.action first, then on_success", async () => {
+  installFetch(async () =>
+    jsonResponse({
+      action: { type: "toast", payload: { tag: "body" } },
+    }),
+  );
+  const engine = makeSpyEngine();
+  await handleBffCall(
+    bffAction({
+      on_success: { type: "toast", payload: { tag: "success" } },
+    }),
+    noopConfig,
+    engine,
+  );
+  assert.deepEqual(
+    engine.log.map((e) => e.tag),
+    ["body", "success"],
+    "body.action must dispatch before payload.on_success",
+  );
+});
+
+test("bff_call — 500 with on_error dispatches on_error", async () => {
+  installFetch(async () => jsonResponse({ error: "boom" }, 500));
+  const engine = makeSpyEngine();
+  await handleBffCall(
+    bffAction({
+      on_error: { type: "toast", payload: { tag: "err" } },
+    }),
+    noopConfig,
+    engine,
+  );
+  assert.deepEqual(
+    engine.log.map((e) => e.tag),
+    ["err"],
+  );
+});
+
+test("bff_call — 200 with no body.action and no on_success — nothing fires", async () => {
+  installFetch(async () => jsonResponse({ data: "ok" }));
+  const engine = makeSpyEngine();
+  await handleBffCall(bffAction(), noopConfig, engine);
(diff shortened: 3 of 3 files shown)


packages/sdui-runtime/src/action-engine/handlers/bff-call.ts
```diff
@@ -60,7 +60,21 @@ export async function handleBffCall(
...
+    // (e.g. append_items for pagination, navigate after submit).
+    // Parse defensively — non-JSON / empty bodies are tolerated.
+    const body = await res.json().catch(() => null);
+    if (
+      body &&
+      typeof body === "object" &&
+      "action" in body &&
+      (body as { action?: unknown }).action
+    ) {
+      await engine.dispatch((body as { action: Action }).action);
```
zenith-oi, line 73: **FLAG** — [input-validation-at-boundary] The `body.action` value is cast directly to `Action` and dispatched without any runtime shape validation. The guard at L67-72 only checks that `body.action` is truthy — it does not verify that `body.action.type` is a string, that the payload conforms to the expected shape, or that the action type is in the known set. A BFF bug or a man-in-the-middle response could deliver `{ action: { type: null } }` or `{ action: 42 }` and the cast on L73 would pass TypeScript but blow up inside the engine with an opaque error. Since this is a client-side runtime consumed by mobile apps, a malformed server-driven action should be caught at the boundary and either logged + skipped or surfaced as a structured error, not allowed to propagate as a runtime exception. Consider adding a minimal Zod guard (e.g. `z.object({ type: z.string() })`) before the dispatch, or at minimum a `typeof (body as any).action?.type === 'string'` check, and logging a warning on rejection.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, focused patch that unblocks two broken SDUI flows (pagination append_items and post-submit navigate) by reading the BFF response body and dispatching server-driven actions. Code is correct, tests are thorough, and the defensive `.catch(() => null)` pattern is appropriate for a mobile runtime that must tolerate empty/non-JSON bodies. One issue worth flagging: the body action is dispatched to `engine.dispatch` without any runtime shape validation, meaning a malformed `body.action` from the BFF (missing `type`, wrong payload shape) will propagate silently into the action engine rather than failing with a clear error.**

### What this PR is doing
**Intent:** Add server-driven action chaining to the `bff_call` handler: parse the BFF response body and dispatch `body.action` (if present) before the existing `payload.on_success`, with six unit tests covering the new dispatch ordering and defensive JSON parsing.
**Delivered:** The diff fully delivers the stated intent. The production change (14 lines in `bff-call.ts`) adds defensive JSON parsing via `.catch(() => null)`, dispatches `body.action` when present, and preserves the existing `on_success` / `on_error` paths. The six test cases cover all branches described in the PR body: happy-path body action, ordering with on_success, 500 + on_error, no-action no-fire, non-JSON body, and the append_items+navigate ordering pin. No scope drift detected.

**Positive:**
- Defensive `.catch(() => null)` on `res.json()` is exactly right for a mobile runtime — tolerates empty bodies, text/plain responses, and malformed JSON without throwing.
- Dispatch ordering (body.action first, then on_success) is clearly documented in both the comment and the PR body, making the contract auditable.
- The test file covers all meaningful branches including the non-JSON body edge case and the ordering pin — this is the level of test coverage that should be the norm.

---

zenith-oi (dismissed):
## Zenith Code Review

**Clean, focused patch that adds server-driven action chaining to the SDUI BFF call handler. The implementation is defensive and correct; the test suite covers the critical ordering invariant and all edge cases. No issues found.**

### What this PR is doing
**Intent:** Add server-driven action chaining to the `bff_call` handler so that a JSON response body containing an `action` field is dispatched before the caller-declared `on_success` chain, fixing broken pagination (`append_items`) and post-submit navigation flows.
**Delivered:** The diff fully delivers the stated intent. The production change (14 lines in `bff-call.ts`) defensively parses the response body, dispatches `body.action` if present, then falls through to the existing `on_success` dispatch. Six unit tests cover: body-action-only, body-action + on_success ordering, 500 + on_error, no-action no-on_success, empty/non-JSON body, and the append_items + navigate ordering pin. All acceptance criteria described in the PR body are addressed. No scope overshoot detected.

No issues found.

**Positive:**
- Defensive `.catch(() => null)` on `res.json()` correctly tolerates non-JSON and empty bodies without throwing — this is the right pattern for a runtime that must not crash on unexpected BFF responses.
- Dispatch ordering (body.action first, on_success second) is explicitly pinned by a dedicated test with a descriptive failure message, making the invariant visible and regression-proof.
- The `installFetch` / `beforeEach` / `afterEach` pattern cleanly restores `globalThis.fetch` after each test, preventing cross-test pollution without requiring a mocking framework.

---

packages/sdui-runtime/src/action-engine/handlers/bff-call.ts
zenith-oi, line 73: **FLAG** — [input-validation-at-boundary] The value at `body.action` is cast directly to `Action` and dispatched without any runtime shape validation. The guard at L67–L72 only checks that the key exists and is truthy — it does not verify that the value has the `type` string or `payload` object that `Action` requires. A BFF returning `{ action: 42 }` or `{ action: { typo: 'navigate' } }` will pass the guard and be dispatched as a malformed action, likely causing a silent no-op or a runtime crash deep in the engine rather than a clear error at the boundary.

Fix: validate the shape before dispatching. A minimal Zod schema is sufficient:
```ts
const ActionSchema = z.object({ type: z.string(), payload: z.record(z.unknown()).optional() });
const parsed = ActionSchema.safeParse((body as Record<string, unknown>).action);
if (parsed.success) {
  await engine.dispatch(parsed.data as Action);
}
```
This matches the `zod-boundary-validation` pattern already used elsewhere in the TypeScript codebase and closes the gap without adding meaningful overhead.

_Cited:_ `patterns/universal/security/input-validation-at-boundary.md — external API responses are external input; downstream code should receive validated typed objects, not raw casts`

zenith-oi (approved):
## Zenith Code Review

**Small, well-scoped patch that re-enables server-driven action chaining in the SDUI runtime. The production change is correct and defensive. One pre-existing FLAG from the previous round remains unaddressed: `body.action` is cast directly to `Action` without any runtime shape validation, meaning a malformed BFF response can dispatch an untyped object into the engine.**

### What this PR is doing
**Intent:** Add response-body parsing to the `bff_call` action handler so that a `body.action` returned by the BFF is dispatched before the caller-declared `on_success` chain, enabling server-driven action chaining for flows like `append_items` pagination and post-submit navigation.
**Delivered:** The diff fully delivers the stated intent. The three-line production change (L63–L74 in bff-call.ts) implements defensive JSON parsing, dispatches `body.action` when present, then falls through to the existing `on_success` path. Six unit tests cover the happy path, ordering guarantee, error path, empty-body tolerance, and the concrete `append_items`+`navigate` ordering case. No linked task page or brief was present, so acceptance items cannot be checked by number.

**Positive:**
- Defensive `.catch(() => null)` on `res.json()` correctly tolerates non-JSON and empty bodies without throwing — exactly the right approach for a runtime that must not crash on unexpected BFF responses.
- The dispatch ordering (body.action first, then on_success) is clearly documented in both the changeset and inline comments, making the invariant auditable without running the code.
- Six unit tests cover all meaningful branches including the empty-body edge case and the concrete ordering pin for `append_items`+`navigate` — this is the right level of test coverage for a small but behaviorally significant change.
- Using `node:test` and `node:assert/strict` with a hand-rolled fetch stub keeps the test file dependency-free and fast, consistent with the repo's lightweight test philosophy.

**Read against:**
- `PR body — stated intent + test plan`
- `patterns/universal/security/input-validation-at-boundary.md — body.action cast to Action without runtime shape validation`
- `patterns/universal/security/zod-boundary-validation.md — TypeScript boundary validation discipline`

---

packages/sdui-runtime/src/action-engine/handlers/bff-call.ts
```diff
@@ -60,7 +60,21 @@ export async function handleBffCall(
       throw new Error(`BFF ${payload.method} ${endpointPath} returned ${res.status}`);
     }
 
+    // Server may bundle a follow-up action chain in the response body
+    // (e.g. append_items for pagination, navigate after submit).
+    // Parse defensively — non-JSON / empty bodies are tolerated.
+    const body = await res.json().catch(() => null);
```
zenith-oi, line 66: **FLAG** — [structured-error-types / silent-catch] The `.catch(() => null)` on `res.json()` silently swallows parse failures with no logging. Per Amplify coding standards, 'every catch block must log with context (IDs, operation name, relevant state). Silent catches are forbidden.' In practice this means a malformed JSON body from the BFF (a server bug) will be silently ignored and the runtime will proceed as if no body action was present — making the failure invisible in production logs. Fix: log at debug/warn level with the endpoint path before returning null, e.g. `.catch((err) => { logger.warn('bff_call: failed to parse response body', { endpoint: payload.endpoint, err }); return null; })`.

zenith-oi (approved):
## Zenith Code Review

**Small, focused patch that adds server-driven action chaining to the BFF call handler. The implementation is correct, the defensive `.catch(() => null)` is appropriate, and the test coverage is thorough. One issue worth flagging: the body is consumed via `res.json()` after the non-2xx error throw, but the response body has already been read by the time `res.json()` is called — this is fine because the throw exits before the `.json()` call on error paths. However, there is a silent-catch concern worth noting.**

### What this PR is doing
**Intent:** Add server-driven action chaining to the `bff_call` handler by parsing the response JSON body and dispatching `body.action` (if present) before any declared `on_success` chain, with defensive handling for non-JSON/empty responses.
**Delivered:** The diff fully delivers the stated intent. The implementation in `bff-call.ts` parses the response body defensively, dispatches `body.action` when present, then proceeds to the existing `on_success` dispatch. Six unit tests cover: body.action dispatch, ordering (body.action before on_success), error path (on_error on 5xx), no-op when neither present, non-JSON body tolerance, and the append_items+navigate ordering pin. The changeset entry is present. No acceptance items from a task page to check against (no linked task page detected).

**Positive:**
- Defensive `.catch(() => null)` correctly tolerates non-JSON and empty bodies without crashing the runtime — the right call for a client-side action engine that must not break on unexpected server responses.
- Dispatch ordering is explicit and well-documented in both the code comment and the changeset: body.action (server-driven) fires before on_success (caller-declared). This is the correct precedence for SDUI patterns.
- Test coverage is genuinely thorough: happy path, ordering invariant, error path, no-op path, non-JSON tolerance, and a concrete use-case pin (append_items + navigate). The baseline-failure note in the PR body ('without the fix, 3 of the new tests fail') is good engineering hygiene.
- The `optimistic` guard on `on_success` dispatch is preserved correctly — the new body.action dispatch is inserted before that guard, which is the right place.

---

PR #146, opened 2026-05-26
fix(sdui-runtime): bottom sheet imperative present + sheet registry pattern

mridulgupta-oi:
## Summary

Three connected fixes that make SDUI bottom sheets actually appear:

- **`useBottomSheetStore`** — split state into `registry` (sheet definitions by id), `openSheets` (which are presented), and `contexts` (per-sheet runtime payload). New `register(id, sheet)` adds without opening; `open(id, ctx?)` looks up the registered sheet and flips its open flag; `close(id?)` / `closeAll()` flip flags but preserve the registry so a sheet can be reopened.
- **`BottomSheetHost`** — now renders a `BottomSheetHostSheet` child per registered sheet. Each child owns its own gorhom modal ref and calls `ref.current?.present()` / `dismiss()` in a `useEffect` keyed on its open state. The previous declarative host (rendering `<BottomSheetModal>` with no ref) never invoked the imperative `present()` API so sheets were never visible.
- **Page renderers + `sheet` action** — `PageStandard`, `PageFeed`, `PageStickyFooter`, and the `BottomSheet` snippet now `register()` sheets on mount (no more open-then-immediately-close hack). The `sheet` action handler resolves the registered sheet by id and calls `open()`.

Adds unit tests for `register` / `open` / `close` / `closeAll` covering registration without opening, no-op on unregistered ids, context stamping, and the close-most-recent fallback.

## Why

`@gorhom/bottom-sheet`'s `BottomSheetModal` is an imperative component — there is no `visible` / `open` prop. You must hold a ref and call `.present()` to show it and `.dismiss()` to hide it. The existing host was written in the declarative style, so no SDUI bottom sheet ever rendered after the user tapped a trigger. This was first surfaced and patched in `amplify-creator-app/node_modules/@one-impression/sdui-runtime/dist` during Brief 264 Phase 4 validation (patches 8–10 in the rebuild log). This PR upstreams that fix into the source package.

## Test plan

- [x] `npm run build -w @one-impression/sdui-runtime` — clean tsup + tsc build
- [x] `npm test -w @one-impression/sdui-runtime` — 33 new + existing tests pass. (3 pre-existing failures unrelated to this PR — they fail on `origin/main` because `@one-impression/sdk-native-sdui` is not installed in CI; not introduced here.)
- [ ] Manual: tap an SDUI button with a `sheet` action in creator-app dev build; verify the sheet presents and dismisses, and that a second tap on a different sheet replaces correctly.

### DIFF
--- a/packages/sdui-runtime/src/bottom-sheet/BottomSheetHost.tsx
+++ b/packages/sdui-runtime/src/bottom-sheet/BottomSheetHost.tsx
@@ -1,9 +1,12 @@
-import React, { useCallback } from "react";
-import { BottomSheetModal, BottomSheetScrollView } from "@gorhom/bottom-sheet";
-import { useBottomSheetStore } from "./useBottomSheetStore.js";
+import React, { useCallback, useEffect, useRef } from "react";
+import {
+  BottomSheetModal,
+  BottomSheetScrollView,
+} from "@gorhom/bottom-sheet";
+import type { BottomSheetModalMethods } from "@gorhom/bottom-sheet/lib/typescript/types";
+import { useBottomSheetStore, type SheetEntry } from "./useBottomSheetStore.js";
 import { BottomSheetContext } from "./BottomSheetContext.js";
 import { Interpreter } from "../interpreter/Interpreter.js";
-import type { Node } from "@one-impression/sdk-native-sdui";
 
 const SIZE_TO_SNAP: Record<string, string[]> = {
   small: ["25%"],
@@ -12,13 +15,64 @@ const SIZE_TO_SNAP: Record<string, string[]> = {
   full: ["95%"],
 };
 
+interface BottomSheetHostSheetProps {
+  sheet: SheetEntry;
+  open: boolean;
+  onDismiss: () => void;
+}
+
 /**
- * Singleton host component — mount once at app root in _layout.tsx.
- * Subscribes to the Zustand bottom-sheet store and renders BottomSheetModals
- * for each entry in the stack.
+ * Per-sheet renderer that owns its own gorhom modal ref.
+ *
+ * gorhom's `BottomSheetModal` is an imperative API — it does NOT take a
+ * `visible` / `open` prop. To present a sheet you must call `ref.present()`,
+ * and to hide it you call `ref.dismiss()`. This child component bridges the
+ * declarative `open` flag from `useBottomSheetStore` to those imperative
+ * calls via a `useEffect` keyed on `open`.
+ */
+function BottomSheetHostSheet({
+  sheet,
+  open,
+  onDismiss,
+}: BottomSheetHostSheetProps): React.ReactElement {
+  const ref = useRef<BottomSheetModalMethods>(null);
+
+  useEffect(() => {
+    if (open) {
+      ref.current?.present();
+    } else {
+      ref.current?.dismiss();
+    }
+  }, [open]);
+
+  return (
+    <BottomSheetModal
+      ref={ref}
+      snapPoints={SIZE_TO_SNAP[sheet.size] ?? SIZE_TO_SNAP["medium"]}
+      enableDynamicSizing={sheet.size === "dynamic"}
+      onDismiss={onDismiss}
+    >
+      <BottomSheetScrollView>
+        <BottomSheetContext.Provider value={{ insideSheet: true }}>
+          {sheet.items.map((node, i) => (
+            <Interpreter key={node.id ?? i} node={node} />
+          ))}
+        </BottomSheetContext.Provider>
+      </BottomSheetScrollView>
+    </BottomSheetModal>
+  );
+}
+
+/**
+ * Singleton host component — mount once at app root in `_layout.tsx`.
+ *
+ * Subscribes to the Zustand bottom-sheet store and renders a child
+ * `BottomSheetHostSheet` for every sheet currently in the registry. Each
+ * child holds its own gorhom modal ref and reacts to its open state.
  */
 export function BottomSheetHost(): React.ReactElement {
-  const stack = useBottomSheetStore((s) => s.stack);
+  const registry = useBottomSheetStore((s) => s.registry);
+  const openSheets = useBottomSheetStore((s) => s.openSheets);
   const close = useBottomSheetStore((s) => s.close);
 
   const handleDismiss = useCallback(
@@ -28,23 +82,17 @@ export function BottomSheetHost(): React.ReactElement {
     [close],
   );
 
+  const entries = Object.values(registry);
+
   return (
     <>
-      {stack.map((sheet) => (
-        <BottomSheetModal
+      {entries.map((sheet) => (
+        <BottomSheetHostSheet
           key={sheet.id}
-          snapPoints={SIZE_TO_SNAP[sheet.size] ?? SIZE_TO_SNAP["medium"]}
-          enableDynamicSizing={sheet.size === "dynamic"}
+          sheet={sheet}
+          open={openSheets[sheet.id] === true}
           onDismiss={() => handleDismiss(sheet.id)}
-        >
-          <BottomSheetScrollView>
-            <BottomSheetContext.Provider value={{ insideSheet: true }}>
-              {(sheet.items as Node[]).map((node, i) => (
-                <Interpreter key={node.id ?? i} node={node} />
-              ))}
-            </BottomSheetContext.Provider>
-          </BottomSheetScrollView>
-        </BottomSheetModal>
+        />
       ))}
     </>
   );
--- a/packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
+++ b/packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
@@ -1,49 +1,92 @@
 import { create } from "zustand";
+import type { Node } from "@one-impression/sdk-native-sdui";
 
-const MAX_STACK_DEPTH = 2;
-
+/**
+ * Inline bottom-sheet record stored in the registry.
+ *
+ * Mirrors the `InlineBottomSheetSchema` payload from
+ * `@one-impression/sdk-native-sdui` (id, size, items, optional title +
+ * lifecycle triggers). The registry holds these by `sheet_id` so the
+ * `sheet` action handler can open them by id without re-defining the sheet.
+ */
 export interface SheetEntry {
   id: string;
   title?: string;
   size: string;
-  items: unknown[];
+  items: Node[];
   on_dismiss?: unknown;
   on_open?: unknown;
 }
 
 interface BottomSheetState {
-  stack: SheetEntry[];
-  open: (sheet: SheetEntry) => void;
-  close: (id?: string) => void;
+  /** All sheets declared by the active page, keyed by sheet_id. */
+  registry: Record<string, SheetEntry>;
+  /** Which registered sheets are currently presented. */
+  openSheets: Record<string, boolean>;
+  /** Per-sheet runtime context payload, stamped at open() time. */
+  contexts: Record<string, Record<string, unknown> | undefined>;
+
+  /**
+   * Register a sheet definition without opening it. Called by page renderers
+   * on mount for each `page.bottom_sheets[]` entry. Idempotent — overwrites
+   * any existing entry with the same id.
+   */
+  register: (sheetId: string, sheet: SheetEntry) => void;
+
+  /**
+   * Open a previously-registered sheet. Looks up the sheet by id from the
+   * registry; if no sheet was registered the call is a no-op (and logs a
+   * warning). An optional `contextPayload` is stamped against the sheet id
+   * so renderers / handlers inside the sheet can access it.
+   */
+  open: (sheetId: string, contextPayload?: Record<string, unknown>) => void;
+
+  /** Close a sheet by id; if omitted, closes the most-recently opened sheet. */
+  close: (sheetId?: string) => void;
+
+  /** Close every currently-open sheet (registry is preserved). */
   closeAll: () => void;
-  replace: (sheet: SheetEntry) => void;
 }
 
-export const useBottomSheetStore = create<BottomSheetState>((set) => ({
-  stack: [],
+export const useBottomSheetStore = create<BottomSheetState>((set, get) => ({
+  registry: {},
+  openSheets: {},
+  contexts: {},
 
-  open: (sheet) =>
+  register: (sheetId, sheet) =>
+    set((state) => ({
+      registry: { ...state.registry, [sheetId]: sheet },
+    })),
+
+  open: (sheetId, contextPayload) => {
+    const sheet = get().registry[sheetId];
+    if (!sheet) {
+      console.warn(
+        `[BottomSheetStore] open("${sheetId}") — no sheet registered with that id`,
+      );
+      return;
+    }
+    set((state) => ({
+      openSheets: { ...state.openSheets, [sheetId]: true },
+      contexts: { ...state.contexts, [sheetId]: contextPayload },
+    }));
+  },
+
+  close: (sheetId) =>
     set((state) => {
-      if (state.stack.length >= MAX_STACK_DEPTH) {
-        console.warn(
-          `[BottomSheetStore] Stack depth limit (${MAX_STACK_DEPTH}) reached — refusing to push "${sheet.id}"`,
-        );
-        return state;
+      if (sheetId) {
+        const { [sheetId]: _, ...remaining } = state.openSheets;
+        const { [sheetId]: __, ...remainingCtx } = state.contexts;
+        return { openSheets: remaining, contexts: remainingCtx };
       }
-      return { stack: [...state.stack, sheet] };
+      // No id: close the most-recently opened sheet.
+      const openIds = Object.keys(state.openSheets);
+      if (openIds.length === 0) return state;
+      const last = openIds[openIds.length - 1] as string;
+      const { [last]: _, ...remaining } = state.openSheets;
+      const { [last]: __, ...remainingCtx } = state.contexts;
+      return { openSheets: remaining, contexts: remainingCtx };
     }),
 
-  close: (id) =>
-    set((state) => ({
-      stack: id
-        ? state.stack.filter((s) => s.id !== id)
-        : state.stack.slice(0, -1),
-    })),
-
-  closeAll: () => set({ stack: [] }),
-
-  replace: (sheet) =>
-    set((state) => ({
-      stack: [...state.stack.slice(0, -1), sheet],
-    })),
+  closeAll: () => set({ openSheets: {}, contexts: {} }),
 }));
--- a/packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
+++ b/packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
@@ -43,6 +43,15 @@ interface BottomSheetState {
   register: (sheetId: string, sheet: SheetEntry) => void;
 
   /**
+   * Remove a registered sheet entirely — its definition, open flag,
+   * open-order entry, and stamped context all go. Called by renderers
+   * in their `useEffect` cleanup when a sheet (or its hosting page)
+   * unmounts, so navigating away does not leave orphan sheets in the
+   * registry that the host would keep rendering.
+   */
+  unregister: (sheetId: string) => void;
+
+  /**
    * Open a previously-registered sheet. Looks up the sheet by id from the
    * registry; if no sheet was registered the call is a no-op (and logs a
    * warning). An optional `contextPayload` is stamped against the sheet id
@@ -67,6 +76,19 @@ export const useBottomSheetStore = create<BottomSheetState>((set, get) => ({
     set((state) => ({
       registry: { ...state.registry, [sheetId]: sheet },
     })),
+
+  unregister: (sheetId) =>
+    set((state) => {
+      const { [sheetId]: _r, ...remainingRegistry } = state.registry;
+      const { [sheetId]: _o, ...remainingOpen } = state.openSheets;
+      const { [sheetId]: _c, ...remainingCtx } = state.contexts;
+      return {
+        registry: remainingRegistry,
+        openSheets: remainingOpen,
+        openOrder: state.openOrder.filter((id) => id !== sheetId),
+        contexts: remainingCtx,
+      };
+    }),
 
   open: (sheetId, contextPayload) => {
     const sheet = get().registry[sheetId];
--- a/packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
+++ b/packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
@@ -23,6 +23,15 @@ interface BottomSheetState {
   registry: Record<string, SheetEntry>;
   /** Which registered sheets are currently presented. */
   openSheets: Record<string, boolean>;
+  /**
+   * Explicit open-order history (oldest → newest). The last entry is the
+   * topmost open sheet. Tracked separately from `openSheets` so the
+   * "most recently opened" answer does not rely on JS object key
+   * insertion order (which the ECMAScript spec does not guarantee for
+   * arbitrary string keys and which breaks under serialization /
+   * Zustand persist middleware).
+   */
+  openOrder: string[];
   /** Per-sheet runtime context payload, stamped at open() time. */
   contexts: Record<string, Record<string, unknown> | undefined>;
 
@@ -51,6 +60,7 @@ interface BottomSheetState {
 export const useBottomSheetStore = create<BottomSheetState>((set, get) => ({
   registry: {},
   openSheets: {},
+  openOrder: [],
   contexts: {},
 
   register: (sheetId, sheet) =>
@@ -68,6 +78,10 @@ export const useBottomSheetStore = create<BottomSheetState>((set, get) => ({
     }
     set((state) => ({
       openSheets: { ...state.openSheets, [sheetId]: true },
+      // Move id to the end of the open-order history. Filtering an
+      // existing entry ensures reopening an already-open sheet promotes
+      // it to topmost rather than producing duplicates.
+      openOrder: [...state.openOrder.filter((id) => id !== sheetId), sheetId],
       contexts: { ...state.contexts, [sheetId]: contextPayload },
     }));
   },
@@ -77,16 +91,23 @@ export const useBottomSheetStore = create<BottomSheetState>((set, get) => ({
       if (sheetId) {
         const { [sheetId]: _, ...remaining } = state.openSheets;
         const { [sheetId]: __, ...remainingCtx } = state.contexts;
-        return { openSheets: remaining, contexts: remainingCtx };
+        return {
+          openSheets: remaining,
+          openOrder: state.openOrder.filter((id) => id !== sheetId),
+          contexts: remainingCtx,
+        };
       }
-      // No id: close the most-recently opened sheet.
-      const openIds = Object.keys(state.openSheets);
-      if (openIds.length === 0) return state;
-      const last = openIds[openIds.length - 1] as string;
+      // No id: close the most-recently opened sheet (last of openOrder).
+      if (state.openOrder.length === 0) return state;
+      const last = state.openOrder[state.openOrder.length - 1] as string;
       const { [last]: _, ...remaining } = state.openSheets;
       const { [last]: __, ...remainingCtx } = state.contexts;
-      return { openSheets: remaining, contexts: remainingCtx };
+      return {
+        openSheets: remaining,
+        openOrder: state.openOrder.slice(0, -1),
+        contexts: remainingCtx,
+      };
     }),
 
-  closeAll: () => set({ openSheets: {}, contexts: {} }),
+  closeAll: () => set({ openSheets: {}, openOrder: [], contexts: {} }),
 }));
--- a/packages/sdui-runtime/src/hooks/useBottomSheetData.ts
+++ b/packages/sdui-runtime/src/hooks/useBottomSheetData.ts
@@ -1,14 +1,23 @@
+import { useShallow } from "zustand/react/shallow";
 import { useBottomSheetStore } from "../bottom-sheet/useBottomSheetStore.js";
 
 /**
  * Returns the currently active (most-recently opened) bottom sheet entry,
  * or null if none open. Resolves the topmost open sheet against the store
  * registry — consumers read sheet data for conditional rendering.
+ *
+ * Uses `useShallow` so the registry + openOrder read is taken as a single
+ * atomic snapshot. Under React 18 concurrent rendering, two separate
+ * `useBottomSheetStore(selector)` calls could otherwise see torn state
+ * (one selector observes the new value while the other still sees the
+ * old) when a single store update mutates both fields.
  */
 export function useBottomSheetData() {
-  const registry = useBottomSheetStore((s) => s.registry);
-  const openOrder = useBottomSheetStore((s) => s.openOrder);
-  if (openOrder.length === 0) return null;
-  const topId = openOrder[openOrder.length - 1] as string;
-  return registry[topId] ?? null;
+  return useBottomSheetStore(
+    useShallow((s) => {
+      if (s.openOrder.length === 0) return null;
+      const topId = s.openOrder[s.openOrder.length - 1] as string;
+      return s.registry[topId] ?? null;
+    }),
+  );
 }
--- a/packages/sdui-runtime/src/hooks/useBottomSheetData.ts
+++ b/packages/sdui-runtime/src/hooks/useBottomSheetData.ts
@@ -1,10 +1,15 @@
 import { useBottomSheetStore } from "../bottom-sheet/useBottomSheetStore.js";
 
 /**
- * Returns the currently active (topmost) bottom sheet entry, or null if none open.
- * Ported 1:1 from legacy — consumers read sheet data for conditional rendering.
+ * Returns the currently active (most-recently opened) bottom sheet entry,
+ * or null if none open. Resolves the topmost open sheet against the store
+ * registry — consumers read sheet data for conditional rendering.
  */
 export function useBottomSheetData() {
-  const stack = useBottomSheetStore((s) => s.stack);
-  return stack.length > 0 ? stack[stack.length - 1] : null;
+  const registry = useBottomSheetStore((s) => s.registry);
+  const openSheets = useBottomSheetStore((s) => s.openSheets);
+  const openIds = Object.keys(openSheets).filter((id) => openSheets[id]);
+  if (openIds.length === 0) return null;
+  const topId = openIds[openIds.length - 1] as string;
+  return registry[topId] ?? null;
 }
--- a/packages/sdui-runtime/src/hooks/useBottomSheetData.ts
+++ b/packages/sdui-runtime/src/hooks/useBottomSheetData.ts
@@ -7,9 +7,8 @@ import { useBottomSheetStore } from "../bottom-sheet/useBottomSheetStore.js";
  */
 export function useBottomSheetData() {
   const registry = useBottomSheetStore((s) => s.registry);
-  const openSheets = useBottomSheetStore((s) => s.openSheets);
-  const openIds = Object.keys(openSheets).filter((id) => openSheets[id]);
-  if (openIds.length === 0) return null;
-  const topId = openIds[openIds.length - 1] as string;
+  const openOrder = useBottomSheetStore((s) => s.openOrder);
+  if (openOrder.length === 0) return null;
+  const topId = openOrder[openOrder.length - 1] as string;
   return registry[topId] ?? null;
 }
--- a/packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
+++ b/packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
@@ -53,6 +53,7 @@ const styles = StyleSheet.create({
 export function PageFeedRenderer({ page }: PageProps): React.ReactElement {
   const actionEngine = useActionEngine();
   const register = useBottomSheetStore((s) => s.register);
+  const unregister = useBottomSheetStore((s) => s.unregister);
   const { refreshing, onRefresh } = usePageRefresh(page.on_refresh);
   const [loadingMore, setLoadingMore] = useState(false);
   const loadingMoreRef = useRef(false);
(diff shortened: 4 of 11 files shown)


packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
```diff
@@ -1,49 +1,92 @@
...
+
+  close: (sheetId) =>
+    set((state) => {
+      if (sheetId) {
+        const { [sheetId]: _, ...remaining } = state.openSheets;
+        const { [sheetId]: __, ...remainingCtx } = state.contexts;
+        return { openSheets: remaining, contexts: remainingCtx };
+      }
+      // No id: close the most-recently opened sheet.
+      const openIds = Object.keys(state.openSheets);
```
zenith-oi: **FLAG** — [spec-undefined-insertion-order-for-close-fallback] The no-id `close()` path (and the `useBottomSheetData` hook) determines the 'most recently opened' sheet by taking `Object.keys(openSheets)[length - 1]`. ECMAScript guarantees insertion order for string keys in modern engines (V8 honours it), but the spec only guarantees it for non-integer-like string keys — and more importantly, when a sheet is closed and then re-opened, its key moves to the end of the object only if it was deleted and re-inserted. The current `open()` implementation does `{ ...state.openSheets, [sheetId]: true }` which, if the key already exists, preserves its original position in the spread rather than moving it to the end. So if sheet 'a' is opened, then sheet 'b' is opened, then 'a' is closed and re-opened, `Object.keys` will return `['a', 'b']` — the 'most recent' close would incorrectly target 'b'. Fix: maintain an explicit `openOrder: string[]` array in state, push on open and splice on close, so ordering is unambiguous and spec-guaranteed.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-motivated fix that correctly upstreams the imperative-API bridge from the patched dist. The store redesign is sound, the test coverage is thorough, and the page-renderer changes are consistent. One minor concern worth flagging: the `close()` fallback (no-id path) relies on `Object.keys(openSheets)` insertion order to determine 'most recently opened', which is reliable in V8 for string keys but is not guaranteed by the ECMAScript spec — a small ordering structure would make this robust.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet store from a stack-push model to a registry+openSheets model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never visible because the old host never called `present()`.
**Delivered:** Fully delivered. The store is cleanly split into `registry` / `openSheets` / `contexts`; `BottomSheetHostSheet` correctly bridges the declarative open flag to the imperative gorhom API via `useEffect`; all three page renderers and the snippet registrar are updated to call `register()` instead of the open-then-close hack; the `sheet` action handler now calls `open(id)` with the new signature; and a comprehensive unit-test suite covers all store operations including edge cases.

**Positive:**
- The `BottomSheetHostSheet` component correctly isolates each sheet's gorhom ref and bridges the declarative `open` flag to `present()` / `dismiss()` — this is exactly the right pattern for imperative third-party APIs.
- Removing the open-then-immediately-close hack from all three page renderers is a clean, correct fix; the new `register()` API makes intent explicit.
- Unit tests are thorough: they cover registration without opening, idempotent re-registration, context stamping, the no-op on unregistered ids (including the console.warn assertion), close-by-id, close-most-recent, close-all, and the full round-trip reopen — good edge-case discipline.
- The dynamic import in `sheet.ts` is preserved, keeping the bottom-sheet store out of the action-engine's static dependency graph.

---

packages/sdui-runtime/src/bottom-sheet/useBottomSheetStore.ts
zenith-oi: **FLAG** — [implicit-insertion-order-dependency] The `close()` no-id path identifies the most-recently opened sheet via `Object.keys(state.openSheets)[length - 1]`, relying on JS object property insertion order. V8 preserves insertion order for string keys in practice, but the ECMAScript spec only guarantees this for integer-indexed keys (ES2015+). If `openSheets` is ever reconstructed from serialized/deserialized state (e.g. Zustand persist middleware added later, or a test that calls `setState` with a pre-built object), insertion order is not preserved and the 'topmost' sheet will be wrong silently. The same logic is duplicated in `useBottomSheetData.ts` (L11-13). Fix: track open order explicitly with a parallel `openOrder: string[]` array in the store state — push on `open()`, pop/filter on `close()`. This makes the intent explicit and order-stable regardless of serialization.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-motivated fix that upstreams a validated patch from the creator-app dist into the source package. The store redesign is sound, the gorhom imperative bridge is correctly implemented, and the test coverage is thorough. One real concern: the `close()` fallback (no-id path) relies on `Object.keys(openSheets)` insertion order to identify the 'most recently opened' sheet, which is reliable in V8 for string keys but is not a language guarantee and could silently break in non-V8 environments or if the store state is ever reconstructed from serialized JSON.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet store from a stack-push model to a registry+openSheets model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never visually appearing.
**Delivered:** The diff fully delivers the stated intent. The store is correctly split into `registry` / `openSheets` / `contexts`. `BottomSheetHost` now renders a stable child per registered sheet and bridges the declarative open flag to the imperative gorhom API via `useEffect`. All three page renderers and the `BottomSheet` snippet are updated to call `register()` instead of the open-then-close hack. The `sheet` action handler now calls `open(id)` with the new signature. Unit tests cover all store operations including edge cases. No acceptance items from a linked task page to check against — intent is inferred from diff + body.

**Positive:**
- The gorhom imperative bridge via `useEffect` keyed on `open` is the correct pattern — the comment in `BottomSheetHostSheet` explaining WHY the imperative API requires this is genuinely useful and will save the next engineer from re-discovering it.
- Removing the open-then-immediately-close hack from all three page renderers is a meaningful correctness improvement — the old pattern was racy and semantically wrong.
- Unit tests are comprehensive: they cover registration without opening, idempotent re-registration, context stamping, no-op on unregistered id with warn assertion, close-by-id, close-most-recent, close-no-op when empty, closeAll, and full round-trip. The `console.warn` stub pattern is clean.
- The `eslint-disable-next-line react-hooks/exhaustive-deps` comments are correctly placed and the omission of `bottomSheetStore` from deps is justified — Zustand store references are stable across renders.

---

zenith-oi (dismissed):
## Zenith Code Review

**This PR correctly upstreams the gorhom imperative-API fix from the patched dist into the source package. The store redesign is clean, the bridge component pattern is the right approach for gorhom's ref-based API, and the test coverage is solid. The previously-flagged insertion-order issue is resolved by `openOrder`. No blocking issues remain.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet store from a push/pop stack model to a registry+openSheets model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never visible because the old host never called `present()` (inferred from diff + body).
**Delivered:** Fully delivered. The store is correctly split into `registry` / `openSheets` / `openOrder` / `contexts`. `BottomSheetHost` now renders a stable child per registered sheet and bridges the declarative open flag to the imperative gorhom API via `useEffect`. All three page renderers and the `BottomSheet` snippet are updated to call `register()` instead of the open-then-close hack. The `sheet` action handler now calls `open(id)` instead of `open({ id })`. 186-line unit test suite covers the full store contract. The one previously-flagged issue (spec-undefined-insertion-order-for-close-fallback) was addressed by introducing `openOrder: string[]` — the close-without-id path and `useBottomSheetData` now use `openOrder` rather than `Object.keys()` ordering.

No issues found.

**Positive:**
- `openOrder: string[]` cleanly resolves the spec-undefined insertion-order concern from the previous review round — the close-without-id path and `useBottomSheetData` are now deterministic regardless of JS engine or serialization.
- Splitting `BottomSheetHostSheet` into its own component so each sheet owns its own gorhom ref is exactly the right pattern for bridging a declarative store to an imperative third-party API.
- The `open()` guard that warns and no-ops on unregistered sheet IDs (with a test asserting the warning fires) is good defensive programming for SDUI action dispatch.
- Test suite covers all meaningful state transitions — register without open, context stamping, close-by-id vs close-topmost, closeAll, round-trip reopen — without any mocking of the store itself.

---

packages/sdui-runtime/src/pages/PageStandard/PageStandard.renderer.tsx
```diff
@@ -23,25 +23,24 @@ export function PageStandardRenderer({ page }: PageProps): React.ReactElement {
...
           items: sheet.items ?? [],
           on_dismiss: sheet.on_dismiss,
           on_open: sheet.on_open,
         });
-        // Immediately close — we only need them registered, not visible.
-        bottomSheetStore.close(sheet.id);
       }
     }
-  }, [page.bottom_sheets, bottomSheetStore]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
zenith-oi: **FLAG** — [react-hooks/exhaustive-deps suppression hides stale-closure risk] The `// eslint-disable-next-line react-hooks/exhaustive-deps` comment suppresses a lint warning about `bottomSheetStore` being omitted from the dependency array. The store reference from `useBottomSheetStore()` is a stable Zustand store object (same reference across renders), so omitting it is safe in practice. However, the suppression comment is the only thing communicating that intent — a future reader has no way to know whether the omission was deliberate or accidental. The same pattern appears in PageFeed.renderer.tsx (L81) and PageStickyFooter.renderer.tsx (L71). The idiomatic fix is to extract the `register` action directly from the store with a selector (`const register = useBottomSheetStore((s) => s.register)`) and include it in the dep array — Zustand action references are stable by construction, so the effect still runs only when `page.bottom_sheets` changes, and the lint rule is satisfied without suppression. This also applies to the BottomSheet.renderer.tsx snippet (L70), which already had the suppression but now calls `register` instead of `open`.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-motivated fix that correctly bridges gorhom's imperative BottomSheetModal API to a declarative Zustand store. The implementation is sound and the test coverage is thorough. One minor concern worth flagging: the eslint-disable-next-line comments suppressing react-hooks/exhaustive-deps in the page renderers are hiding a real stale-closure risk.**

### What this PR is doing
**Intent:** Upstream the gorhom bottom-sheet imperative-API fix from the creator-app hotpatch into the sdui-runtime source package by splitting the store into registry/openSheets/contexts, bridging the declarative open flag to gorhom's ref.present()/dismiss() calls, and adding unit tests for the store.
**Delivered:** The diff fully delivers the stated intent. The store refactor (registry + openSheets + openOrder + contexts), the BottomSheetHostSheet bridge component with useEffect → ref.present()/dismiss(), the page renderer and action handler updates to use register() instead of open-then-close, and the 186-line test file covering all store operations are all present and correct. No acceptance items are missing.

**Positive:**
- The split of store state into registry/openSheets/openOrder/contexts is a clean design — separating 'what sheets exist' from 'which are open' makes the imperative-to-declarative bridge in BottomSheetHostSheet straightforward and easy to reason about.
- Extracting BottomSheetHostSheet as a child component that owns its own gorhom ref is exactly the right pattern for bridging an imperative API to a declarative parent — each sheet gets its own stable ref and its own useEffect keyed on the open flag.
- The openOrder array tracking insertion order separately from the openSheets object is a thoughtful correctness decision, and the comment explaining why (JS object key order is not spec-guaranteed under serialization) is exactly the kind of WHY documentation that belongs in code.
- Test coverage is comprehensive: register idempotency, open no-op on unregistered id with console.warn verification, close-most-recent fallback, closeAll preserving registry, openOrder promotion on reopen, and a full round-trip test. Using node:test with no test framework dependency is appropriate for a pure store module.

**Read against:**
- `PR body — Why section (gorhom imperative API, Brief 264 Phase 4 patches 8–10)`
- `PR body — Test plan`

---

packages/sdui-runtime/src/snippets/BottomSheet/BottomSheet.renderer.tsx
```diff
@@ -57,22 +57,19 @@ function BottomSheetRegistrar({
...
-      items,
+      items: items as Node[],
     });
-
-    return () => {
-      close(id);
-    };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [id, title, size, items, register]);
```
zenith-oi: **FLAG** — [missing-cleanup-on-unmount] The previous implementation returned a cleanup function from the `useEffect` that called `close(id)` on unmount. The new implementation removes that cleanup entirely (the `return () => { close(id); }` block is deleted). For page renderers (`PageStandard`, `PageFeed`, `PageStickyFooter`) this is intentional — the registry is page-scoped and the host is a singleton, so stale entries from a navigated-away page will persist in the registry until the next page registers over them. However, the `BottomSheetRegistrar` snippet is a component that can be conditionally rendered within a page (e.g. a sheet that only appears when a certain data state is true). When it unmounts mid-page, its sheet entry stays in the registry indefinitely. If the sheet id is reused on a different page with different content, the stale entry will be overwritten (idempotent), but if the id is unique, the registry grows unboundedly across navigation. More concretely: the `BottomSheetHost` renders a `BottomSheetHostSheet` for every entry in `registry` — a leaked entry means a hidden-but-mounted `BottomSheetModal` DOM node persists after the registrar unmounts. Consider adding an `unregister(sheetId)` action to the store and restoring the cleanup: `return () => { unregister(id); }` in `BottomSheetRegistrar`'s effect. Page renderers can omit cleanup since their sheets are intentionally page-lifetime.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-motivated fix that upstreams a previously hot-patched bug. The store redesign is sound, the imperative-bridge pattern is the correct approach for gorhom, and the test coverage is thorough. One real issue worth flagging: the `BottomSheetRegistrar` snippet removes its cleanup effect, meaning sheets registered by the snippet are never unregistered from the store when the snippet unmounts — this is a registry leak that can cause stale sheets to appear on subsequent page visits.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet store from a stack-push model to a registry+open-flags model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never actually visible.
**Delivered:** Fully delivered. The store is cleanly split into `registry` / `openSheets` / `openOrder` / `contexts`; page renderers now call `register()` instead of the open-then-close hack; `BottomSheetHostSheet` bridges the declarative flag to the imperative gorhom ref; the `sheet` action handler calls `open(id)` by lookup; and 186 lines of unit tests cover all store operations. No acceptance items from a linked task page are present, but the PR body's stated goals are all addressed in the diff.

**Positive:**
- The imperative-bridge pattern (`useEffect` keyed on `open` calling `ref.current?.present()` / `dismiss()`) is exactly the right way to drive gorhom's `BottomSheetModal` from declarative state — well-reasoned and clearly documented in the JSDoc.
- Splitting `openOrder` as a separate array (rather than relying on JS object key insertion order) is a thoughtful correctness decision, and the comment explaining why is genuinely useful.
- Test coverage is excellent: 13 tests covering registration, open/close semantics, context stamping, the no-op guard on unregistered ids, `closeAll`, and the full round-trip. The `resetStore()` helper keeps tests hermetic.

---

packages/sdui-runtime/src/snippets/BottomSheet/BottomSheet.renderer.tsx
zenith-oi: **FLAG** — [missing-cleanup-on-unmount] The previous `BottomSheetRegistrar` returned a cleanup function from `useEffect` that called `close(id)` on unmount. The new version drops the cleanup entirely — there is no `unregister` action on the store, and the `useEffect` has no return value. When a `BottomSheet` snippet unmounts (e.g. the page navigates away), its entry stays in `registry` and its gorhom `BottomSheetHostSheet` child keeps rendering in `BottomSheetHost`. This is a memory/render leak and can cause stale sheets to reappear if the same id is reused on a different page. The store should expose an `unregister(sheetId)` action that removes the entry from `registry` (and closes it if open), and `BottomSheetRegistrar` should call it in the `useEffect` cleanup. Page renderers (PageStandard, PageFeed, PageStickyFooter) have the same gap — they register on mount but never unregister on unmount.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-motivated fix that upstreams a previously hot-patched bug. The store redesign is sound, the gorhom bridge pattern is correct, and the test coverage is thorough. One real issue worth flagging: the `BottomSheetRegistrar` cleanup path was silently dropped, which can leave stale registry entries when a snippet unmounts.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet store from a stack-push model to a registry+open-flags model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never visible because the old host never called `present()` (inferred from diff + body).
**Delivered:** Fully delivered. The store is cleanly split into `registry` / `openSheets` / `openOrder` / `contexts`. `BottomSheetHost` now renders a stable child per registered sheet and bridges the declarative flag to the imperative gorhom API via `useEffect`. All three page renderers and the `BottomSheet` snippet drop the open-then-immediately-close hack in favour of `register()`. The `sheet` action handler calls `open(id)` instead of `open({...})`. 186-line unit-test suite covers every store operation including edge cases. No acceptance items to cite (no linked task page).

**Positive:**
- The `openOrder` array as an explicit ordering mechanism (rather than relying on JS object key insertion order) is a thoughtful design choice — the comment explaining why it exists is exactly the right level of documentation.
- Splitting `BottomSheetHostSheet` into its own component so each sheet owns its own gorhom ref is the correct pattern for bridging declarative state to an imperative API.
- The 186-line unit-test suite is comprehensive: it covers registration without opening, no-op on unregistered ids, context stamping, close-most-recent fallback, `openOrder` promotion on reopen, and the full round-trip. Using Node's built-in `node:test` runner with no extra dependencies is a clean choice for a pure-store test.
- Replacing `useBottomSheetStore()` (full store subscription) with `useBottomSheetStore((s) => s.register)` in all three page renderers is a correct performance improvement — components now only re-render when the `register` function reference changes, not on every store mutation.

---

zenith-oi (dismissed):
## Zenith Code Review

**This PR correctly upstreams the gorhom imperative-API fix from the creator-app patch into the sdui-runtime source package. The implementation is clean and the test coverage is thorough. One previously-raised FLAG about missing cleanup on unmount in `BottomSheet.renderer.tsx` is now resolved — the new `unregister(id)` cleanup is present in the effect's return function.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet store from a push/pop stack model to a registry+openSheets model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing the root cause of sheets never appearing after a `sheet` action.
**Delivered:** The diff fully delivers the stated intent. The store is restructured (registry, openSheets, openOrder, contexts), the host component now renders a stable `BottomSheetHostSheet` per registered sheet and bridges the declarative open flag to the imperative gorhom API via `useEffect`, page renderers call `register()` on mount and `unregister()` on cleanup, the `sheet` action handler calls `open(id)` instead of stamping a new sheet definition, and 216 lines of unit tests cover the store's public API. The previous open-then-immediately-close hack is gone.

No issues found.

**Positive:**
- The `openOrder` array as a separate structure from `openSheets` is a well-reasoned design choice — the JSDoc comment explaining why JS object key insertion order is not reliable under serialization/persist middleware is exactly the kind of 'why' documentation that saves future engineers from regressing this.
- Splitting `BottomSheetHostSheet` into its own component so each sheet owns its own gorhom ref is the correct structural fix — it avoids the ref-per-dynamic-child problem that would arise if refs were managed in the parent.
- The test suite is comprehensive: it covers registration without opening, no-op on unregistered ids, context stamping, close-most-recent fallback, openOrder promotion on reopen, and the full round-trip. The `resetStore()` helper keeps tests isolated without mocking.

---

packages/sdui-runtime/src/hooks/useBottomSheetData.ts
```diff
@@ -1,10 +1,14 @@
...
- * Returns the currently active (topmost) bottom sheet entry, or null if none open.
- * Ported 1:1 from legacy — consumers read sheet data for conditional rendering.
+ * Returns the currently active (most-recently opened) bottom sheet entry,
+ * or null if none open. Resolves the topmost open sheet against the store
+ * registry — consumers read sheet data for conditional rendering.
  */
 export function useBottomSheetData() {
-  const stack = useBottomSheetStore((s) => s.stack);
-  return stack.length > 0 ? stack[stack.length - 1] : null;
+  const registry = useBottomSheetStore((s) => s.registry);
```
zenith-oi: **FLAG** — [concurrent-write-protection / torn-selector-read] Two separate `useBottomSheetStore(selector)` calls in the same render function (`registry` on L9, `openOrder` on L10) are not atomic. Under React 18 concurrent rendering, a state update that changes both `registry` and `openOrder` simultaneously can be observed mid-flight: one selector sees the new value while the other still sees the old one. The result is a stale `topId` that resolves to `undefined` in the registry (or vice-versa), causing the hook to return `null` when a sheet is actually open.

Fix: use Zustand's `useShallow` or a single combined selector:
```ts
import { useShallow } from 'zustand/react/shallow';

export function useBottomSheetData() {
  return useBottomSheetStore(
    useShallow((s) => {
      if (s.openOrder.length === 0) return null;
      const topId = s.openOrder[s.openOrder.length - 1] as string;
      return s.registry[topId] ?? null;
    }),
  );
}
```
This is a single atomic snapshot so the registry and openOrder are always consistent with each other.

zenith-oi (dismissed):
## Zenith Code Review

**This PR correctly upstreams the gorhom imperative-API fix from the creator-app hotpatch into the sdui-runtime source package. The architecture is sound — registry/open-flags separation is the right model, the per-sheet ref pattern is the correct bridge, and the store tests are thorough. One genuine issue worth flagging: the `useBottomSheetData` hook reads two separate Zustand selectors in the same render, which can produce a torn read under concurrent React rendering.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet subsystem from a stack-push model to a registry+open-flags model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never visible because the old host never called `present()`.
**Delivered:** The diff fully delivers the stated intent. The store is correctly split into `registry` / `openSheets` / `openOrder` / `contexts`; `BottomSheetHostSheet` bridges the declarative open flag to the imperative gorhom API via `useEffect`; all three page renderers and the `BottomSheet` snippet now call `register()` on mount and `unregister()` on cleanup instead of the open-then-immediately-close hack; the `sheet` action handler calls `open(id)` against the registry; and 216 lines of unit tests cover the store's public surface. No acceptance items are missing.

**Positive:**
- The `BottomSheetHostSheet` component is exactly the right abstraction — each sheet owns its own gorhom ref and the `useEffect([open])` bridge is clean and idiomatic for imperative third-party APIs.
- Removing the open-then-immediately-close hack from all three page renderers and replacing it with a proper `register` / `unregister` lifecycle is a meaningful correctness improvement, not just a refactor.
- The `openOrder` array as a separate structure (rather than relying on JS object key insertion order) is a thoughtful design decision, and the JSDoc comment explaining why is exactly the right level of documentation.
- The unit test suite is comprehensive — it covers registration without opening, no-op on unregistered ids, context stamping, close-most-recent fallback, `closeAll` registry preservation, and full round-trip reopening. The `console.warn` stub pattern in the no-op test is clean.

**Read against:**
- `PR body — stated intent + why section (gorhom imperative API, Brief 264 Phase 4 validation context)`
- `PR body — test plan (33 tests pass, manual sheet-present/dismiss verification pending)`

---

packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
```diff
@@ -63,22 +64,26 @@ export function PageFeedRenderer({ page }: PageProps): React.ReactElement {
   const loader = pageData.loader;
   const emptyState = pageData.empty_state;
 
-  // Register inline bottom sheets
+  // Register inline bottom sheets (do not open) — see PageStandard for details.
   useEffect(() => {
```
zenith-oi, line 68: **FLAG** — [DRY / code-quality] The register/unregister `useEffect` block is copy-pasted verbatim across `PageFeed.renderer.tsx`, `PageStandard.renderer.tsx`, and `PageStickyFooter.renderer.tsx` (all three are identical: early-return guard, `for` loop calling `register`, cleanup loop calling `unregister`, same deps array). Extract this into a shared hook — e.g. `useRegisterBottomSheets(page.bottom_sheets)` — and call it from each renderer. This removes three identical blocks and means any future change to registration logic (e.g. handling `on_open` triggers) only needs to happen in one place. The hook is trivial: it takes `bottom_sheets: BottomSheet[] | undefined`, calls `register`/`unregister` from the store, and returns nothing.

zenith-oi (dismissed):
## Zenith Code Review

> ⚠️ **Round 5 — engineer attention requested.** The reviewer-engineer loop has reached round 5; this is the threshold where auto-iteration typically stops being productive (either the review is missing context the engineer has, or there's a genuine design call to make). Auto-fixes are still posted, but please review this thread end-to-end rather than relying on another round to converge.

**Clean, well-motivated fix that upstreams a previously hot-patched bug. The store redesign is sound, the imperative-bridge pattern is the correct approach for gorhom, and the test coverage is thorough. One DRY issue worth addressing: the register/unregister `useEffect` block is copy-pasted verbatim across three page renderers.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet store from a push-stack model to a registry+openSheets model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never visible because the old declarative host never called `present()`.
**Delivered:** Fully delivered. The store is correctly split into `registry` / `openSheets` / `openOrder` / `contexts`. `BottomSheetHostSheet` bridges the declarative `open` flag to the imperative gorhom API via `useEffect`. All three page renderers and the `BottomSheet` snippet now `register()` on mount and `unregister()` on cleanup instead of the open-then-immediately-close hack. The `sheet` action handler calls `open(id)` against the registry. 216 lines of unit tests cover every store operation including edge cases. No linked task page or brief to check acceptance items against — inferred from diff + body.

**Positive:**
- The `openOrder` array as an explicit open-order tracker (rather than relying on JS object key insertion order) is a thoughtful correctness decision — the comment explaining why is exactly the right level of documentation.
- Using `useShallow` in `useBottomSheetData` to take a single atomic snapshot of `registry + openOrder` is correct and the comment explaining the React 18 concurrent-rendering torn-state risk is genuinely useful.
- Test coverage is excellent: 216 lines covering register idempotency, open no-op on unregistered id, context stamping, close-most-recent fallback, closeAll registry preservation, unregister full cleanup, and the full round-trip. The `console.warn` stub in the no-op test is a nice touch.

---

packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
zenith-oi, line 68: **FLAG** — [DRY / code-quality] The register/unregister `useEffect` block (lines 68-86 in PageFeed, 32-50 in PageStandard, 59-77 in PageStickyFooter) is copy-pasted verbatim across all three page renderers. Any future change to registration logic (e.g. adding a new `SheetEntry` field) must be applied in three places. Extract a `useRegisterBottomSheets(sheets, register, unregister)` hook — or more simply a `usePageBottomSheets(page)` hook that reads `register`/`unregister` from the store internally — and call it from each renderer. The `BottomSheet.renderer.tsx` snippet already has its own slightly different variant (it also passes `apiEndpoint`), so the hook can accept the common subset. This is not blocking because the logic is correct and the duplication is currently only three sites, but it will drift.

zenith-oi (dismissed):
## Zenith Code Review

**This PR correctly upstreams a runtime fix for SDUI bottom sheets that were never presenting because gorhom's `BottomSheetModal` requires imperative `ref.present()` calls. The implementation is clean and well-tested. One DRY issue is worth flagging: the register/unregister `useEffect` block is copy-pasted identically across three page renderers.**

### What this PR is doing
**Intent:** Refactor the SDUI bottom-sheet system from a stack-push model to a registry+open-flag model so that gorhom's imperative `BottomSheetModal` API is correctly driven via `ref.present()` / `ref.dismiss()`, fixing sheets that were never actually visible.
**Delivered:** The diff fully delivers the stated intent. The store is split into `registry` / `openSheets` / `openOrder` / `contexts`; page renderers now call `register()` on mount and `unregister()` on cleanup instead of the open-then-immediately-close hack; `BottomSheetHost` renders a stable `BottomSheetHostSheet` child per registered sheet that bridges the declarative flag to the imperative gorhom API via `useEffect`; the `sheet` action handler calls `open(id)` against the registry; and 216 lines of unit tests cover the store's state machine. The PR body's description matches the diff precisely.

**Positive:**
- The `openOrder` array to track insertion order is the right call — explicitly documented why JS object key order can't be relied on for the 'most recently opened' query, especially under Zustand persist middleware.
- Using `useShallow` in `useBottomSheetData` to take a single atomic snapshot of `openOrder` + `registry` is correct and the comment explaining the React 18 concurrent-rendering torn-state risk is genuinely useful.
- The `BottomSheetHostSheet` decomposition is clean: each sheet owns its own gorhom ref and the `useEffect([open])` bridge is the minimal correct way to drive the imperative API declaratively.
- Test coverage is thorough — all state-machine transitions (register, open, close by id, close without id, closeAll, unregister, round-trip reopen) are exercised, including the no-op / warning path for unregistered sheet ids.
- The `unregister` cleanup in page renderer `useEffect` return functions correctly prevents orphan registry entries when navigating away.

---

packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
zenith-oi, line 68: **FLAG** — [DRY / code-quality] The register/unregister `useEffect` block is copy-pasted verbatim across `PageFeed.renderer.tsx`, `PageStandard.renderer.tsx`, and `PageStickyFooter.renderer.tsx`. This was flagged in the previous round and remains unaddressed. Extract a `useRegisterBottomSheets(sheets: BottomSheet[] | undefined)` hook that encapsulates the effect, cleanup, and deps array. All three renderers then call the hook in one line. The duplication means any future change to registration logic (e.g. adding error handling, changing the cleanup condition) must be applied in three places.

zenith-oi, line 68: **FLAG** — [DRY / code-quality] The register/unregister `useEffect` block is copy-pasted verbatim across `PageFeed.renderer.tsx`, `PageStandard.renderer.tsx`, and `PageStickyFooter.renderer.tsx`. All three iterate `page.bottom_sheets`, call `register(sheet.id, { id, title, size, items, on_dismiss, on_open })`, and return a cleanup that calls `unregister(sheet.id)`. This was flagged in the previous round and remains unchanged. Extract a `useRegisterBottomSheets(bottomSheets: BottomSheet[] | undefined)` hook (or a shared helper) so the logic lives in one place. When the mapping shape changes (e.g. a new field is added to `SheetEntry`), it currently requires three identical edits.

zenith-oi (approved):
## Zenith Code Review

> ⚠️ **Round 6 — engineer attention requested.** The reviewer-engineer loop has reached round 6; this is the threshold where auto-iteration typically stops being productive (either the review is missing context the engineer has, or there's a genuine design call to make). Auto-fixes are still posted, but please review this thread end-to-end rather than relying on another round to converge.

**This PR correctly fixes the root cause of SDUI bottom sheets never appearing by bridging gorhom's imperative API through a registry-based Zustand store. The implementation is sound and well-tested. One previously-flagged DRY issue (copy-pasted register/unregister `useEffect` across three page renderers) remains unaddressed.**

### What this PR is doing
**Intent:** Rewire the SDUI bottom-sheet runtime from a declarative stack model to a registry/open-flag model that correctly drives gorhom's imperative `BottomSheetModal` API (`present()`/`dismiss()` via ref), fixing the root cause of sheets never appearing.
**Delivered:** The diff fully delivers the stated intent. The store is reorganised into `registry`/`openSheets`/`openOrder`/`contexts` with `register`/`unregister`/`open`/`close`/`closeAll` actions. `BottomSheetHost` now renders per-sheet children that own their own refs and bridge the declarative open flag to imperative calls via `useEffect`. All three page renderers and the `BottomSheet` snippet renderer pre-register on mount and unregister on unmount. The `sheet` action handler calls `open(id)` (registry lookup). `useBottomSheetData` uses `useShallow` for atomic snapshot. 216-line test suite covers all store actions. No acceptance items are missing.

**Positive:**
- The `openOrder` array as an explicit ordering mechanism (rather than relying on JS object key insertion order) is a well-reasoned design choice — the comment explaining why is exactly the right level of documentation.
- Using `useShallow` in `useBottomSheetData` to take an atomic snapshot across `openOrder` + `registry` is the correct fix for React 18 concurrent rendering torn-state risk.
- The `BottomSheetHostSheet` component cleanly separates the imperative ref concern from the parent host, and the `useEffect([open])` bridge is the idiomatic way to drive gorhom's modal API declaratively.
- 216-line test suite covers all store actions including edge cases (no-op on unregistered id, close-without-id on empty stack, reopen promotes to topmost, full round-trip). The use of `node:test` + `node:assert/strict` without a heavy test framework is appropriate for a pure store unit test.

---

zenith-oi (approved):
## Zenith Code Review

> ⚠️ **Round 6 — engineer attention requested.** The reviewer-engineer loop has reached round 6; this is the threshold where auto-iteration typically stops being productive (either the review is missing context the engineer has, or there's a genuine design call to make). Auto-fixes are still posted, but please review this thread end-to-end rather than relying on another round to converge.

**This PR correctly upstreams the imperative-API fix for gorhom `BottomSheetModal` and is well-structured. One previous FLAG (DRY violation across three page renderers) remains unaddressed; no new blocking issues are introduced.**

### What this PR is doing
**Intent:** Rewire the SDUI bottom-sheet runtime from a declarative stack model to a registry/open-flag model that correctly drives gorhom's imperative `BottomSheetModal` API (`present()`/`dismiss()` via ref), fixing sheets that were never visible because `present()` was never called.
**Delivered:** The diff fully delivers the stated intent. The store is reorganised into `registry`/`openSheets`/`openOrder`/`contexts` with `register`/`unregister`/`open`/`close`/`closeAll` actions. `BottomSheetHost` now renders one `BottomSheetHostSheet` per registered sheet, each owning its own ref and bridging the declarative `open` flag to imperative `present()`/`dismiss()` calls. Page renderers and the `BottomSheet` snippet renderer pre-register on mount and unregister on unmount. The `sheet` action handler calls `open(sheet_id)` (registry lookup). 216 unit tests cover the store's state machine. The changeset is present. No acceptance items from a linked task page exist to check against.

**Positive:**
- The `openOrder` array as an explicit ordering mechanism (rather than relying on JS object key insertion order) is a thoughtful design choice — the comment explaining why it exists is exactly the right level of documentation.
- 216 unit tests covering every action (register, unregister, open, close, closeAll) including edge cases (no-op on unregistered id, close-without-id, reopen promotes to topmost, full round-trip) give strong confidence in the store's state machine.
- Using `useShallow` in `useBottomSheetData` to take an atomic snapshot of `openOrder` + `registry` is the correct fix for React 18 concurrent-rendering torn-state risk — and the comment explaining why is clear.
- The `BottomSheetHostSheet` component cleanly separates the imperative bridge concern (ref + useEffect) from the host's declarative rendering loop, making the gorhom API mismatch explicit and contained.

---

packages/sdui-runtime/src/bottom-sheet/BottomSheetHost.tsx
```diff
@@ -12,13 +15,64 @@ const SIZE_TO_SNAP: Record<string, string[]> = {
...
+ * calls via a `useEffect` keyed on `open`.
+ */
+function BottomSheetHostSheet({
+  sheet,
+  open,
+  onDismiss,
+}: BottomSheetHostSheetProps): React.ReactElement {
+  const ref = useRef<BottomSheetModalMethods>(null);
+
+  useEffect(() => {
```
zenith-oi, line 40: **FLAG** — [defensive-effect-guard] The `useEffect` runs on mount with `open === false` and calls `ref.current?.dismiss()`. On initial mount `ref.current` is `null` so the optional-chain makes this a no-op today, but if gorhom ever initialises the ref before the first effect fires (or if the component is mounted with `open: true` then immediately toggled to `false` before the ref attaches), the dismiss path could race. More concretely: every newly-registered sheet that is not open will call `dismiss()` on every render cycle where `open` changes to `false` — including the very first render. A simple guard `if (open) { ref.current?.present(); } else if (ref.current) { ref.current.dismiss(); }` avoids the spurious dismiss on initial mount and makes the intent explicit.

packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
zenith-oi, line 68: **FLAG** — [dry — duplicated registration logic] The `useEffect` that registers/unregisters `page.bottom_sheets` is copy-pasted verbatim across `PageFeed.renderer.tsx`, `PageStandard.renderer.tsx`, and `PageStickyFooter.renderer.tsx` (all three are identical modulo the comment). Extract a `useRegisterBottomSheets(sheets, register, unregister)` custom hook and call it from each renderer. This is a DRY violation across 3 files — the next time the registration shape changes (e.g. a new field on `SheetEntry`) it will need to be updated in three places. Also applies to `PageStandard.renderer.tsx` (L32) and `PageStickyFooter.renderer.tsx` (L59).

zenith-oi (approved):
## Zenith Code Review

**This PR correctly upstreams the imperative gorhom `BottomSheetModal` fix that was previously patched in the creator-app's `node_modules`. The implementation is clean, well-tested, and the changeset is accurate. One genuine concern: the `BottomSheetHostSheet` `useEffect` calls `ref.current?.dismiss()` on initial mount (when `open` is `false`), which is a no-op today but could cause a flicker or unexpected dismiss if gorhom's internal state is ever non-null at mount time — worth a defensive guard.**

### What this PR is doing
**Intent:** Rewire the SDUI bottom-sheet runtime from a declarative stack model to a registry/imperative model that correctly calls gorhom's `BottomSheetModal` `present()`/`dismiss()` API, fixing sheets that were never actually visible.
**Delivered:** The diff fully delivers the stated intent. The store is reorganised around `registry`/`openSheets`/`openOrder`/`contexts` with `register`/`unregister`/`open`/`close`/`closeAll` actions. `BottomSheetHost` now renders one `BottomSheetHostSheet` per registered sheet, each owning its own ref and bridging the declarative open flag to imperative `present()`/`dismiss()` calls. All three page renderers and the `BottomSheet` snippet renderer pre-register on mount and unregister on unmount. The `sheet` action handler calls `open(sheet_id)` via registry lookup. Unit tests cover all new store actions. No linked task page or brief was provided, so assessment is based on the PR body alone.

**Positive:**
- The `openOrder` array as an explicit ordering primitive (rather than relying on JS object key insertion order) is a thoughtful design decision — the comment in the store explaining why this is necessary under Zustand persist middleware is exactly the kind of 'why' documentation that saves future engineers.
- The `useShallow` usage in `useBottomSheetData` is correct and the comment explaining the React 18 concurrent-rendering torn-state risk is genuinely useful.
- Test coverage is thorough — 16 tests covering register, open, close, closeAll, unregister, openOrder promotion, and round-trip reopening. The `console.warn` stub pattern in the no-op test is clean.
- The `BottomSheetHostSheet` component decomposition (one child per sheet, each owning its own ref) is the right architectural choice for gorhom's imperative API — avoids the ref-array management complexity that would arise from a single host managing multiple refs.

---

PR #151, opened 2026-05-26
fix(sdui-runtime): SduiNode parses defensively, renders SduiFallback on ZodError

mridulgupta-oi:
## Summary

- Wrap `SduiNode`'s `schema.parse(...)` call in a small `parseNodeData` helper that returns a discriminated `ParseResult` (ok/err).
- On `ZodError`: emit `sdui.node.parse_error` telemetry (with node id / type / message) and render `SduiFallback` instead of crashing the page.
- On any non-Zod error: re-throw (real bugs still surface).
- `SduiFallback` now accepts optional `nodeId`, `nodeType`, and `error` props. In `__DEV__` it renders a discreet dashed-red placeholder so the failing node is easy to spot in the simulator; in production it stays empty.
- Add 6 unit tests for the pure parse helper (happy path, multiple shape mismatches, null / undefined input, schema strip behaviour, and the non-Zod-error rethrow path).

## Why

`SduiNode` is the single chokepoint every renderer flows through, and the naked `props.schema.parse(props.data)` call threw on every malformed or stale wire payload — taking the whole page down via the nearest error boundary. During the SDUI migration window where handler emits and renderers are constantly evolving, that's a real risk every time the gateway and runtime drift apart.

Falling back to a discreet placeholder + telemetry event keeps one bad node from disrupting the rest of the page while still surfacing the failure in dashboards and (visibly) in the simulator.

## Test plan

- [x] `npm test -w packages/sdui-runtime` — 6 new tests pass, no existing tests regressed (3 pre-existing baseline failures in `branch.test.ts` / `compound.test.ts` / `set-local.test.ts` are unchanged and unrelated to this PR).
- [x] `npm run build -w packages/sdui-runtime` succeeds with no type errors.
- [ ] Manual verification: feed a deliberately malformed node into a creator-app page and confirm the surrounding nodes still render + the fallback box shows in dev.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #164, opened 2026-05-26
fix(sdui-runtime): Interpreter normalises legacy props→data on render

mridulgupta-oi:
## Summary

The SDUI Interpreter now normalises `props`-shaped wire payloads to the canonical `data` field before resolving the renderer. When a node arrives with `props` but no `data`, the props are copied into `data` and a `sdui.interpreter.props_normalized` telemetry event is emitted so stale BFF emits are easy to spot in dev. Nodes that already have `data` are passed through untouched — the helper never overwrites the canonical field.

## Why

The wire contract per `@one-impression/sdui-primitives@1.0.0` places renderer payloads under `node.data`. Stale or legacy handler emits sometimes still ship them under `node.props`, which silently breaks rendering during the SDUI migration window. This is the sibling resilience pattern to `SduiNode`'s defensive `ZodError` handling (#151): one bad emit shouldn't take down the whole tree, and migration drift should leave a breadcrumb.

The normalizer is a pure function in `normalize-node.ts` so it's trivial to unit-test under `node:test` (no React, no jest harness).

## Test plan

- [x] 5 new `node:test` cases under `interpreter/__tests__/normalize-node.test.ts` cover:
  - valid `data` passes through unchanged (same reference, no clone)
  - only `props` gets rewritten to `data`, and `props` is stripped
  - both `data` and `props` keeps `data` (never overwrite)
  - neither field is untouched
  - input node is not mutated
- [x] `npm test -w packages/sdui-runtime` — 5/5 new tests pass; 4 pre-existing failures (`branch.test.ts` / `compound.test.ts` / `set-local.test.ts` / `bff-call.test.ts`) are the known peer-dep baseline issue, unrelated to this PR
- [ ] Manual: render a legacy node with `props` in the creator-app simulator and confirm it renders + telemetry fires

### DIFF
--- a/.changeset/sdui-interpreter-props-normalize.md
+++ b/.changeset/sdui-interpreter-props-normalize.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+`Interpreter` now normalises legacy `props`-shaped wire payloads to the canonical `data` field before resolving the renderer. When a node arrives with `props` but no `data`, the props are copied into `data` and a `sdui.interpreter.props_normalized` telemetry event is emitted so stale BFF emits are easy to spot in dev. Nodes that already have `data` are passed through untouched — the helper never overwrites the canonical field. Defensive migration aid: matches the `data` contract from `@one-impression/sdui-primitives@1.0.0` while old handler emits are still being migrated.
--- a/packages/sdui-runtime/package.json
+++ b/packages/sdui-runtime/package.json
@@ -34,7 +34,7 @@
     "build": "tsup && tsc --emitDeclarationOnly --noCheck",
     "dev": "tsup --watch",
     "lint": "eslint src/",
-    "test": "tsx --test src/action-engine/cond/__tests__/*.test.ts src/action-engine/handlers/__tests__/*.test.ts src/bottom-sheet/__tests__/*.test.ts src/snippets/_shared/__tests__/*.test.ts src/state/__tests__/*.test.ts src/snippets/Form/__tests__/*.test.ts src/sdui-node/__tests__/*.test.ts"
+    "test": "tsx --test src/action-engine/cond/__tests__/*.test.ts src/action-engine/handlers/__tests__/*.test.ts src/bottom-sheet/__tests__/*.test.ts src/snippets/_shared/__tests__/*.test.ts src/state/__tests__/*.test.ts src/snippets/Form/__tests__/*.test.ts src/sdui-node/__tests__/*.test.ts src/interpreter/__tests__/*.test.ts"
   },
   "peerDependencies": {
     "react": ">=18.0.0",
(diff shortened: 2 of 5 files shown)


zenith-oi (approved):
## Zenith Code Review

**Clean, well-scoped migration-resilience patch. The normalizer is a pure function, the Interpreter integration is minimal, tests cover every branch including the immutability invariant, and the changeset accurately describes the change. No blocking issues found.**

### What this PR is doing
**Intent:** Add a `normalizeNode` helper to `sdui-runtime`'s Interpreter that silently upgrades legacy `props`-shaped wire payloads to the canonical `data` field, emitting a telemetry event when it does so, so stale BFF emits don't silently break rendering during the SDUI migration window.
**Delivered:** The diff fully delivers the stated intent. `normalize-node.ts` implements the pure normalization function with the correct never-overwrite semantics; `Interpreter.tsx` wires it in before renderer resolution; the changeset documents the patch; the test suite covers all five cases described in the PR body (pass-through, props→data rewrite, both-fields-keeps-data, neither-field, immutability). No acceptance items from a linked task page are present, so evaluation is against the PR body alone — all stated goals are met.

No issues found.

**Positive:**
- The normalizer is a pure function with no side effects — the telemetry emit lives in the React component where it belongs, not inside the helper. This makes the helper trivially testable under `node:test` without any React or telemetry harness.
- The never-overwrite invariant (`data` present → pass through unchanged, `didNormalize: false`) is explicitly tested and the test comment explains the design rationale — good defensive documentation.
- The immutability test (JSON.stringify before/after) is a concrete guard against the most common mistake in this kind of shim (mutating the caller's input). Worth calling out as a model for similar migration helpers.
- Returning the same object reference on the no-op path (`return { normalized: node, didNormalize: false }`) avoids unnecessary allocations on the hot render path — a small but correct performance choice.

**Read against:**
- `PR body — stated intent, test plan, and five test cases`
- `packages/sdui-runtime/src/interpreter/normalize-node.ts — implementation`
- `packages/sdui-runtime/src/interpreter/Interpreter.tsx — integration point`
- `packages/sdui-runtime/src/interpreter/__tests__/normalize-node.test.ts — test coverage`

---

mridulgupta-oi:
Closing — this would add a defensive runtime workaround that masks BFF contract drift.

Per the user's strict-contract principle (just re-locked this morning):
- `NodeSchema.data` is required in `@one-impression/sdk-native-sdui`
- `SduiNode` already throws ZodError on missing `data` field and renders `SduiFallback` (PR #151, already shipped in 2.1.0)
- ESLint `no-restricted-syntax` rule #2 (`bff-no-props-field`) on amplify-gateway blocks any new `props:` Property in handler code

The 74 legacy `props:` handlers on amplify-gateway's `shadow` are baselined (`.eslint-baseline.json` — 292 `no-restricted-syntax` entries). Each Phase 1 handler PR shrinks the baseline by migrating one BC. We want them to render as `SduiFallback` (visibly broken) until migrated — silently normalizing at runtime hides the drift.

Not opening a follow-up. Closing in favor of the existing strict-at-the-boundary enforcement.

PR #166, opened 2026-05-26
fix(sdui-runtime): resolveRenderer returns null on unknown type (graceful degrade)

mridulgupta-oi:
## Summary

`resolveRenderer` now returns `null` explicitly (not `undefined`, never throws) when a node type isn't registered. The first occurrence of each unknown type emits a `sdui.renderer.unknown_type` telemetry event through the sink wired by `SduiRuntimeProvider`; subsequent occurrences are silenced via an internal `Set` so a recurring bad node in a feed doesn't flood telemetry.

## Why

One bad node should never crash the whole page. Returning `undefined` from `resolveRenderer` worked as long as every caller did a truthy check before invoking — that contract was implicit and brittle. Tightening it to `null` makes the "no renderer" path explicit and gives us a single, throttled telemetry breadcrumb so stale or unmigrated handler types are easy to spot in dev. Sibling pattern to `SduiNode`'s defensive `ZodError` handling (#151).

## Test plan

- [x] 6 new `node:test` cases under `registries/__tests__/node-registry.test.ts`:
  - known `ui_component` type returns a renderer
  - known `snippet` type returns a renderer
  - unknown type returns `null` (does not throw)
  - unknown bare type (no `creator.` prefix) returns `null`
  - warning fires exactly once per unknown type (suppressed on repeat)
  - distinct unknown types each warn once
- [x] `npm test -w packages/sdui-runtime` — the new test file hits the same pre-existing peer-dep baseline failure (`@one-impression/sdk-native-sdui` unresolved at runtime in the workspace) as `branch.test.ts` / `compound.test.ts` / `set-local.test.ts` / `bff-call.test.ts`. That is a separate workspace issue, not introduced here. The logic is straightforward enough that the test asserts are reviewable inline.
- [ ] Manual: drop an unknown node type into a page in the creator-app simulator and confirm the page renders + the warning fires once

### DIFF
--- a/.changeset/sdui-resolve-renderer-graceful.md
+++ b/.changeset/sdui-resolve-renderer-graceful.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+`resolveRenderer` now returns `null` (not `undefined`, never throws) when a node type isn't registered, so one bad node degrades gracefully instead of taking down the page. The first occurrence of each unknown type emits `sdui.renderer.unknown_type` through the telemetry sink wired by `SduiRuntimeProvider`; subsequent occurrences are silenced via an internal `Set` to avoid flooding telemetry. Sibling pattern to `SduiNode`'s defensive `ZodError` handling.
--- a/packages/sdui-runtime/package.json
+++ b/packages/sdui-runtime/package.json
@@ -34,7 +34,7 @@
     "build": "tsup && tsc --emitDeclarationOnly --noCheck",
     "dev": "tsup --watch",
     "lint": "eslint src/",
-    "test": "tsx --test src/action-engine/cond/__tests__/*.test.ts src/action-engine/handlers/__tests__/*.test.ts src/bottom-sheet/__tests__/*.test.ts src/snippets/_shared/__tests__/*.test.ts src/state/__tests__/*.test.ts src/snippets/Form/__tests__/*.test.ts src/sdui-node/__tests__/*.test.ts"
+    "test": "tsx --test src/action-engine/cond/__tests__/*.test.ts src/action-engine/handlers/__tests__/*.test.ts src/bottom-sheet/__tests__/*.test.ts src/snippets/_shared/__tests__/*.test.ts src/state/__tests__/*.test.ts src/snippets/Form/__tests__/*.test.ts src/sdui-node/__tests__/*.test.ts src/registries/__tests__/*.test.ts"
   },
   "peerDependencies": {
     "react": ">=18.0.0",
--- a/packages/sdui-runtime/src/SduiRuntimeProvider.tsx
+++ b/packages/sdui-runtime/src/SduiRuntimeProvider.tsx
@@ -5,6 +5,7 @@ import { ActionEngineContext } from "./action-engine/useActionEngine.js";
 import { TelemetryContext } from "./telemetry/useTelemetry.js";
 import type { TelemetryEmitter } from "./telemetry/useTelemetry.js";
 import type { ActionEngineConfig } from "./action-engine/types.js";
+import { setResolveRendererWarnSink } from "./registries/node-registry.js";
 
 interface BffConfig {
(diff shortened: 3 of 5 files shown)


zenith-oi (approved):
## Zenith Code Review

**Clean, focused defensive-rendering improvement. `resolveRenderer` now has an explicit null contract, telemetry is throttled per unknown type, and the implementation is well-tested. No correctness, security, or architecture issues found.**

### What this PR is doing
**Intent:** Make `resolveRenderer` return `null` (never `undefined`, never throw) for unknown node types, and emit a deduplicated `sdui.renderer.unknown_type` telemetry event through the `SduiRuntimeProvider`-wired sink on first occurrence of each unknown type.
**Delivered:** The diff fully delivers the stated intent. `resolveRenderer` now returns `ComponentType<Node> | null` instead of `ComponentType<Node> | undefined`, the `if/if` prefix-check branches are collapsed into `if/else if` (fixing a latent double-lookup bug), the warn-sink + dedup `Set` are wired correctly, `SduiRuntimeProvider` connects the sink to the telemetry emitter via a memoised `useMemo`, and 6 test cases cover all the described behaviours. No acceptance items from a linked task page are present, but the PR body's stated goals are all addressed.

No issues found.

**Positive:**
- The `if/else if` refactor (replacing the original `if/if` shape) is a genuine correctness improvement — the old code would fall through to the second prefix check even after the first matched, which was harmless only because the prefixes are mutually exclusive, but the new shape is unambiguously correct.
- Exporting `_unknownTypesWarned` and `_resetResolveRendererState` with underscore-prefixed names and explicit 'test-only' JSDoc is the right idiom for test-seam exports — makes the intent clear without polluting the public API surface.
- Using `useMemo` keyed on `telemetryConfig.emitter` to re-wire the sink is the correct React pattern here — it handles test-rig swaps and avoids stale-closure issues without needing a `useEffect` cleanup.
- The changeset description is unusually thorough and accurate — it describes the exact contract change, the dedup mechanism, and the sibling pattern reference, which is exactly what a consumer of the package needs to understand the patch.

---

zenith-oi (approved):
## Zenith Code Review

**Clean, focused defensive-rendering improvement. The implementation is correct, the deduplication logic is sound, the telemetry wiring is idiomatic, and the test coverage is thorough. One minor concern worth flagging: the module-level mutable singletons (`warnSink` and `_unknownTypesWarned`) are a deliberate design choice that works fine in a browser bundle but could cause test-isolation issues if tests run in parallel — the `_resetResolveRendererState` helper mitigates this and is used consistently in the test file.**

### What this PR is doing
**Intent:** Make `resolveRenderer` return `null` (never `undefined`, never throw) for unknown node types, and emit a deduplicated `sdui.renderer.unknown_type` telemetry event through the `SduiRuntimeProvider`-wired sink on first occurrence of each unknown type.
**Delivered:** Fully delivered. The diff changes the return type from `ComponentType<Node> | undefined` to `ComponentType<Node> | null`, adds a module-level `warnSink` with a `setResolveRendererWarnSink` setter, deduplicates warnings via `_unknownTypesWarned: Set<string>`, wires the sink in `SduiRuntimeProvider` via `useMemo`, adds 6 test cases covering all stated behaviours, and registers the new test file in `package.json`. The changeset entry accurately describes the change.

No issues found.

**Positive:**
- The `if/else if` refactor (replacing two independent `if` blocks) correctly prevents a type that starts with `creator.ui_component.` from also being checked against the snippet registry — a subtle correctness fix bundled cleanly with the main change.
- Deduplication via a `Set` with a noop default sink is the right shape: zero overhead on the hot path when telemetry isn't configured, and no telemetry flooding when it is.
- The `useMemo` dependency on `telemetryConfig.emitter` means swapping the emitter (e.g. in test rigs) automatically re-wires the sink — good defensive design.
- Test coverage is comprehensive: known types, unknown types, bare-prefix types, single-warn deduplication, and multi-type deduplication are all exercised. The test file is self-contained and readable.

---

PR #149, opened 2026-05-26
feat(sdui-runtime): add usePageStore.replaceNode + appendItems

mridulgupta-oi:
## Summary

- Add `page: Page | null` state to `usePageStore` (the full SDUI page envelope).
- Add `setPageTree(page)` setter for the initial page fetch.
- Add `replaceNode(targetId, next)` — recursive node-tree replacement by id.
- Add `appendItems(targetId, items, options?)` — append to `data.items` and merge optional `cursor` / `has_more` into the target's `data`.
- Extract pure tree-walkers (`replaceNodeInTree`, `appendItemsInTree`) into an `__internal` namespace and add 13 unit tests covering happy paths, no-op fallbacks, and warning emission.

## Why

Three action handlers were calling methods on `usePageStore` that did not exist, silently crashing the action dispatch loop:

| Handler | Call site | Method called |
|---|---|---|
| `handleReloadSection` | `src/action-engine/handlers/reload-section.ts:40` | `replaceNode(target, node)` |
| `handleReplaceSection` | `src/action-engine/handlers/replace-section.ts:16` | `replaceNode(target, with_node)` |
| `handleAppendItems` | `src/action-engine/handlers/append-items.ts:17` | `appendItems(target, items, { cursor, hasMore })` |

The store previously only exposed section-granularity setters (`replaceSection`, `appendSection`) keyed by a flat sections map. The action handlers, however, address nodes by id anywhere in the page tree — which is what these new methods now support. Both methods are no-ops with `console.warn` when the target id is missing or has no `data.items` array, so stale actions degrade gracefully.

## Consumer files

- `packages/sdui-runtime/src/action-engine/handlers/reload-section.ts`
- `packages/sdui-runtime/src/action-engine/handlers/replace-section.ts`
- `packages/sdui-runtime/src/action-engine/handlers/append-items.ts`

## Test plan

- [x] `npm test -w packages/sdui-runtime` — 13 new tests pass, no existing tests regressed (3 pre-existing baseline failures in `branch.test.ts` / `compound.test.ts` / `set-local.test.ts` are unchanged and unrelated to this PR).
- [x] `npm run build -w packages/sdui-runtime` succeeds with no type errors.
- [ ] Manual verification: trigger `reload_section`, `replace_section`, and `append_items` actions from the creator-app simulator and confirm the page tree updates without dispatch errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #167, opened 2026-05-26
fix(sdui-runtime): PageFeedRenderer subscribes to usePageStore for reactive item updates

mridulgupta-oi:
## Summary

- `PageFeedRenderer` now syncs the server-provided page tree into `usePageStore` on mount via `setPageTree(page)`.
- The FlatList reads `items` reactively from the store with a `useShallow` selector scoped by `page.id`.
- Falls back to the `page.items` prop when the store hasn't been populated yet or currently holds a different page (e.g. mid-navigation), preserving correct initial-render behavior.

## Why

`replaceNode` / `appendItems` were added to `usePageStore` in #149 specifically so action handlers (`reload_section`, `replace_section`, `append_items`) could mutate page state without going through a full re-fetch. But `PageFeedRenderer` was reading `page.items` from props only — the store updated, no re-render fired, and the feed never reflected the new items.

That breaks end-to-end:
- Infinite scroll: `on_load_more` → `bff_call` → response carries `append_items` → store updates → **feed silently stale**.
- Section reload: `reload_section` → store `replaceNode` → **feed silently stale**.

The home-tabs section-reload pattern (the headline use case for #149) cannot work without this subscription. Closing the gap is a strict bug fix — additive, no API change.

## Test plan

- [x] `npm run build -w packages/sdui-runtime` — green.
- [x] 5 new tests in `usePageStore.test.ts` covering the exact selector PageFeedRenderer uses:
  - initial render returns prop items when store empty,
  - after `setPageTree`, returns store items,
  - `replaceNode` on a top-level item flows through,
  - `appendItems` on a container flows through,
  - different page id in store falls back to prop items (navigation transition guard).
- [x] Existing 73 store/handler tests still pass (77 → 82 total, +5 new).

Note: the `bff-call`/`branch`/`compound`/`set-local` test failures on `main` are pre-existing module-resolution issues unrelated to this change.

### DIFF
--- a/.changeset/sdui-page-feed-store-subscription.md
+++ b/.changeset/sdui-page-feed-store-subscription.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+`PageFeedRenderer` now subscribes to `usePageStore` for its `items` data so `replaceNode` and `appendItems` (added in #149) take effect end-to-end. On mount the renderer syncs the server-provided page tree into the store via `setPageTree(page)`; the FlatList then reads `items` reactively with a `useShallow` selector scoped by `page.id`, falling back to the prop value when the store hasn't been populated yet or holds a different page. Without this, infinite-scroll `append_items` and section reload were updating the store but never re-rendering the feed.
--- a/packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
+++ b/packages/sdui-runtime/src/pages/PageFeed/PageFeed.renderer.tsx
@@ -6,12 +6,14 @@ import {
   StyleSheet,
   View,
 } from "react-native";
+import { useShallow } from "zustand/react/shallow";
 import type { Page, Node, Action } from "@one-impression/sdk-native-sdui";
 import { Interpreter } from "../../interpreter/index.js";
 import { useActionEngine } from "../../action-engine/useActionEngine.js";
 import { useBottomSheetStore } from "../../bottom-sheet/useBottomSheetStore.js";
 import { usePageRefresh } from "../../hooks/usePageRefresh.js";
 import { useAppStateSession } from "../../hooks/useAppStateSession.js";
+import { usePageStore } from "../../state/usePageStore.js";
 
 interface PageProps {
   page: Page;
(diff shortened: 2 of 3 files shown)


zenith-oi (approved):
## Zenith Code Review

**Clean, focused bug fix that closes the store-subscription gap in PageFeedRenderer. The implementation is correct, the selector logic is sound, and the test coverage directly mirrors the selector contract. No issues found.**

### What this PR is doing
**Intent:** Wire PageFeedRenderer to subscribe to usePageStore so that replaceNode and appendItems mutations (added in #149) actually trigger re-renders of the FlatList, closing the end-to-end gap where infinite-scroll and section-reload actions updated the store but the feed never reflected the changes.
**Delivered:** The diff fully delivers the stated intent. The useEffect syncs the server page tree into the store on mount/page-prop-change, the useShallow selector reads items reactively with a correct page-id guard and a safe fallback to prop items, and the FlatList data binding is updated from the static prop to the reactive value. Five targeted tests cover the selector contract including the navigation-transition edge case. No acceptance items from a linked task page are present, but the PR body's stated goals are all addressed in the diff.

No issues found.

**Positive:**
- The useShallow selector is correctly scoped by page.id — it guards against torn snapshots during navigation transitions where the store might hold a different page, which is a subtle concurrent-rendering correctness concern that is easy to miss.
- Calling usePageStore.getState().setPageTree(page) inside useEffect (rather than during render) is the right pattern — it avoids triggering a synchronous store mutation during the render phase.
- The test suite mirrors the exact selector shape used in the renderer (selectItemsForPage helper), making the tests a precise contract test rather than an indirect integration test. The navigation-transition guard test in particular covers a real edge case.

**Read against:**
- `PR body — stated intent: sync setPageTree on mount, reactive FlatList via useShallow selector, fallback to prop items when store empty or holds different page`
- `PR body — test plan: 5 new tests covering initial render, post-setPageTree, replaceNode, appendItems, different-page fallback`

---

zenith-oi (approved):
## Zenith Code Review

**Clean, focused bug fix that closes the store-subscription gap in `PageFeedRenderer`. The implementation is correct, the selector logic is sound, and the tests directly exercise the selector contract. No issues found.**

### What this PR is doing
**Intent:** Wire `PageFeedRenderer` to subscribe to `usePageStore` so that `replaceNode` and `appendItems` mutations (added in #149) actually trigger re-renders of the FlatList, closing the end-to-end gap for infinite-scroll and section-reload action handlers.
**Delivered:** The diff fully delivers the stated intent. The `useEffect` on L73-75 syncs the server page tree into the store on mount/page-prop change. The `useShallow` selector on L83-90 reads `items` reactively with a correct page-ID guard and a safe fallback to prop values for the first-render race and navigation-transition cases. The FlatList on L198 is switched from `page.items` to the reactive `items`. Five targeted tests on L264-L326 cover the selector contract end-to-end including the fallback path. No acceptance items from a linked task page are present, but the PR body's stated goals are all addressed.

No issues found.

**Positive:**
- The `useShallow` selector is scoped by both `pageId` AND `page` presence, preventing torn reads under React 18 concurrent rendering — this is exactly the right defensive shape for a store that can hold a different page during navigation transitions.
- Calling `usePageStore.getState().setPageTree(page)` inside `useEffect` (rather than during render) is correct — it avoids triggering a synchronous store update on the render path that could cause tearing or double-render.
- The fallback `return page.items` in the selector correctly handles the first-render window before the `useEffect` commits, meaning the FlatList is never empty on initial paint.
- Tests mirror the exact selector shape used in the renderer (`selectItemsForPage` helper on L256-L262), making them a reliable contract test rather than an indirect integration test.

**Read against:**
- `PR body — stated goals: infinite-scroll append_items and section reload re-render`
- `PR body — test plan: 5 new selector tests covering empty store, setPageTree, replaceNode, appendItems, different-page fallback`

---

PR #169, opened 2026-05-26
fix(ui-native): Box forwards position, zIndex, opacity, overflow, individual borderWidths

mridulgupta-oi:
## Summary

Forwards five style-prop categories that `Box` was silently dropping. All additive — no existing prop changes shape or behavior.

- `position` (`'absolute' | 'relative' | …`)
- `zIndex` (`number`)
- `opacity` (`0..1`)
- `overflow` (`'hidden' | 'visible' | 'scroll'`)
- Individual border widths: `borderTopWidth`, `borderBottomWidth`, `borderLeftWidth`, `borderRightWidth` — each accepts `BorderWidthToken | number`, the same shape as the existing `borderWidth` prop, resolved through `resolveBorderWidth`.

## Why

SDUI handlers emit these props on `Box` nodes for legitimate visual cases that currently render incorrectly:

- **Absolute-positioned overlays** (badges, floating buttons, tooltips, sticky headers) need `position` + `zIndex`.
- **Animated transitions / hide-without-unmount** need `opacity`.
- **Scroll containers and clipping** need `overflow`.
- **One-sided dividers / asymmetric borders** need the individual `border*Width` props.

Until now these were stripped at the Box boundary, so every handler that needed them either fell back to wrapping a raw `<View>` (defeating the design-system layer) or relied on a manual ui-native dist patch. Forwarding them at the source closes that gap.

## Test plan

- [x] `npm run build -w packages/ui-native` — green.
- [x] New `Box.test.tsx` with cases for each new prop reaching the View's flattened style, plus a "no leak when unset" check and a "coexists with existing layout props" check.
- [x] No change to existing props' resolution path (verified by re-reading existing prop-mapping code).

Note: the ui-native workspace `jest` binary isn't installed in CI today, so the Box test file lands alongside the 18 existing untested-in-CI sibling test files. That's a pre-existing infra gap, not introduced here. Box tsup build succeeds; the only Box-related tsc errors (`'View' cannot be used as a JSX component`) are pre-existing across every component on main due to a React version mismatch.

### DIFF
--- a/.changeset/ui-native-box-style-props.md
+++ b/.changeset/ui-native-box-style-props.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/ui-native": patch
+---
+
+`Box` now forwards five additional React Native style props that were previously silently dropped: `position`, `zIndex`, `opacity`, `overflow`, and the four individual border widths (`borderTopWidth`, `borderBottomWidth`, `borderLeftWidth`, `borderRightWidth`, all of which accept either a `BorderWidthToken` or a raw number). SDUI handlers emit these on Box nodes for absolute-positioned overlays, animated fades, and one-sided dividers — without forwarding, those layouts rendered without their styling. Additive only — existing props are untouched.
--- a/packages/ui-native/src/layout/Box.test.tsx
+++ b/packages/ui-native/src/layout/Box.test.tsx
@@ -1,0 +1,94 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react-native';
+import { Text as RNText } from 'react-native';
+import { Box } from './Box';
+
+/** Flatten Box's `style` array into a single object for assertions. */
+function flatStyle(testID: string): Record<string, unknown> {
+  const el = screen.getByTestId(testID);
+  const s = el.props.style;
+  return Array.isArray(s)
+    ? Object.assign({}, ...(s.filter(Boolean) as object[]))
+    : (s as Record<string, unknown>);
+}
+
+describe('Box', () => {
+  it('renders children', () => {
+    render(
+      <Box testID="box">
+        <RNText>Hello</RNText>
+      </Box>,
+    );
+    expect(screen.getByTestId('box')).toBeTruthy();
+    expect(screen.getByText('Hello')).toBeTruthy();
+  });
+
+  it('forwards position prop to underlying View style', () => {
+    render(<Box testID="box" position="absolute" />);
+    expect(flatStyle('box').position).toBe('absolute');
+  });
+
+  it('forwards zIndex prop', () => {
+    render(<Box testID="box" zIndex={10} />);
+    expect(flatStyle('box').zIndex).toBe(10);
+  });
+
+  it('forwards opacity prop', () => {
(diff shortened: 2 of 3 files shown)


zenith-oi (approved):
## Zenith Code Review

**Clean, additive prop-forwarding change with good test coverage. No linked brief or task page; reviewed against diff + PR body only. No issues found.**

### What this PR is doing
**Intent:** Add five previously-silently-dropped style prop categories to the `Box` component in `@one-impression/ui-native` (`position`, `zIndex`, `opacity`, `overflow`, and individual border widths), forwarding them through `resolveBorderWidth` / direct pass-through into the underlying View's style object, so SDUI handlers can use them without falling back to raw `<View>` wrappers.
**Delivered:** Fully delivered. The diff adds the props to the `BoxProps` interface, destructures them in the `forwardRef` implementation, resolves the border-width variants through the existing `resolveBorderWidth` helper, and passes the four layout props directly into the style object. A comprehensive test file covers each new prop individually, the 'no-leak-when-unset' case, and coexistence with existing layout props. The changeset file is present and correctly scoped as a patch. No existing prop's resolution path is touched.

No issues found.

**Positive:**
- The 'does not leak the new props onto the underlying View when unset' test case (Box.test.tsx L63-L74) is exactly the right regression guard for a component that strips undefined values — it would catch any future accidental default-value injection.
- Reusing the existing `resolveBorderWidth` helper for all four individual border-width props keeps the token-resolution path consistent with `borderWidth` rather than introducing a parallel raw-number path.
- The changeset description is unusually informative — it names the SDUI use-cases that motivated each prop, which makes future archaeology straightforward.

---

zenith-oi (approved):
## Zenith Code Review

**Clean, additive prop-forwarding change to `Box` in `ui-native`. All new props are correctly typed, resolved through existing helpers where applicable, and covered by tests. No existing behaviour is altered. No issues to flag.**

### What this PR is doing
**Intent:** Add five previously-silently-dropped style prop categories (`position`, `zIndex`, `opacity`, `overflow`, and four individual border-width props) to the `Box` component in `@one-impression/ui-native`, forwarding them through to the underlying React Native `View` style (inferred from diff + body).
**Delivered:** Fully delivered. The diff adds the new props to the `BoxProps` interface, destructures them in the `forwardRef` implementation, resolves the border-width variants through the existing `resolveBorderWidth` helper, and passes all eight new values into the style object. A comprehensive test file covers each new prop individually, the 'no leak when unset' case, and coexistence with existing layout props. A changeset entry documents the additive change at patch level.

No issues found.

**Positive:**
- The 'does not leak the new props onto the underlying View when unset' test (Box.test.tsx L63-L74) is exactly the right regression guard for a prop-forwarding change — it proves undefined values are stripped rather than forwarded as explicit `undefined` style entries.
- Reusing `resolveBorderWidth` for the four individual border-width props is the correct choice: it keeps token resolution consistent with the existing `borderWidth` prop and avoids a parallel resolution path.
- The changeset entry clearly explains the SDUI motivation (absolute-positioned overlays, animated fades, one-sided dividers), making the 'why' traceable without digging into the PR body.

---

PR #170, opened 2026-05-26
fix(sdui-runtime): handleNavigate falls back to action.target for legacy emits

mridulgupta-oi:
## Summary

`handleNavigate` now reads `target` from `payload.target` first (the new, schema-canonical location after [amplify-schemas#172](https://github.com/One-Impression/amplify-schemas/pull/172)), and falls back to `action.target` when the payload field is undefined. The fallback path emits a deprecation warning through `config.logger.warn` (or `console.warn` when no logger is configured), with the op and resolved target as context so stale emits are easy to identify. The new-style flow is unchanged.

## Why

Cached, third-party, and pre-#172 SDUI emits still place `target` at the action level rather than inside the payload. Without a fallback, those navigations silently resolve to an empty string and tap-throughs land on a blank route — a confusing user-facing bug for what's actually a wire-format drift. This is the same gradual-migration resilience pattern as `SduiNode`'s `ZodError` fallback (#151) and the `Interpreter`'s `props→data` normaliser (in this batch): keep the app working, leave a telemetry breadcrumb.

Resolution order:

1. `payload.target` — preferred, never warns
2. `action.target`  — legacy fallback, warns once
3. `""`              — neither field; pop / back-style nav

## Test plan

- [x] 5 new `node:test` cases under `handlers/__tests__/navigate.test.ts`:
  - new-style emit (target in payload) forwarded as-is, no warning
  - legacy emit (`action.target` only) navigates correctly + deprecation warns through the logger
  - both fields present — payload wins, no warning
  - neither field — empty target, no warning (e.g. `op: "back"`)
  - missing logger falls back to `console.warn`
- [x] `npm test -w packages/sdui-runtime` — the new test file hits the same pre-existing peer-dep baseline as `branch.test.ts` / `compound.test.ts` / `set-local.test.ts` / `bff-call.test.ts` (`@one-impression/sdk-native-sdui` unresolved in the workspace, because `navigate.ts` imports `NavigatePayloadSchema` from it). Separate workspace issue, not introduced here.
- [ ] Manual: trigger a legacy navigate emit in the creator-app simulator and confirm the destination loads + the warning fires

### DIFF
--- a/packages/sdui-runtime/src/action-engine/handlers/navigate.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/navigate.ts
@@ -4,6 +4,19 @@ import type { ActionEngineConfig, ActionEngine } from "../types.js";
 
 /**
  * navigate — delegates to config.onNavigate with the parsed op, target, and params.
+ *
+ * Target resolution order (matches the gradual migration off the
+ * legacy emit shape — see amplify-schemas#172):
+ *
+ *  1. `payload.target` — the new, schema-canonical location.
+ *  2. `action.target`  — legacy fallback. Cached / third-party / pre-#172
+ *                        emits put the destination one level up at the
+ *                        action level. When this branch fires, emit a
+ *                        deprecation warning through `config.logger` so
+ *                        the stale source is easy to identify.
+ *
+ * If neither is present we still call `onNavigate` with an empty string
+ * so existing call sites get the same behaviour as before.
  */
 export async function handleNavigate(
   action: Action,
@@ -11,5 +24,33 @@ export async function handleNavigate(
   _engine: ActionEngine,
 ): Promise<void> {
   const payload = NavigatePayloadSchema.parse(action.payload);
-  config.onNavigate(payload.op, payload.target ?? "", payload.params);
+
+  // Probe the action for the legacy action-level `target` without
+  // widening the schema-typed `Action`.
+  const legacyTarget = (action as Action & { target?: unknown }).target;
+  const legacyTargetIsString = typeof legacyTarget === "string";
+
+  let target: string;
+  if (payload.target !== undefined && payload.target !== null) {
+    // New-style emit wins — never log a deprecation when the payload
+    // already provides the canonical field, even if the legacy field
+    // also happens to be set.
+    target = payload.target;
+  } else if (legacyTargetIsString) {
+    target = legacyTarget;
+    const warn =
+      config.logger?.warn ??
+      ((message: string, context?: Record<string, unknown>) => {
+        // eslint-disable-next-line no-console
+        console.warn(message, context);
+      });
+    warn(
+      "[sdui-runtime] navigate action used legacy action.target — move target into payload (schemas#172)",
+      { op: payload.op, target: legacyTarget },
+    );
+  } else {
+    target = "";
+  }
+
+  config.onNavigate(payload.op, target, payload.params);
 }
--- a/.changeset/sdui-navigate-target-fallback.md
+++ b/.changeset/sdui-navigate-target-fallback.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/sdui-runtime/src/action-engine/handlers/navigate.ts
```diff
@@ -4,12 +4,53 @@ import type { ActionEngineConfig, ActionEngine } from "../types.js";
...
+    target = payload.target;
+  } else if (legacyTargetIsString) {
+    target = legacyTarget;
+    const warn =
+      config.logger?.warn ??
+      ((message: string, context?: Record<string, unknown>) => {
+        // eslint-disable-next-line no-console
+        console.warn(message, context);
+      });
+    warn(
```
zenith-oi, line 47: **FLAG** — [pii-logging-discipline] The deprecation warning passes `{ op: payload.op, target: legacyTarget }` as the structured context. `target` is a navigation route string — in practice these are paths like `/creator/earnings` or `/user/12345/profile`. If any route embeds a user ID, username, or other PII segment (common in mobile SDUI emits), that value lands in whatever log sink `config.logger.warn` routes to. The fix is cheap: log a sanitised form or omit the target value from the warning context entirely — the op alone is usually enough to identify the stale emitter, and the full target can be retrieved from the navigation event itself. At minimum, add a note in the JSDoc that callers must not configure a logger that forwards raw route strings to PII-sensitive sinks.

_Cited:_ `pii-logging-discipline — 'logs never contain personally-identifying information'; route strings with embedded IDs are a common PII vector in mobile navigation events`

zenith-oi (approved):
## Zenith Code Review

**Clean, focused resilience patch. The implementation is correct, the test coverage is thorough, and the changeset is appropriately scoped as a patch. One minor concern worth the engineer's attention: the deprecation warning logs the resolved `target` string value, which could be a URL path containing user-identifying segments — worth a quick check against the PII logging discipline.**

### What this PR is doing
**Intent:** Add a legacy-target fallback to `handleNavigate` so that SDUI actions placing `target` at the action level (pre-schemas#172 shape) still navigate correctly, while emitting a deprecation warning through the configured logger (or `console.warn`) to surface stale emitters in telemetry.
**Delivered:** The diff fully delivers the stated intent. The resolution order (payload.target → action.target → empty string) is implemented correctly, the deprecation warning fires only on the legacy branch, the logger fallback to `console.warn` works, and all five test cases cover the documented scenarios. No scope overshoot detected. No linked task page or brief was provided, so acceptance items cannot be checked by number.

**Positive:**
- The `payload.target !== undefined && payload.target !== null` guard correctly handles both `undefined` (field absent from schema parse) and `null` (explicitly nulled), preventing the legacy branch from firing on a healthy new-style emit that happens to have a null target.
- Using `(action as Action & { target?: unknown }).target` with a subsequent `typeof legacyTarget === 'string'` check is the right way to probe an out-of-schema field without widening the typed surface — avoids any `any` cast.
- The `console.warn` fallback is wrapped inline rather than stored as a module-level variable, which keeps the function pure and avoids the singleton anti-pattern.
- All five test cases are meaningfully distinct and cover the exact edge cases called out in the PR body. The `console.warn` monkey-patch test correctly restores the original in a `finally` block.
- Changeset entry is accurate and links the upstream schema PR — good provenance for consumers upgrading the package.

**Read against:**
- `PR body — resolution order: payload.target > action.target > empty string`
- `PR body — deprecation warning via config.logger.warn or console.warn fallback`
- `PR body — test plan: 5 node:test cases`
- `.changeset/sdui-navigate-target-fallback.md — patch bump for @one-impression/sdui-runtime`

---

mridulgupta-oi:
Closing — same strict-contract principle as #164.

After [amplify-schemas#172](https://github.com/One-Impression/amplify-schemas/pull/172) shipped (CR-5), the `navigate`/`dismiss` builders put `target` inside `payload`. The runtime handler reads `payload.target`. Anything still emitting `action.target` at the action-level is by definition wrong — it should fail loudly (silently no-op until SduiFallback catches it) so the emitting code gets migrated, not have its bug papered over at runtime.

Backward-compat fallback would lengthen the migration window indefinitely. Not opening a follow-up.

PR #183, opened 2026-05-26
fix(changesets): only bump peer-dependents when out of range

mridulgupta-oi:
## Summary
Adds `onlyUpdatePeerDependentsWhenOutOfRange: true` to the Changesets config — the same guard `amplify-schemas` already uses.

## Why
`sdui-runtime` and `ui-native` both declare `@one-impression/tokens-creator: ">=3.0.0"` as a **peerDependency**. Changesets' default behavior MAJOR-bumps every peer-dependent on *any* peer-dep version change — even when the new version still satisfies the dependent's range.

Concretely, the pending release showed:
- `tokens-creator`: 3.0.0 → 3.1.0 (minor — correct, a new additive gradient token)
- `sdui-runtime`: 2.2.0 → **3.0.0** (false major — only had patch changesets)
- `ui-native`: 2.0.2 → **3.0.0** (false major — had *no* changeset at all)

3.1.0 still satisfies `>=3.0.0`, so neither consumer is actually broken. Without this guard, **every additive token release would burn a major version across the entire design system** — unsustainable.

With the guard, peer-dependents only bump when the new peer version falls *outside* their declared range. After this lands, re-running the release produces honest bumps:
- `tokens-creator` → 3.1.0 (minor)
- `sdui-runtime` → 2.2.1 (patch, from its own changesets)
- `ui-native` → no bump

## Test plan
- [x] `config.json` is valid JSON
- [ ] Re-run publish workflow after merge → Version PR shows corrected (non-major) bumps

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -10,6 +10,9 @@
   "access": "restricted",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
+  },
   "ignore": [
     "@one-impression/storybook",
     "@one-impression/templates",


zenith-oi (approved):
## Zenith Code Review

**Minimal, correct config change that guards against Changesets' default behaviour of major-bumping every peer-dependent on any peer-dep version change. The PR body explains the concrete problem (false 3.0.0 bumps on `sdui-runtime` and `ui-native`) and the fix is precisely scoped. No issues found.**

### What this PR is doing
**Intent:** Add `onlyUpdatePeerDependentsWhenOutOfRange: true` to the Changesets experimental config to prevent false major-version bumps on peer-dependents (`sdui-runtime`, `ui-native`) when a peer dependency (`tokens-creator`) releases a minor version that still satisfies the declared `>=3.0.0` range (inferred from diff + body).
**Delivered:** Fully delivered. The single-line config addition is exactly what the PR body describes, the JSON is structurally valid (the new block is correctly placed before the existing `ignore` array), and the flag name matches the Changesets experimental API. No overshoot, no scope drift.

No issues found.

**Positive:**
- PR body is exemplary: it names the exact packages affected, shows the before/after version numbers, explains why the default behaviour is wrong for this repo's semver contract, and includes a concrete post-merge verification step in the test plan.

---

zenith-oi (approved):
## Zenith Code Review

**Minimal, correct config-only change that guards against Changesets' default behaviour of major-bumping peer-dependents unnecessarily. The JSON is valid, the option is placed correctly, and the rationale is well-documented in the PR body. No issues found.**

### What this PR is doing
**Intent:** Add `onlyUpdatePeerDependentsWhenOutOfRange: true` to the Changesets experimental config to prevent false major-version bumps on peer-dependent packages when a peer dependency release still satisfies the declared semver range (inferred from diff + body).
**Delivered:** Fully delivered. The single config change at L13–L15 is exactly what the PR body describes. No linked task page or brief to check acceptance items against.

No issues found.

**Positive:**
- PR body clearly explains the concrete problem (false major bumps on sdui-runtime and ui-native), the root cause (Changesets default peer-dep behaviour), and the expected outcome after the fix — exactly the level of context a future engineer needs to understand why this experimental flag is present.

---

PR #189, opened 2026-05-28
fix(sdui-runtime): home BC renderer bugs + optimistic tab-bar active state

mridulgupta-oi:
## Summary

Renderer-level bugs and a UX gap in the home BC found by on-device E2E against `sdk-native-sdui@^2.6.0` (which ships `IconSchema`, flat `TextSchema` `label`, and `EndpointPaths`). Each visible bug was caught by `SduiErrorBoundary` so the page rendered as fallback boxes rather than crashing — but every campaign card showed a fallback box where its title/subtitle should be, and the 3-tab footer rendered as 3 fallback boxes.

After fixing those, the tab-bar still had a UX gap: the active indicator only moved when the BFF's `replace_section` response arrived, so every tab tap had a perceptible lag before the highlight flipped. Added a generic `TabBarActiveContext` to make that visual instant while keeping content server-driven.

### 1. `TabRenderer` — icon crash _(visible)_
The renderer wrapped `v.icon` in `<Interpreter node={v.icon} />`, but `icon` is a flat `IconSchema` (`{ name, size?, color? }`), not an SDUI node — so `Interpreter` called `resolveRenderer(undefined)` and crashed on `.startsWith()`. The footer's 3 tabs rendered as fallback boxes.

**Fix:** render `<DSIcon name={v.icon.name} size={v.icon.size} color={v.icon.color} />`, matching the established pattern in `ChipRenderer`, `PageHeader`, `InfoIconRow`, and 6 other renderers that consume `IconSchema`.

### 2. `TabRenderer` — tap swallowed _(visible)_
`DSTab` is itself a `Pressable`. Passing `on_click` to `SduiNode` wrapped the tab in an outer `Clickable` Pressable, but in RN the deepest Pressable under the touch point wins — so the inner DSTab Pressable swallowed the tap and the outer handler never fired. (`DSTab.types.ts` already documents this expectation: _"taps land on the inner Pressable first and would otherwise be swallowed"_.)

**Fix:** hold the action engine in TabRenderer, build the press handler locally, and pass it directly as `DSTab.onPress`. Intentionally do **not** forward `on_click` to `SduiNode` (so no outer Clickable wraps). `on_load` / `on_view` / `on_dismount` continue to flow through `SduiNode` unchanged.

### 3. `InfoRow` — `status_tag` flat-text bug _(visible)_
The renderer passed the whole `v.status_tag.label` object to `DSTag.label`, but `status_tag.label` is a flat `TextSchema` (`{ text, color?, font_size?, font_weight? }`), not a string. React rendered it as a child and threw _"Objects are not valid as a React child (found: object with keys {text})"_. Every `info_row` on the home Explore feed rendered as a fallback box (no title / subtitle / status tag).

**Fix:** `label={v.status_tag.label.text}` — same pattern title/subtitle/badge already use in this renderer.

### 4. Optimistic tab-bar active state _(UX gap)_
"Which tab is highlighted" is intrinsically a client-UI concern: the user just touched the button, the indicator should follow the finger — not wait for the BFF round-trip that loads the new tab's content. Previously the active indicator was driven only by `data.active_index` from the server, so every tab switch had a perceptible lag.

**Fix:** added a generic `TabBarActiveContext` (exported from `state/`). `TabsFooterRenderer` owns a `useState<string | null>` for the optimistic active tab id and provides it via the context. `TabRenderer` consumes the context and on press calls `setActiveTabId(node.id)` **before** dispatching `on_click`. The effective active index resolves to `local override ?? data.active_index` — the server still seeds selection at mount and can drive it via re-renders, but the local override wins until the next tap.

- **Zero wire cost** (no payload growth, no extra round-trip).
- **Zero gateway changes** — the home tab dispatcher still emits the same 3 `replace_section` actions; it never has to ship the footer to update the indicator.
- **No remount of the footer.**
- The context is generic so future tab-bar shapes (top tabs, segmented controls, etc.) can reuse it without duplicating the pattern.

## Verified on iPhone 15 Pro simulator

- Footer tabs render with correct labels and the active-underline indicator.
- Tapping a tab flips the active indicator **instantly** while the BFF's `on_click` fires asynchronously — gateway log shows `/v1/creator/home/tab` hit immediately after each tap, and the new tab's `replace_section` swaps the page sections.
- Campaign cards render full title / subtitle / status tag (no more fallback boxes).
- Pull-to-refresh re-fires the explore tab fetch.

## Notes

- One pre-existing test failure in `node-registry.test.ts` (tsx can't parse `react-native/index.js`) is on `main` already — not from these changes. Worth a separate cleanup issue if not already tracked.
- Changeset bumps `@one-impression/sdui-runtime` as `patch`. The optimistic-active addition is technically a small feature, but the home-BC bug fixes dominate; the `state/TabBarActiveContext` export is additive and non-breaking.

### DIFF
--- a/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
+++ b/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
@@ -1,20 +1,34 @@
-import React from "react";
+import React, { useContext } from "react";
 import type { Node } from "@one-impression/sdk-native-sdui";
 import { TabComponentSchema } from "@one-impression/sdk-native-sdui";
 import { Tab as DSTab, Icon as DSIcon } from "@one-impression/ui-native";
 import { SduiNode } from "../../sdui-node/index.js";
 import { useActionEngine } from "../../action-engine/useActionEngine.js";
+import { TabBarActiveContext } from "../../state/TabBarActiveContext.js";
 
 export function TabRenderer(node: Node): React.ReactElement {
   const actionEngine = useActionEngine();
+  // Optional parent tab bar (e.g. TabsFooter). When present, taps update
+  // its optimistic active state synchronously so the indicator follows
+  // the finger without waiting for the BFF round-trip; the BFF call then
+  // proceeds and loads the new tab's content. When absent, the Tab still
+  // works standalone — only the `on_click` dispatch happens.
+  const tabBar = useContext(TabBarActiveContext);
   // DSTab is itself a Pressable, so a wrapping <Clickable> in SduiNode would
   // be swallowed by the inner Pressable (taps land on the deepest Pressable
   // first). Dispatch on_click here and pass it directly as DSTab.onPress;
   // intentionally do NOT forward on_click to SduiNode so the outer wrap is
   // skipped. on_load / on_view / on_dismount still go through SduiNode.
-  const onPress = node.on_click
-    ? () => actionEngine.dispatch(node.on_click!)
-    : undefined;
+  const onPress =
+    node.on_click || tabBar
+      ? () => {
+          // 1. Optimistic visual: parent tab bar updates active state
+          //    synchronously (no-op if not inside one).
+          tabBar?.setActiveTabId(node.id);
+          // 2. Server-driven content: fire the BFF action.
+          if (node.on_click) actionEngine.dispatch(node.on_click);
+        }
+      : undefined;
   return (
     <SduiNode
       data={node.data}
--- a/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
+++ b/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
@@ -1,17 +1,25 @@
 import React from "react";
 import type { Node } from "@one-impression/sdk-native-sdui";
 import { TabComponentSchema } from "@one-impression/sdk-native-sdui";
-import { Tab as DSTab } from "@one-impression/ui-native";
+import { Tab as DSTab, Icon as DSIcon } from "@one-impression/ui-native";
 import { SduiNode } from "../../sdui-node/index.js";
-import { Interpreter } from "../../interpreter/index.js";
+import { useActionEngine } from "../../action-engine/useActionEngine.js";
 
 export function TabRenderer(node: Node): React.ReactElement {
+  const actionEngine = useActionEngine();
+  // DSTab is itself a Pressable, so a wrapping <Clickable> in SduiNode would
+  // be swallowed by the inner Pressable (taps land on the deepest Pressable
+  // first). Dispatch on_click here and pass it directly as DSTab.onPress;
+  // intentionally do NOT forward on_click to SduiNode so the outer wrap is
+  // skipped. on_load / on_view / on_dismount still go through SduiNode.
+  const onPress = node.on_click
+    ? () => actionEngine.dispatch(node.on_click!)
+    : undefined;
   return (
     <SduiNode
       data={node.data}
       schema={TabComponentSchema.shape.data}
       id={node.id}
-      on_click={node.on_click}
       on_load={node.on_load}
       on_view={node.on_view}
       on_dismount={node.on_dismount}
@@ -22,7 +30,16 @@ export function TabRenderer(node: Node): React.ReactElement {
         <DSTab
           label={v.label.text}
           active={v.active}
-          icon={v.icon ? <Interpreter node={v.icon} /> : undefined}
+          onPress={onPress}
+          icon={
+            v.icon ? (
+              <DSIcon
+                name={v.icon.name}
+                size={v.icon.size}
+                color={v.icon.color}
+              />
+            ) : undefined
+          }
         />
       )}
     </SduiNode>
--- a/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
+++ b/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useMemo } from "react";
+import React, { useCallback, useContext } from "react";
 import type { Node } from "@one-impression/sdk-native-sdui";
 import { TabComponentSchema } from "@one-impression/sdk-native-sdui";
 import { Tab as DSTab, Icon as DSIcon } from "@one-impression/ui-native";
@@ -14,27 +14,28 @@ export function TabRenderer(node: Node): React.ReactElement {
   // proceeds and loads the new tab's content. When absent, the Tab still
   // works standalone — only the `on_click` dispatch happens.
   const tabBar = useContext(TabBarActiveContext);
+  // Pull out the setter (a React useState setter — referentially stable
+  // across renders) so the callback dependency list doesn't churn when
+  // the parent's `activeTabId` changes. Depending on the whole `tabBar`
+  // context value would cause every sibling Tab's handler to be re-
+  // allocated on every tap; depending only on the setter keeps it
+  // stable for the lifetime of the parent tab bar.
+  const setActiveTabId = tabBar?.setActiveTabId;
   // DSTab is itself a Pressable, so a wrapping <Clickable> in SduiNode would
   // be swallowed by the inner Pressable (taps land on the deepest Pressable
   // first). Dispatch on_click here and pass it directly as DSTab.onPress;
   // intentionally do NOT forward on_click to SduiNode so the outer wrap is
   // skipped. on_load / on_view / on_dismount still go through SduiNode.
-  // Memoize the handler so DSTab receives a stable reference across
-  // re-renders — important for any future React.memo on DSTab and for
-  // keeping the Pressable's internal gesture state from churning. The
-  // handler is `undefined` when there's nothing for it to do, so the
-  // memo returns the value (not a function); useMemo (not useCallback)
-  // is the right hook here.
-  const onPress = useMemo<(() => void) | undefined>(() => {
-    if (!node.on_click && !tabBar) return undefined;
-    return () => {
-      // 1. Optimistic visual: parent tab bar updates active state
-      //    synchronously (no-op if not inside one).
-      tabBar?.setActiveTabId(node.id);
-      // 2. Server-driven content: fire the BFF action.
-      if (node.on_click) actionEngine.dispatch(node.on_click);
-    };
-  }, [node.id, node.on_click, tabBar, actionEngine]);
+  const handlePress = useCallback(() => {
+    // 1. Optimistic visual: parent tab bar updates active state
+    //    synchronously (no-op if not inside one).
+    setActiveTabId?.(node.id);
+    // 2. Server-driven content: fire the BFF action.
+    if (node.on_click) actionEngine.dispatch(node.on_click);
+  }, [node.id, node.on_click, setActiveTabId, actionEngine]);
+  // Only attach a handler when there's actually work to do; otherwise let
+  // DSTab render without an `onPress` so its Pressable doesn't claim taps.
+  const onPress = node.on_click || setActiveTabId ? handlePress : undefined;
   return (
     <SduiNode
       data={node.data}
--- a/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
+++ b/packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from "react";
+import React, { useContext, useMemo } from "react";
 import type { Node } from "@one-impression/sdk-native-sdui";
 import { TabComponentSchema } from "@one-impression/sdk-native-sdui";
 import { Tab as DSTab, Icon as DSIcon } from "@one-impression/ui-native";
@@ -19,16 +19,22 @@ export function TabRenderer(node: Node): React.ReactElement {
   // first). Dispatch on_click here and pass it directly as DSTab.onPress;
   // intentionally do NOT forward on_click to SduiNode so the outer wrap is
   // skipped. on_load / on_view / on_dismount still go through SduiNode.
-  const onPress =
-    node.on_click || tabBar
-      ? () => {
-          // 1. Optimistic visual: parent tab bar updates active state
-          //    synchronously (no-op if not inside one).
-          tabBar?.setActiveTabId(node.id);
-          // 2. Server-driven content: fire the BFF action.
-          if (node.on_click) actionEngine.dispatch(node.on_click);
-        }
-      : undefined;
+  // Memoize the handler so DSTab receives a stable reference across
+  // re-renders — important for any future React.memo on DSTab and for
+  // keeping the Pressable's internal gesture state from churning. The
+  // handler is `undefined` when there's nothing for it to do, so the
+  // memo returns the value (not a function); useMemo (not useCallback)
+  // is the right hook here.
+  const onPress = useMemo<(() => void) | undefined>(() => {
+    if (!node.on_click && !tabBar) return undefined;
+    return () => {
+      // 1. Optimistic visual: parent tab bar updates active state
+      //    synchronously (no-op if not inside one).
+      tabBar?.setActiveTabId(node.id);
+      // 2. Server-driven content: fire the BFF action.
+      if (node.on_click) actionEngine.dispatch(node.on_click);
+    };
+  }, [node.id, node.on_click, tabBar, actionEngine]);
   return (
     <SduiNode
       data={node.data}
--- a/.changeset/sdui-runtime-home-bc-renderer-bugs.md
+++ b/.changeset/sdui-runtime-home-bc-renderer-bugs.md
@@ -2,9 +2,9 @@
 "@one-impression/sdui-runtime": patch
 ---
 
-fix(sdui-runtime): home BC renderer bugs — Tab icon + tap dispatch, InfoRow flat status_tag label
+fix(sdui-runtime): home BC renderer fixes — Tab icon + tap dispatch, InfoRow flat status_tag label, optimistic tab-bar active state
 
-Two renderer-level bugs found by on-device E2E of the home BC against `sdk-native-sdui@^2.6.0`. Each broke a visible piece of the home page; both are caught by `SduiErrorBoundary` so they render as fallback boxes rather than crash the screen.
+Renderer-level bugs and a UX gap found by on-device E2E of the home BC against `sdk-native-sdui@^2.6.0`. Visible bugs are caught by `SduiErrorBoundary` so they render as fallback boxes rather than crash the screen.
 
 - **`TabRenderer` — icon crash.** The renderer wrapped `v.icon` in `<Interpreter node={v.icon} />`, but `icon` is a flat `IconSchema` (`{ name, size?, color? }`), not an SDUI node — so `Interpreter` called `resolveRenderer(undefined)` and crashed on `.startsWith()`. The footer's 3 tabs rendered as fallback boxes. Fix: render `<DSIcon name={v.icon.name} size={v.icon.size} color={v.icon.color} />`, matching the established pattern in `ChipRenderer`, `PageHeader`, `InfoIconRow`, and 6 other renderers that consume `IconSchema`.
 
@@ -12,4 +12,6 @@ Two renderer-level bugs found by on-device E2E of the home BC against `sdk-nativ
 
 - **`InfoRow` — status_tag label flat-text bug.** The renderer passed the whole `v.status_tag.label` object to `DSTag.label`, but `status_tag.label` is a flat `TextSchema` (`{ text, color?, font_size?, font_weight? }`), not a string. React rendered it as a child and threw "Objects are not valid as a React child (found: object with keys {text})." Every `info_row` on the home Explore feed rendered as a fallback box (no title/subtitle/status tag). Fix: `label={v.status_tag.label.text}`, same pattern title/subtitle/badge already use in this renderer.
 
-Verified on iPhone 15 Pro simulator: footer tabs render with correct labels and the active-underline indicator, tapping a tab fires the BFF's `on_click` and the gateway's `replace_section` action is applied, and campaign cards render full title / subtitle / status tag.
+- **Tab-bar optimistic active state.** "Which tab is highlighted" is intrinsically a client-UI concern: the user just touched the button, the indicator should follow the finger — not wait for the BFF round-trip that loads the new tab's content. Previously the active indicator was driven only by `data.active_index` from the server, which meant every tab switch had a perceptible lag (the highlight only moved once the new tab's `replace_section` response arrived). Fix: added a generic `TabBarActiveContext` (exported from `state/`) that a parent tab-bar renderer provides and child Tabs consume. `TabsFooterRenderer` now owns a `useState<string | null>` for the optimistic active tab id and provides it; `TabRenderer` updates it on press *before* dispatching `on_click`. The effective active index is `local override ?? data.active_index` — the server still seeds selection at mount and can drive it via re-renders, but the local override wins until the next tap. Zero wire cost, zero gateway changes, no remount of the footer. Reusable for future tab bars (top tabs, segmented controls, etc.) since the context is generic.
+
+Verified on iPhone 15 Pro simulator: footer tabs render with correct labels, tapping a tab flips the active indicator **instantly** (no round-trip wait) while the BFF's `on_click` fires asynchronously and the gateway's `replace_section` action swaps the page sections, and campaign cards render full title / subtitle / status tag.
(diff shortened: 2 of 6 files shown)


packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
```diff
@@ -1,17 +1,39 @@
...
+  // the finger without waiting for the BFF round-trip; the BFF call then
+  // proceeds and loads the new tab's content. When absent, the Tab still
+  // works standalone — only the `on_click` dispatch happens.
+  const tabBar = useContext(TabBarActiveContext);
+  // DSTab is itself a Pressable, so a wrapping <Clickable> in SduiNode would
+  // be swallowed by the inner Pressable (taps land on the deepest Pressable
+  // first). Dispatch on_click here and pass it directly as DSTab.onPress;
+  // intentionally do NOT forward on_click to SduiNode so the outer wrap is
+  // skipped. on_load / on_view / on_dismount still go through SduiNode.
+  const onPress =
```
zenith-oi: **FLAG** — [unnecessary-render-allocation] The `onPress` handler is created on every render whenever `tabBar` is non-null, even if `node.on_click` is undefined. The condition `node.on_click || tabBar` means any Tab inside a TabsFooter always allocates a new function object per render. This is harmless for correctness but means DSTab will always receive a new `onPress` reference, defeating any `React.memo` optimisation on DSTab. Consider `useMemo` or `useCallback` wrapping the handler, or restructuring so the function is only created when either condition is actually true (which the current ternary does — but the closure over `tabBar` and `node` means it's always a new reference when tabBar is present). Low priority given the small number of tabs, but worth noting for future tab-heavy layouts.

zenith-oi (dismissed):
## Zenith Code Review

**Three visible renderer crashes and one UX gap in the home BC are fixed cleanly. The code is correct, well-commented, and follows established patterns in the codebase. One minor concern worth flagging: the onPress handler is constructed unconditionally on every render when tabBar context is present but on_click is absent, which is a very minor inefficiency but not a correctness issue. No blocking issues found.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs in the home BC (TabRenderer icon crash, TabRenderer tap-swallowed, InfoRow status_tag flat-text crash) and add an optimistic tab-bar active state via a new TabBarActiveContext to eliminate the perceptible highlight lag on tab switches (inferred from diff + body).
**Delivered:** The diff fully delivers all four stated fixes. The InfoRow label fix (L64) is a one-line surgical correction. The TabRenderer icon fix (L48-L56) matches the established DSIcon pattern. The tap-dispatch fix (L22-L31, L36) correctly bypasses SduiNode's outer Clickable while preserving on_load/on_view/on_dismount. The TabBarActiveContext (new file, 51 lines) is well-documented and correctly typed. TabsFooterRenderer wires the context provider with useMemo-stabilised value and resolves effective active index correctly. No linked task page is present so no acceptance items to check against.

**Positive:**
- The TabBarActiveContext file is exceptionally well-documented — the JSDoc block explains the semantics, the null-default fallback for standalone Tab usage, and the server-override behaviour clearly. Future engineers will understand the design intent without reading the PR.
- Using `useMemo` on `ctxValue` in TabsFooterRenderer (L62-L65) is exactly right — it prevents the context value from being a new object reference on every render, which would otherwise cause all Tab consumers to re-render on every TabsFooter render cycle.
- Intentionally not forwarding `on_click` to SduiNode (with an explicit comment explaining why) is the correct and principled fix for the nested-Pressable tap-swallowing problem. The comment at L17-L21 makes the reasoning durable.
- The effective active index resolution (local override ?? server-supplied) is clean and correctly handles the -1 sentinel from findIndex when no local tap has occurred yet.
- Changeset entry is thorough and accurately describes each fix with root-cause analysis — exactly the right level of detail for a patch release.

---

zenith-oi (dismissed):
## Zenith Code Review

**Three genuine renderer bugs (icon crash, swallowed tap, object-as-React-child) are fixed correctly, and a clean optimistic-active-state context is added. The implementation is well-reasoned and the changeset description is thorough. No issues found.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs in the home BC (TabRenderer icon crash, TabRenderer tap swallowed, InfoRow status_tag label object-not-string) and add a generic TabBarActiveContext to give the tab-bar optimistic active-state so the indicator follows the finger without waiting for a BFF round-trip.
**Delivered:** The diff fully delivers all four stated fixes. (1) Icon crash: `<Interpreter node={v.icon} />` replaced with `<DSIcon name={v.icon.name} size={v.icon.size} color={v.icon.color} />` — correct. (2) Tap swallowed: `on_click` removed from SduiNode props, dispatched locally via `useActionEngine` and passed as `DSTab.onPress` — correct. (3) InfoRow label: `v.status_tag.label` → `v.status_tag.label.text` — correct. (4) Optimistic active state: `TabBarActiveContext` created, provided by `TabsFooterRenderer`, consumed by `TabRenderer` — correct. No acceptance items from a linked task page are present; intent is inferred from the PR body and changeset, and the diff matches the stated intent exactly.

No issues found.

**Positive:**
- The `onPress` construction at Tab.renderer.tsx L22-31 is precise: it only creates a handler when there is actually something to do (`node.on_click || tabBar`), avoiding an unnecessary re-render-triggering function reference when the tab is purely decorative.
- Using `useMemo` for the context value in TabsFooterRenderer (L62-65) is the correct pattern — it prevents every child Tab from re-rendering on unrelated parent state changes.
- The `TabBarActiveContext` default of `null` (not an empty object) means child Tabs outside a tab-bar provider get a clean `null` signal and fall back to BFF-supplied `data.active` with zero behaviour change — a well-considered fallback.
- The changeset description is exceptionally detailed: it names the root cause, the crash site, the established pattern being followed, and the on-device verification steps — exactly the level of context future engineers need when bisecting regressions.

---

packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
```diff
@@ -1,17 +1,45 @@
...
+  // first). Dispatch on_click here and pass it directly as DSTab.onPress;
+  // intentionally do NOT forward on_click to SduiNode so the outer wrap is
+  // skipped. on_load / on_view / on_dismount still go through SduiNode.
+  // Memoize the handler so DSTab receives a stable reference across
+  // re-renders — important for any future React.memo on DSTab and for
+  // keeping the Pressable's internal gesture state from churning. The
+  // handler is `undefined` when there's nothing for it to do, so the
+  // memo returns the value (not a function); useMemo (not useCallback)
+  // is the right hook here.
+  const onPress = useMemo<(() => void) | undefined>(() => {
```
zenith-oi: **FLAG** — [unnecessary-render-allocation] The previous round flagged that the `onPress` handler is allocated on every render when `tabBar` is non-null even if `node.on_click` is undefined. The diff at L28-L37 shows the guard `if (!node.on_click && !tabBar) return undefined` — but when `tabBar` is non-null and `node.on_click` is undefined, the memo still returns a new closure on every render (the early-return only fires when BOTH are falsy). The `useMemo` dependencies include `tabBar` (the context object), which is a new object reference on every `TabsFooterRenderer` render because `ctxValue` is memoized on `localActiveTabId` but the context object itself changes whenever `localActiveTabId` changes. This means every tab-bar re-render (e.g. on any tab tap) allocates a new `onPress` closure in every sibling Tab. For a 3-tab footer this is negligible, but the pattern is worth noting: consider `useCallback` with `[node.id, node.on_click, tabBar?.setActiveTabId]` — `setActiveTabId` is the stable `useState` setter (referentially stable across renders), so the dependency would only change when `node.id` or `node.on_click` changes, not on every active-tab flip.

zenith-oi (dismissed):
## Zenith Code Review

**Four targeted bug fixes and one UX improvement for the home BC SDUI renderers. The code is clean, well-commented, and the optimistic tab-bar pattern is correctly implemented with a stable useMemo dependency array. One previous FLAG (unnecessary-render-allocation) remains unaddressed.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs (Tab icon crash, Tab tap swallowed, InfoRow status_tag label object-not-string) found during on-device E2E of the home BC, and add a generic TabBarActiveContext to make the tab-bar active indicator update optimistically on tap rather than waiting for the BFF round-trip.
**Delivered:** All four stated fixes are present and correctly implemented. The InfoRow fix (L64) extracts `.text` from the TextSchema object. The Tab icon fix (L54-L62) renders DSIcon directly instead of wrapping in Interpreter. The tap-swallowed fix (L42-L37) dispatches on_click locally via useActionEngine and passes the handler as DSTab.onPress without forwarding on_click to SduiNode. The optimistic active-state fix introduces TabBarActiveContext (new file), wires it in TabsFooterRenderer (L61-L100), and consumes it in TabRenderer (L16, L33). All four fixes are coherent and the implementation matches the stated design.

**Positive:**
- The TabBarActiveContext design is clean and reusable — null-default context means standalone Tab components degrade gracefully with zero behaviour change, which is exactly the right defensive pattern.
- Intentionally not forwarding on_click to SduiNode is well-documented in both the changeset and inline comments, making the non-obvious omission easy to understand for future reviewers.
- useMemo on ctxValue in TabsFooterRenderer correctly prevents the context value from being a new object on every render unless localActiveTabId actually changes — good React discipline.
- The changeset entry is exceptionally thorough, explaining the root cause, the fix, and the verification steps for each of the four changes.

---

packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
zenith-oi: **FLAG** — [unnecessary-render-allocation] The `useMemo` guard `if (!node.on_click && !tabBar) return undefined` only short-circuits when BOTH are absent. When `tabBar` is non-null (i.e. the Tab is inside a TabsFooter — the common case after this PR), a new closure is allocated on every render even if `node.on_click` is undefined and `node.id` has not changed. The memo deps `[node.id, node.on_click, tabBar, actionEngine]` are correct for correctness, but the early-return condition should be `if (!node.on_click && !tabBar) return undefined` — which it already is — so the real issue is that `tabBar` being non-null forces a closure even when there is nothing to dispatch. Consider splitting: if `node.on_click` is undefined and `tabBar` is non-null, the handler only calls `tabBar.setActiveTabId(node.id)` — that closure is stable as long as `node.id` and `tabBar` are stable, which they typically are across re-renders. The current shape is not wrong, but it allocates a new function object on every render of any Tab inside a TabsFooter, which was the original complaint and remains structurally unchanged.

zenith-oi (dismissed):
## Zenith Code Review

**Three visible renderer crashes and one UX lag on the home BC are fixed cleanly. The optimistic tab-bar context is well-designed (generic, null-safe fallback, stable memo reference). One previous-round FLAG about unnecessary handler allocation on every render remains unaddressed — the `useMemo` guard condition still creates a new closure on every render when `tabBar` is non-null, regardless of whether `node.on_click` changed.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs in the home BC (TabRenderer icon crash, TabRenderer tap-swallowed, InfoRow status_tag flat-text crash) and add an optimistic tab-bar active state via a new TabBarActiveContext so the active indicator follows the user's finger without waiting for a BFF round-trip.
**Delivered:** The diff fully delivers all four stated fixes. (1) InfoRow: `v.status_tag.label.text` correctly extracts the string from the flat TextSchema object. (2) TabRenderer icon: `<DSIcon name={v.icon.name} .../>` replaces the broken `<Interpreter node={v.icon} />` call. (3) TabRenderer tap: `on_click` is no longer forwarded to SduiNode; instead it is dispatched locally via `useActionEngine` and passed as `DSTab.onPress`, eliminating the double-Pressable swallow. (4) Optimistic active state: `TabBarActiveContext` is created, exported from `state/index.ts`, provided by `TabsFooterRenderer`, and consumed by `TabRenderer` — the effective active index resolves local override first, then falls back to `data.active_index`. All four fixes match the PR body's stated intent. No scope overshoot detected.

**Positive:**
- The TabBarActiveContext design is clean: generic interface, null default so standalone Tabs degrade gracefully, `useMemo` on the context value to avoid unnecessary Provider re-renders in TabsFooterRenderer.
- Correctly not forwarding `on_click` to SduiNode while still threading `on_load`/`on_view`/`on_dismount` — the comment explaining the double-Pressable RN behaviour is exactly the right level of documentation for a non-obvious platform constraint.
- The effective-active-index resolution (`localIndex >= 0 ? localIndex : v.active_index`) is a clean, minimal override pattern — server still seeds on mount, local wins on tap, no remount required.
- Changeset entry is thorough and accurately describes all four fixes with root-cause analysis, matching the established pattern in other renderers cited as reference.

---

zenith-oi (dismissed):
## Zenith Code Review

**Three real crash-level renderer bugs are fixed correctly and the optimistic active-state UX gap is closed with a clean, reusable React context. The code is well-commented, the changeset entry is thorough, and the pattern matches established conventions in the codebase. No issues found.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs (Tab icon crash, Tab tap swallowed, InfoRow status_tag label object-not-string) found during on-device E2E of the home BC, and add a generic TabBarActiveContext to make the tab-bar active indicator update optimistically on tap rather than waiting for the BFF round-trip.
**Delivered:** Fully delivered. All four stated fixes are present in the diff: (1) Tab icon now renders via DSIcon directly instead of Interpreter [L54-L62 Tab.renderer.tsx]; (2) on_click is dispatched locally via useActionEngine and passed as DSTab.onPress, not forwarded to SduiNode [L28-L37, L42 Tab.renderer.tsx]; (3) InfoRow status_tag label now uses .text [L64 InfoRow.renderer.tsx]; (4) TabBarActiveContext is introduced [TabBarActiveContext.ts], provided by TabsFooterRenderer [L61-L100 TabsFooter.renderer.tsx], and consumed by TabRenderer [L16, L33 Tab.renderer.tsx]. No linked task page or brief was provided, so acceptance items are inferred from the PR body — all stated items are addressed.

No issues found.

**Positive:**
- The useMemo on onPress correctly returns undefined (not a no-op function) when there is nothing to do, avoiding unnecessary Pressable re-renders and keeping DSTab's gesture state stable.
- TabBarActiveContext defaults to null so Tab renderers outside a tab-bar context degrade gracefully to BFF-driven active state — zero behaviour change for standalone Tab usage.
- The effective-active-index logic (local override ?? server active_index) is clean and the comment explaining the semantics is accurate and concise.
- Changeset entry is unusually thorough — each bug is explained with root cause, symptom, and fix, which will be valuable for consumers upgrading the package.

**Read against:**
- `PR body — fix 1: TabRenderer icon crash (Interpreter on flat IconSchema)`
- `PR body — fix 2: TabRenderer tap swallowed (nested Pressable, on_click forwarded to SduiNode)`
- `PR body — fix 3: InfoRow status_tag label object passed to DSTag.label`
- `PR body — fix 4: optimistic tab-bar active state via TabBarActiveContext`

---

packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
```diff
@@ -1,17 +1,46 @@
...
+  // context value would cause every sibling Tab's handler to be re-
+  // allocated on every tap; depending only on the setter keeps it
+  // stable for the lifetime of the parent tab bar.
+  const setActiveTabId = tabBar?.setActiveTabId;
+  // DSTab is itself a Pressable, so a wrapping <Clickable> in SduiNode would
+  // be swallowed by the inner Pressable (taps land on the deepest Pressable
+  // first). Dispatch on_click here and pass it directly as DSTab.onPress;
+  // intentionally do NOT forward on_click to SduiNode so the outer wrap is
+  // skipped. on_load / on_view / on_dismount still go through SduiNode.
+  const handlePress = useCallback(() => {
```
zenith-oi, line 29: **FLAG** — [unnecessary-render-allocation] The `handlePress` callback (and therefore `onPress`) is created whenever `setActiveTabId` is truthy, even if `node.on_click` is absent. In a tab bar with N tabs, every tap causes the parent to re-render all N siblings; each sibling re-runs `useCallback` and the guard on L38 (`node.on_click || setActiveTabId ? handlePress : undefined`) still attaches a handler. The previous round flagged this as the `useMemo` guard only short-circuiting when BOTH are absent — the same structural condition is still present. The fix is to split the dependency list: if `node.on_click` is falsy, the callback body reduces to `setActiveTabId?.(node.id)` which can be inlined or memoised separately, avoiding the allocation when there is no BFF action to dispatch. This is a minor perf concern on large tab bars, not a correctness bug.

zenith-oi (approved):
## Zenith Code Review

**Three visible renderer crashes and one UX lag are fixed cleanly. The code is well-reasoned, well-commented, and the optimistic-state pattern is correctly scoped (context is null-safe, setter is stable, useMemo guards the context value). One pre-existing FLAG from the previous round remains technically unresolved — the `handlePress` callback is allocated even when only `setActiveTabId` is present and `on_click` is absent — but this is a minor allocation concern, not a correctness issue.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs (TabRenderer icon crash, TabRenderer tap swallowed, InfoRow status_tag label object-not-string) found during on-device E2E of the home BC, and add a generic TabBarActiveContext to make the tab-bar active indicator update optimistically on tap rather than waiting for the BFF round-trip.
**Delivered:** All four stated fixes are present and correct in the diff. (1) Icon crash: `<Interpreter node={v.icon} />` replaced with `<DSIcon name={v.icon.name} size={v.icon.size} color={v.icon.color} />` at Tab.renderer.tsx L55-63. (2) Tap swallowed: `on_click` removed from SduiNode props, dispatched locally via `useActionEngine` and passed directly as `DSTab.onPress` at Tab.renderer.tsx L29-38. (3) InfoRow label: `v.status_tag.label` → `v.status_tag.label.text` at InfoRow.renderer.tsx L64. (4) Optimistic active state: `TabBarActiveContext` added, provided by `TabsFooterRenderer`, consumed by `TabRenderer`. All four are fully delivered.

> ℹ️ 2 previous Zenith reviews dismissed. Inline comments from those reviews are superseded — only comments from **this review** are current.

**Positive:**
- The three bug fixes are minimal and surgical — each touches exactly the line that was wrong and matches the established pattern used elsewhere in the codebase (DSIcon, .text extraction).
- TabBarActiveContext is well-designed: null default means standalone Tabs degrade gracefully with zero behaviour change, the setter is pulled out of the context value before the useCallback dependency list to keep sibling handlers stable across taps, and useMemo on the context value prevents unnecessary re-renders of all consumers on unrelated parent renders.
- The changeset entry is unusually thorough — it documents the root cause, the symptom (fallback boxes vs crash), the fix, and the established pattern each fix follows. This is exactly the right level of detail for a renderer bug fix.

---

packages/sdui-runtime/src/ui_components/Tab/Tab.renderer.tsx
zenith-oi, line 29: **FLAG** — [unnecessary-render-allocation] The `handlePress` callback is created unconditionally by `useCallback` on every render, regardless of whether `node.on_click` or `setActiveTabId` are present. The guard at L38 (`const onPress = node.on_click || setActiveTabId ? handlePress : undefined`) only controls whether the already-allocated callback is *assigned* to `onPress` — it does not prevent the allocation. When both are absent (a standalone Tab with no click handler and no parent tab bar), `handlePress` is still allocated and then immediately discarded. The fix is to move the condition into the `useCallback` itself or wrap the whole hook in a conditional: `const handlePress = useCallback(node.on_click || setActiveTabId ? () => { ... } : undefined, [...])`, or restructure so the callback is only created when there is work to do. This is the same structural issue flagged in the previous review round — the `onPress` guard does not address the allocation.

zenith-oi (approved):
## Zenith Code Review

**Three visible renderer bugs (icon crash, swallowed tap, object-as-label) are correctly fixed, and the optimistic tab-bar active state is cleanly implemented via a new React context. The previous-round FLAG about unnecessary render allocation on `handlePress` remains structurally unaddressed — the `onPress` guard at L38 still allocates `handlePress` unconditionally and only conditionally assigns it, which is the same shape the prior review flagged.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs (TabRenderer icon crash, TabRenderer tap swallowed, InfoRow status_tag label object-not-string) found during on-device E2E of the home BC, and add a generic TabBarActiveContext to make the tab-bar active indicator update optimistically on tap rather than waiting for the BFF round-trip (inferred from diff + body).
**Delivered:** All four stated fixes are present and correctly implemented. The InfoRow fix (L64) is a one-line change from the object to `.text`. The TabRenderer icon fix replaces `<Interpreter node={v.icon} />` with `<DSIcon name/size/color>` (L55-63). The tap-swallowed fix removes `on_click` from SduiNode and dispatches it locally via `useActionEngine` with `handlePress` passed to `DSTab.onPress` (L29-38, L43-44). The TabBarActiveContext is a clean new file (L1-51 of TabBarActiveContext.ts) with correct `useMemo` memoisation in TabsFooterRenderer and `useContext` consumption in TabRenderer. The changeset file is present. No scope overshoot detected.

**Positive:**
- The `setActiveTabId` extraction from the context value (L23) to stabilise the `useCallback` dependency list is a genuinely thoughtful optimisation — depending only on the setter (which is referentially stable as a React useState setter) rather than the whole context value prevents every sibling Tab's handler from being reallocated on every tap. The comment explaining why is clear and correct.
- The `useMemo` wrapping of the context value in TabsFooterRenderer (L62-65) is correct — without it, every render of the footer would produce a new context object reference and force all consuming Tabs to re-render.
- The fallback semantics (`localActiveTabId ?? data.active_index`) are well-designed: the server seeds selection at mount, the local override wins on first tap, and the server can still drive re-renders. Zero wire cost, no gateway changes.
- The changeset entry is detailed and accurate — it explains each bug, the root cause, and the fix pattern, which is exactly what a patch-level changeset should contain.

---

zenith-oi (approved):
## Zenith Code Review

**Four targeted bug fixes and a UX improvement for the home BC SDUI renderers. The code is clean, the rationale is well-documented in both the changeset and inline comments, and the design of TabBarActiveContext (null default, setter-only dependency in useCallback) is thoughtful. No issues worth blocking or flagging.**

### What this PR is doing
**Intent:** Fix three renderer-level bugs in the home BC (TabRenderer icon crash, TabRenderer tap swallowed, InfoRow status_tag label object-not-string) and add a generic TabBarActiveContext to make the tab-bar active indicator update optimistically on tap rather than waiting for the BFF round-trip.
**Delivered:** The diff fully delivers all four stated fixes. The InfoRow fix (L64) is a one-line surgical correction. The TabRenderer icon fix (L55-L63) correctly replaces the Interpreter wrapping with direct DSIcon rendering. The tap-swallowing fix (L29-L38, L43-L44) correctly bypasses SduiNode's outer Clickable by dispatching on_click locally and passing the handler directly to DSTab.onPress. The TabBarActiveContext (new file, L1-L51) is well-designed with a null default so standalone Tabs degrade gracefully. The optimistic active state in TabsFooterRenderer (L61-L100) correctly computes effectiveActiveIndex as local-override-wins-over-server. No acceptance items are declared in a linked task page; the PR body's four stated fixes are all present in the diff.

No issues found.

**Positive:**
- The useCallback dependency list in TabRenderer (L35) depends only on setActiveTabId (the stable setter) rather than the full tabBar context value — this correctly avoids re-allocating the handler on every sibling tap, and the inline comment explains exactly why.
- TabBarActiveContext defaults to null so Tabs rendered outside a tab bar fall back to the BFF-supplied data.active flag with zero behaviour change — good defensive design.
- The changeset entry is unusually thorough: it names the root cause, the crash site, the fix, and the established pattern being followed for each of the three visible bugs. This is the right level of detail for a renderer bug fix that would otherwise be hard to audit from the diff alone.

**Read against:**
- `PR body — fix 1: TabRenderer icon crash`
- `PR body — fix 2: TabRenderer tap swallowed`
- `PR body — fix 3: InfoRow status_tag flat-text bug`
- `PR body — fix 4: optimistic tab-bar active state`

---

mridulgupta-oi:
Re the two `unnecessary-render-allocation` flags on Tab.renderer.tsx:

acknowledged but accepting the current shape:

1. **`useCallback` with stable deps returns a stable reference** — `setActiveTabId` is a React useState setter (referentially stable for the lifetime of the parent tab bar), `actionEngine` is from `useActionEngine` (stable for the lifetime of the runtime provider), and `node.id` / `node.on_click` are stable for a given node. So the per-render allocation is the one-time hook bookkeeping cost — the returned function reference does NOT change between renders. DSTab therefore receives the same `onPress` reference across all sibling re-renders during a tab tap.

2. **The remaining concern** is the pathological standalone case — a Tab rendered with neither `on_click` nor a parent tab bar. In that case `useCallback` still allocates a no-op closure that's immediately discarded. The cost is one closure per Tab per render. For the home BC's 3-tab footer that's 3 closures × 1 render-on-tap — sub-microsecond on Hermes, dominated by the surrounding view rebuild.

Restructuring to avoid the standalone-case allocation would require conditional hook calls (violates rules of hooks) or splitting into two components. Both are heavier than the cost they save. Documenting here so a future engineer doesn't repeat the optimisation cycle.

cc @zenith-oi

PR #207, opened 2026-06-03
feat(ui): PullToRefreshRing — pull-to-refresh animation for Oportunities

apaksh-gupta:
## Summary

- Adds `PullToRefreshRing` component — a 3-state animated SVG ring for pull-to-refresh interactions on scrollable surfaces in the Oportunities creator app
- States: **idle** (hidden), **pulling** (arc fills proportional to progress 0-1), **refreshing** (continuous spin with 75% arc), **done** (green checkmark scale-in → hold → fade-out)
- Uses singleton keyframe injection pattern (matching GentlePausePill/SignalDot), consumes `--amp-oportunities-*` tokens, respects `prefers-reduced-motion`

## Test plan

- [ ] `npx tsc --noEmit` passes (verified: zero errors in new component)
- [ ] Visually verify idle state renders nothing
- [ ] Verify pulling state fills arc proportional to progress prop (0→1)
- [ ] Verify refreshing state shows continuous spin with partial arc
- [ ] Verify done state shows green checkmark with scale-in → hold → fade-out sequence
- [ ] Verify `prefers-reduced-motion` disables spin animation
- [ ] Verify component exports correctly from `@one-impression/ui`

### DIFF
--- a/packages/ui/src/components/PullToRefreshRing/PullToRefreshRing.tsx
+++ b/packages/ui/src/components/PullToRefreshRing/PullToRefreshRing.tsx
@@ -1,0 +1,231 @@
+'use client';
+
+import React from 'react';
+import { cn } from '../../lib/cn';
+
+/**
+ * PullToRefreshRing — 3-state animated ring for pull-to-refresh interactions
+ * on scrollable surfaces in the Oportunities app.
+ *
+ * States:
+ *   idle      — hidden
+ *   pulling   — arc fills clockwise from 12 o'clock, proportional to `progress`
+ *   refreshing — full ring spins continuously (~75% arc visible)
+ *   done      — green checkmark scales in, holds, then fades out
+ *
+ * Tokens consumed:
+ *   --amp-oportunities-accent              (ring stroke)
+ *   --amp-oportunities-border-soft         (track stroke)
+ *   --amp-oportunities-status-success      (done checkmark bg)
+ *   --amp-oportunities-motion-refresh-spin (spin duration)
+ *   --amp-oportunities-motion-refresh-done-display (done hold time)
+ *   --amp-oportunities-motion-refresh-done-fade    (done fade duration)
+ */
+
+export type PullToRefreshState = 'idle' | 'pulling' | 'refreshing' | 'done';
+
+export interface PullToRefreshRingProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> {
+  /** Current state of the pull-to-refresh interaction. */
+  state: PullToRefreshState;
+  /** Progress 0–1, used in 'pulling' state to fill the ring arc. */
+  progress?: number;
+  /** Pixel diameter of the ring. Default 32. */
+  size?: number;
+  /** Stroke width of the ring. Default 3. */
+  strokeWidth?: number;
+  className?: string;
+}
+
+/* ---------- Singleton keyframe injection ---------- */
+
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/PullToRefreshRing/PullToRefreshRing.tsx
```diff
@@ -0,0 +1,231 @@
...
+    if (state !== 'done') {
+      setPhase('hidden');
+      return;
+    }
+
+    // enter → hold → fade → hidden
+    setPhase('enter');
+
+    // After scale-in (300ms), move to hold
+    const t1 = window.setTimeout(() => setPhase('hold'), 300);
```
zenith-oi, line 88: **FLAG** — [no-hardcoded-business-values] The done-state animation timing values — 300ms (scale-in), 600ms (hold), 200ms (fade) — are inlined as raw integer literals in three `window.setTimeout` calls and in the `animation` string at L150/L152. These values encode the animation contract for the component and are referenced in the JSDoc token list (`--amp-oportunities-motion-refresh-done-display`, `--amp-oportunities-motion-refresh-done-fade`). If the design tokens change, the JS timers and the CSS animation durations will silently diverge. Extract them as named constants at the top of the file (e.g. `const DONE_ENTER_MS = 300`, `const DONE_HOLD_MS = 600`, `const DONE_FADE_MS = 200`) so the relationship between the JS timer and the CSS variable fallback is explicit and co-located. Also applies to the `0.9s` spin duration fallback at L193 and the `0.1s` dashoffset transition at L222.

zenith-oi (approved):
## Zenith Code Review

**Clean, well-structured new component that delivers exactly what the PR body describes. The implementation is consistent with existing Oportunities component patterns (singleton keyframe injection, CSS token consumption, `prefers-reduced-motion` support). One FLAG: the done-state timing values (300ms, 600ms, 200ms) are hardcoded inline in `useDonePhase` rather than extracted to named constants, which makes the animation contract invisible and fragile to drift. The `useInjectedKeyframes` hook is a module-level side-effect pattern that technically conflicts with the `avoid-singletons` rule, but this is an established pattern in this codebase (per the PR body citing GentlePausePill/SignalDot precedent) so it is not flagged.**

### What this PR is doing
**Intent:** Add a new `PullToRefreshRing` UI component — a 3-state animated SVG ring for pull-to-refresh interactions in the Oportunities creator app — and export it from the `@one-impression/ui` package (inferred from diff + body; no linked task page).
**Delivered:** The diff fully delivers the stated intent: the component is implemented with all four states (idle/pulling/refreshing/done), consumes `--amp-oportunities-*` CSS tokens, respects `prefers-reduced-motion`, uses the singleton keyframe injection pattern consistent with other Oportunities components, and is correctly exported from the package index. No linked task page means there are no numbered acceptance items to check against. One genuine issue is worth flagging (hardcoded timing literals that should be constants), and one minor concern about the `avoid-singletons` pattern is worth noting in the summary.

**Positive:**
- `prefers-reduced-motion` is handled correctly via a CSS class on the spinning SVG group, which means the reduced-motion override works even if the animation is applied via inline style.
- The `useDonePhase` hook correctly cleans up all three timers on unmount/state-change via the effect cleanup return, preventing stale state updates.
- The component correctly returns `null` for the idle state and for the post-animation hidden phase, avoiding any layout impact when not in use.
- Exporting both the component and its types (`PullToRefreshRingProps`, `PullToRefreshState`) from the package index is the right pattern for consumer type safety.

---

PR #210, opened 2026-06-05
fix(sdui-runtime): dead buttons, Button label crash, root exports, dev-identity on document fetches

achin-oi:
Ports the fixes discovered during the creator app's first local end-to-end run from node_modules dist patches into source. Five fixes, one changeset (minor → 2.5.0):

1. **Every SDUI button was dead**: `DSButton` renders an inner `Pressable` that becomes the touch responder, so `SduiNode`'s outer `Clickable` never fired. `ButtonRenderer` now forwards `on_click` to `DSButton.onPress` (and passes `on_click={undefined}` to `SduiNode` so disabled buttons can't dispatch via the wrapper).
2. **Every button render crashed**: `ButtonComponentSchema` types `label` as `TextSchema` (`{ text, color?, font_size? }`) but the renderer passed it to `<Interpreter>`, which requires a wire `type` (`Cannot read property 'startsWith' of undefined`). Renderer and schema now agree.
3. **`useLocalStore` exported from the package root** — amplify-creator-app PR #22 imports it; that PR must not merge before this publishes.
4. **Document fetches now send `X-Dev-Identity`** for localhost BFF URLs, matching the action engine (shared `is-localhost` util; `bffBaseUrl` threaded through `AuthInterceptorConfig`).
5. **`getEndpoint` falls back to the codegen'd `EndpointPaths` catalog** (176 ids) — the hand-maintained registry stub covered 20 ids with divergent naming, so page-document endpoint ids failed with `Unknown endpoint`. Per the registry's own header comment, catalog-backing was the intended end state.

Tests: 145 passing; `node-registry.test` fails identically on unmodified main in a fresh local checkout (react-native flow-syntax transform via tsx) and is green in CI. Build clean.

Verified live (as dist patches) on Android emulator: bootstrap → login renders → button navigates → authenticated home → Explore tab cards from mock BFF.

### DIFF
--- a/packages/sdui-runtime/src/bff/interceptors/auth.ts
+++ b/packages/sdui-runtime/src/bff/interceptors/auth.ts
@@ -9,20 +9,12 @@
  *   - X-UTM-Source / X-UTM-Medium / X-UTM-Campaign (when present)
  */
 import { buildHeaders, type HeaderContext } from '../header-builders.js';
-import { isLocalhostBffUrl } from '../is-localhost.js';
-import { useDevConfigStore } from '../../state/useDevConfigStore.js';
 
 export interface AuthInterceptorConfig {
   /** Returns the current JWT token (may be null if not authenticated). */
   getAuthToken: () => string | null;
   /** App version string. */
   appVersion: string;
-  /**
-   * BFF base URL — used only to gate the localhost-only X-Dev-Identity
-   * header (mirrors the action engine's bff_call behaviour so document
-   * fetches and action calls present the same identity in local dev).
-   */
-  bffBaseUrl?: string;
   /** Optional UTM parameters for the current session. */
   getUtm?: () => HeaderContext['utm'];
 }
@@ -37,6 +29,13 @@ export interface AuthInterceptorConfig {
 export function applyAuthHeaders(
   init: RequestInit,
   config: AuthInterceptorConfig,
+  /**
+   * Local-dev identity value, read by the CALLER (keeping this function a
+   * stateless transformer — same pattern as the action engine's bff_call,
+   * which reads the dev-config store at its call site). Only ever set for
+   * localhost BFF URLs.
+   */
+  devIdentity?: string | null,
 ): RequestInit {
   const ctx: HeaderContext = {
     authToken: config.getAuthToken(),
@@ -46,11 +45,8 @@ export function applyAuthHeaders(
 
   const authHeaders = buildHeaders(ctx);
 
-  if (config.bffBaseUrl && isLocalhostBffUrl(config.bffBaseUrl)) {
-    const devIdentity = useDevConfigStore.getState().devIdentity;
-    if (devIdentity) {
-      (authHeaders as Record<string, string>)['X-Dev-Identity'] = devIdentity;
-    }
+  if (devIdentity) {
+    (authHeaders as Record<string, string>)['X-Dev-Identity'] = devIdentity;
   }
   const existingHeaders = init.headers as Record<string, string> | undefined;
 
--- a/packages/sdui-runtime/src/bff/interceptors/auth.ts
+++ b/packages/sdui-runtime/src/bff/interceptors/auth.ts
@@ -9,12 +9,20 @@
  *   - X-UTM-Source / X-UTM-Medium / X-UTM-Campaign (when present)
  */
 import { buildHeaders, type HeaderContext } from '../header-builders.js';
+import { isLocalhostBffUrl } from '../is-localhost.js';
+import { useDevConfigStore } from '../../state/useDevConfigStore.js';
 
 export interface AuthInterceptorConfig {
   /** Returns the current JWT token (may be null if not authenticated). */
   getAuthToken: () => string | null;
   /** App version string. */
   appVersion: string;
+  /**
+   * BFF base URL — used only to gate the localhost-only X-Dev-Identity
+   * header (mirrors the action engine's bff_call behaviour so document
+   * fetches and action calls present the same identity in local dev).
+   */
+  bffBaseUrl?: string;
   /** Optional UTM parameters for the current session. */
   getUtm?: () => HeaderContext['utm'];
 }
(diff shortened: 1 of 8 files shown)


packages/sdui-runtime/src/bff/interceptors/auth.ts
```diff
@@ -37,6 +45,13 @@ export function applyAuthHeaders(
   };
 
   const authHeaders = buildHeaders(ctx);
+
+  if (config.bffBaseUrl && isLocalhostBffUrl(config.bffBaseUrl)) {
+    const devIdentity = useDevConfigStore.getState().devIdentity;
```
zenith-oi: **FLAG** — [module-internal-structure / controller-no-business-logic (store read inside interceptor)] applyAuthHeaders is a pure header-building function but now calls useDevConfigStore.getState() directly inside it. This couples a Zustand store read (a side effect with global mutable state) into what was a stateless transformer. The pattern used in bff-call.ts is to read devIdentity at the call site and pass it in — the same approach should apply here: add an optional devIdentity?: string field to AuthInterceptorConfig and let the caller (createBffClient) read the store and pass the value in. This keeps the interceptor testable without mocking the store module and avoids the hidden dependency on global state.

zenith-oi (dismissed):
## Zenith Code Review

**Clean bug-fix PR porting five verified dist patches into source. All five stated fixes are present, the isLocalhostBffUrl extraction is a genuine improvement (regex anchoring prevents false positives on hostnames like 'notlocalhost.example.com'), and the EndpointPaths catalog fallback is well-commented. One FLAG: the auth interceptor reads Zustand store state directly inside a pure header-building function, which couples a side-effectful store read into what should be a pure transformation — worth noting but not blocking.**

### What this PR is doing
**Intent:** Port five bug fixes discovered during the creator app's first local end-to-end run from dist patches into source: dead SDUI buttons (on_click not forwarded to DSButton.onPress), crashing button renders (label routed through Interpreter instead of rendered directly), missing useLocalStore root export, missing X-Dev-Identity header on document fetches, and getEndpoint falling back to the codegen'd EndpointPaths catalog.
**Delivered:** All five fixes are present and correct in the diff. (1) ButtonRenderer now forwards on_click to DSButton.onPress and passes on_click={undefined} to SduiNode — matches changeset item 1. (2) label is rendered as DSText directly — matches changeset item 2. (3) useLocalStore + its types are exported from index.ts — matches changeset item 3. (4) bffBaseUrl threaded through AuthInterceptorConfig and X-Dev-Identity injected in applyAuthHeaders for localhost URLs — matches changeset item 4. (5) getEndpoint falls back to EndpointPaths catalog — matches changeset item 5. The isLocalhostBffUrl function is correctly extracted to a shared module and the regex is tightened vs the original substring check. No linked task page with numbered acceptance items; evaluating against PR body + changeset.

**Positive:**
- isLocalhostBffUrl extracted to a shared module with a properly anchored regex (^https?://(localhost|127\.0\.0\.1|10\.0\.2\.2)(:\d+)?(/|$)) — the original substring check would have matched hostnames like 'notlocalhost.example.com'; the regex fix is a genuine security improvement for the dev-identity gate.
- The EndpointPaths catalog fallback is clearly commented with the rationale and the important caveat that non-GET endpoints still need explicit registry entries — exactly the right level of documentation for a subtle fallback path.
- Passing on_click={undefined} to SduiNode is the correct defensive choice — it prevents a disabled button from dispatching via the outer wrapper even if the inner Pressable is somehow bypassed.

**Read against:**
- `PR body — five stated fixes`
- `.changeset/sdui-runtime-button-exports-devid.md — minor bump rationale`

---

zenith-oi (approved):
## Zenith Code Review

**Five targeted bug fixes ported from dist patches into source, all correctly implemented. The previous-round concern about store reads inside the auth interceptor is resolved — devIdentity is now read at the call site and passed as an optional parameter. The isLocalhostBffUrl helper is correctly extracted into a shared module and uses a proper URL-anchored regex rather than a loose substring match. No new issues found.**

### What this PR is doing
**Intent:** Port five bug fixes discovered during the creator app's first local end-to-end run into source: dead SDUI buttons (on_click not reaching DSButton.onPress), crashing button renders (label routed through Interpreter instead of rendered directly), missing useLocalStore root export, missing X-Dev-Identity header on document fetches, and getEndpoint falling back to the codegen'd EndpointPaths catalog.
**Delivered:** All five fixes are present in the diff and match the PR body's description exactly. (1) ButtonRenderer now forwards on_click to DSButton.onPress and passes on_click={undefined} to SduiNode. (2) label is rendered as DSText directly. (3) useLocalStore + its types are exported from index.ts. (4) client.ts reads isLocalhostBffUrl + useDevConfigStore and threads devIdentity into applyAuthHeaders. (5) getEndpoint falls back to EndpointPaths catalog. No linked task page with numbered acceptance items was provided, so no item-by-item check is possible. The previous-round FLAG about applyAuthHeaders calling useDevConfigStore internally has been addressed — the store read was moved to the call site in client.ts and the devIdentity value is passed in as a parameter, making applyAuthHeaders a stateless transformer again.

No issues found.

**Positive:**
- isLocalhostBffUrl upgraded from a loose .includes() substring match to a properly anchored regex (^https?://(localhost|127.0.0.1|10.0.2.2)(:\d+)?(/|$)) — this closes the class of false-positives where a production hostname like 'myapp-localhost-staging.example.com' would have matched the old check.
- applyAuthHeaders kept as a pure stateless transformer — devIdentity is read at the call site and injected as an optional parameter, which is the correct design for a testable interceptor function.
- Changeset entry is clear and accurately describes all five changes, which will make the release notes useful for consuming app engineers.

---

packages/sdui-runtime/src/bff/interceptors/auth.ts
achin-oi: Refactored in 9962dfc exactly as suggested: applyAuthHeaders is a stateless transformer again — it takes devIdentity as an optional value param; the Zustand read moved to the client request call site, localhost-gated, mirroring bff-call.ts. bffBaseUrl removed from AuthInterceptorConfig.

PR #214, opened 2026-06-05
fix(sdui-runtime): register creator.auth.bootstrap endpoint

achin-oi:
The bootstrap route (`GET /auth/bootstrap`) predates the `/v1/creator/*` path convention and is absent from the contracts package's codegen'd catalog — so the app's entry flow can't resolve its page document (`Unknown endpoint`). Adds the one registry entry (the registry's remaining purpose post-catalog-fallback: method overrides + catalog gaps), with a pointer to the real fix (cover bootstrap in the creator-bff OpenAPI spec). Patch changeset → 2.5.1.

### DIFF
--- a/.changeset/sdui-runtime-bootstrap-endpoint.md
+++ b/.changeset/sdui-runtime-bootstrap-endpoint.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+Add creator.auth.bootstrap to the endpoint registry — the bootstrap route predates the /v1/creator/* path convention and is absent from the codegen'd catalog, so authenticated app entry could not resolve it
--- a/packages/sdui-runtime/src/bff/endpoint-registry.ts
+++ b/packages/sdui-runtime/src/bff/endpoint-registry.ts
@@ -47,6 +47,10 @@ export function resolvePath(
  */
 export const endpointRegistry: Record<string, EndpointDefinition> = {
   // Auth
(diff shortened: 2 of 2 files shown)


zenith-oi (approved):
## Zenith Code Review

**Minimal, well-scoped fix: one registry entry + one changeset. The inline comment explains the gap clearly and points to the real fix. No issues found.**

### What this PR is doing
**Intent:** Add a missing `creator.auth.bootstrap` entry to the SDUI runtime's endpoint registry so the app's authenticated entry flow can resolve the `/auth/bootstrap` route, which predates the `/v1/creator/*` path convention and is absent from the codegen'd catalog.
**Delivered:** Fully delivered. The diff adds exactly one registry entry (`creator.auth.bootstrap → GET /auth/bootstrap`) with an inline comment explaining why it lives in the manual registry rather than the catalog, and ships a patch changeset. The changeset message accurately describes the problem and the intended permanent fix (cover bootstrap in the creator-bff OpenAPI spec). No acceptance items or task page are linked, so there is nothing to check against.

No issues found.

**Positive:**
- The inline comment at L50-L52 is exactly the right level of documentation — it explains WHY the entry exists in the manual registry (predates the path convention, absent from codegen'd catalog) and names the exit condition (cover it in the creator-bff OpenAPI spec). Future engineers won't wonder whether this is intentional or forgotten.
- The changeset message mirrors the code comment, giving the same context in the release notes without leaking task/PR IDs into the artifact — consistent with the no-task-refs-in-durable-artifacts pattern.

---

PR #221, opened 2026-06-15
feat(sdui): on-device playground + composite snippet, Tag theming, form system

achin-oi:
Lands the SDUI playground + the runtime/design-system work built alongside it.

## Highlights
- **On-device SDUI playground** (`apps/sdui-playground`, Expo 54 / RN 0.81 / React 19) for iterating renderers in isolation, fed by a local fixture server.
- **Form system** — `form_id`-keyed store, `useFormField`, validation evaluator, decoupled `submit` action; `component.field` token rhythm; composable phone country-picker.
- **Header slot** — `page_header` as a wire slot (safe-area + solid/gradient bg + back affordance) across page types and bottom sheets; native header auto-hidden when present.
- **`sdui.snippet.composite`** — one composing snippet; `data.layout` discriminant (`cover` / `stack` / `row`) names a slot-set + placement, slots hold arbitrary Nodes. Owns arrangement, never contents. `resolveRenderer` now dispatches on the layer segment so `creator.*` and `sdui.*` both resolve.
- **Tag theming** — wire `bg_color` / `text_color` / `gradient` / glyph `icon` now honored (previously every tag silently rendered default-white); `component.tag` token group.
- **Campaign card demo** on `creator.campaigns.detail`, built entirely from the cover composite.

## Verification
CI gates run locally and pass: `tokens-creator`, `ui-native`, `sdui-runtime` builds + `validate-tokens` (0 errors/0 warnings). Composite/Tag/header/group demos verified on-device (Android emulator).

## Follow-ups (tracked, not blocking)
- Promote the `submit` ActionType + the `composite` discriminated-union schema into `@one-impression/sdk-native-sdui` (renderers currently read `data` raw). The local `node_modules` SDK patch for `submit` is ephemeral and not committed.
- Namespace migration `creator.* → sdui.*` (resolver already accepts both).
- Semantic text roles (title/body/caption) to centralize the type scale.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #217, opened 2026-06-12
feat(sdui-playground): on-device playground + runtime card/gutter/icon/nav/token work

achin-oi:
Adds `apps/sdui-playground` (Expo on-device playground that renders SDUI JSON contracts through the real `sdui-runtime`), plus the runtime/ui-native work driven by it:

- **Navigation**: runtime-owned native-stack host + per-nav transitions; bottom sheets as transparentModal routes w/ tappable overlay; async page loading (skeleton/error).
- **Layout**: one container-owned page gutter + symmetric vertical row-gap (`layout/page-gutter`), with per-type + backend full-bleed / gap overrides; snippets made layout-agnostic.
- **Card**: elevation/shadow scale + outer-shadow/inner-clip structure; thin-border, no-shadow defaults (lifted from the legacy SDUI card).
- **Icons**: `IconGlyph` resolves name→SVG via the icon store (renderers were drawing empty boxes); `IconStoreProvider` mounted; manifest served by the playground fixture server.
- **Tokens**: resolvers accept both short keys and the `sdui.<group>.<kebab>` wire form so gateway/schema tokens resolve everywhere.
- Bumps `sdk-native-sdui` to `^3.0.0` (InfoRow `status_tag→tag`; Text `weight→font_weight`).

No changeset included — this lands the source on main without publishing the packages (the playground uses workspace source); publishing the runtime changes to consumers is a deliberate follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #222, opened 2026-06-15
feat(sdui): composite snippet, form system, header slot, Tag theming

achin-oi:
Rebased onto the released `main` (2.6.0) — carries **only** the new SDUI work, not the playground foundation already shipped via #217. Cherry-picked clean, no duplicated history.

## What's new
- **`sdui.snippet.composite`** — one composing snippet; `data.layout` (`cover`/`stack`/`row`) names a slot-set + placement, slots hold arbitrary Nodes. Owns arrangement, never contents. `resolveRenderer` dispatches on the layer segment so `creator.*` and `sdui.*` both resolve.
- **Header slot** — `page_header` as a wire slot (safe-area + solid/gradient bg + back affordance) across pages and bottom sheets; native header auto-hidden when present.
- **Form system** — `form_id`-keyed store, `useFormField`, validation evaluator, decoupled `submit`; `component.field` rhythm; composable `select_trigger`.
- **Tag theming** — wire `bg_color`/`text_color`/`gradient`/`icon` now honored (previously every tag rendered default-white); `component.tag` token group; icon sized to label.
- **group_config** card elevation now matches other cards (no forced shadow).
- Campaign-card demo on `creator.campaigns.detail`, built entirely from the cover composite.

## Verification
CI gates pass locally — `tokens-creator`, `ui-native`, `sdui-runtime` builds + `validate-tokens` (0/0). Card + tags + header + gradients verified on-device (Android emulator).

## Follow-ups (tracked, not blocking)
- Promote the `submit` ActionType + `composite` discriminated-union schema into `@one-impression/sdk-native-sdui` (renderers read `data` raw until then).
- Namespace migration `creator.* → sdui.*` (resolver already accepts both).

Supersedes #221 (which targeted a pre-release base and conflicted).


packages/sdui-runtime/src/form/useFormField.ts
```diff
@@ -0,0 +1,105 @@
...
+  const setErrors = useFormStore((s) => s.setErrors);
+  useEffect(() => {
+    if (!bound || !validations || validations.length === 0) return;
+    const error = evaluateField(validations, value, {
+      getFieldValue: (name) =>
+        useFormStore.getState().forms[formId as string]?.values[name],
+    });
+    setErrors(formId as string, { [fieldName as string]: error });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [bound, formId, fieldName, value, validations]);
```
zenith-oi, line 91: **FLAG** — [unstable-effect-dependency] `validations` is listed as a dependency of the validation `useEffect` (line 91). In practice, `validations` is parsed from `node.data` on every render of the parent renderer (e.g. `InputRenderer` does `data?.validations` from a raw cast), so a new array reference is produced on every render even when the contents are identical. This causes the effect to re-run on every render, calling `setErrors` repeatedly and triggering unnecessary Zustand state updates (which in turn re-render every subscriber). Fix: either stabilize the array with `useMemo` at the call site before passing it to `useFormField`, or use a ref-based deep-equality guard inside the effect. The simplest fix is to memoize in each renderer: `const validations = useMemo(() => data?.validations, [JSON.stringify(data?.validations)])` — or better, accept that the BFF sends a stable JSON blob and memoize by the raw `node.data` reference.

zenith-oi (approved):
## Zenith Code Review

**Large but well-organized PR introducing the SDUI form system, composite snippet, and several UI primitive upgrades. The implementation is thoughtful and the design doc is thorough. One FLAG-worthy issue: the validation `useEffect` in `useFormField` has `validations` (an array) in its dependency array, which will cause spurious re-runs and redundant `setErrors` calls on every render when the parent re-renders with a new array literal. This should be stabilized.**

### What this PR is doing
**Intent:** Introduce a complete SDUI form system (Zustand-keyed store, per-field hooks, declarative validation, submit action handler), a new `composite` layout snippet, floating-label Input primitive, Tag/SelectableItem theming improvements, and a pinned page_header slot with safe-area + gradient support — all wired into the sdui-runtime and ui-native packages.
**Delivered:** The diff delivers all stated goals from the PR body: the form store (`useFormStore`), `useFormField` hook, `evaluateField` validator, `handleSubmit` action, `CompositeRenderer` with cover/stack/row layouts, floating-label `Input`, updated `Tag`/`SelectableItem`, and `PageHeaderRenderer` with safe-area + gradient. The changeset file and playground fixtures are present. No linked task page with numbered acceptance items was provided, so no item-by-item gap check is possible. One meaningful concern exists: the `useFormField` validation `useEffect` calls `useFormStore.getState()` inside the effect body (a direct store read, fine) but also lists `validations` in the dependency array — since `validations` is an array prop that is likely recreated on every render from a raw wire parse, this will cause the effect to re-run and re-call `setErrors` on every render cycle, not just when the value changes. This is a performance/correctness issue worth flagging. A second concern: the `handleSubmit` action calls `store.touchAll` then immediately reads `useFormStore.getState()` for the validity check — this is correct because Zustand's `set` is synchronous, so the state is fresh. No issue there. The `no-task-refs-in-durable-artifacts` pattern is satisfied — no task/brief IDs appear in code comments. The `FORM-SYSTEM-DESIGN.md` design doc references 'D3', 'D4', 'D5', etc. as internal design decision labels, not task IDs — acceptable. Overall the PR is well-structured and the intent is delivered.

**Positive:**
- The Zustand store design is clean: per-field selectors (`s => s.forms[formId]?.values[fieldName]`) mean typing in one field only re-renders that field's component, not the whole form — this is the right granularity.
- The `evaluateField` validator is pure and dependency-free with an explicit note that it can be promoted to the SDK — good forward-thinking design.
- The `FLOAT_OVERLAP_RATIO` / `onLayout`-driven negative margin approach in `CoverLayout` is a principled solution to the straddle-row overlap problem — avoids brittle fixed pixel values.
- The `handleSubmit` action correctly calls `touchAll` before the validity check, ensuring all errors surface before aborting — the validate-on-submit UX pattern is implemented correctly.
- The `MAX_REGEX_INPUT = 2048` ReDoS guard in `evaluate.ts` is a nice defensive touch for a client-side evaluator running server-supplied patterns.
- The `register` action in `useFormStore` is correctly idempotent (non-destructive on re-render) — prevents user edits from being wiped on parent re-renders.

---

packages/sdui-runtime/src/form/useFormField.ts
achin-oi: FLAG acknowledged. `validations` is the field's wire-parsed rule array; re-validating when it changes is intended (so rule updates take effect), and the effect is bounded — `evaluateField` is pure and `setErrors` only writes the error map, which stabilizes after first run (no loop). Accepting as-is; will memoize `validations` at the parse boundary if profiling shows churn. Resolving per review approval.

PR #227, opened 2026-06-15
feat(sdui): viewport-visibility lifecycle + backend-driven infinite-scroll feeds

achin-oi:
Makes `on_view` mean *entered the viewport* (it was an `onLayout` proxy that fired on render) and uses it for fully backend-driven infinite scroll. Verified end-to-end on device.

## Runtime
- **Viewport coordinator** — `PageFeed` detects real visibility via FlatList `onViewableItemsChanged` (50% / 250ms). It fires each node's `on_view` / `on_exit` triggers honoring per-trigger **`policy` (`once` | `every`)**, with `once`-dedup keyed `(node_id, trigger_id)`. `SduiNode` defers its view lifecycle to a managing surface (`ViewportManagedProvider`) so it no longer fires on render inside a feed.
- **Backend-driven infinite scroll** — a feed renders `composite` cards as top-level `page.items`; the BFF puts an `on_view` (policy `once`) load-more on the Nth-last card, returns an `append_items`, and omits `on_view` on the final batch to terminate. **The cursor lives entirely server-side — the client takes no pagination decisions.**
- `usePageStore.appendItems` supports top-level (feed) append; `append_items` now appends **raw** nodes (validated at render) so node-level fields like `viewability` aren't stripped.

## Verified on device
15 campaigns over **2** backend-driven loads: `cursor=5` once → `cursor=10` once → stop. **No cascade on mount** (0 calls until scroll), once-dedup holds (no double-fire), cursor advances server-side, clean termination.

## Contract
Runs on the installed SDK **3.1.0** (coordinator reads `viewability` raw; the append fix preserves it). The typed `viewability` contract is merged in amplify-schemas (#312) and ships in **3.2.0** (Version PR pending). Follow-up: bump the SDK dep `3.1.0 → 3.2.0` once published — typed-contract only, no behavior change.

Pairs with amplify-schemas#312.

### DIFF
--- a/apps/sdui-playground/server/fixture-server.mjs
+++ b/apps/sdui-playground/server/fixture-server.mjs
@@ -55,6 +55,84 @@ const readJsonBody = (req) =>
     });
   });
 
+// ─── Campaigns feed (composite cards + backend-driven infinite scroll) ───────
+// The feed renders `sdui.snippet.composite` cards as top-level page.items. The
+// 2nd-last card of each batch carries a `viewability.on_view` (policy: once)
+// that bff_calls this server for the next batch; the server returns an
+// `append_items` action. The LAST batch omits `on_view`, so it self-terminates
+// — the cursor lives entirely here, never on the client.
+const FEED_PAGE_SIZE = 5;
+const FEED_TOTAL = 15;
+const FEED_BRANDS = [
+  "Lumina Beauty", "Verde Skincare", "Aura Cosmetics", "Nova Wellness",
+  "Bloom Organics", "Ceré Paris", "Indigo Hair", "Solstice Fragrance",
+];
+
+function campaignCard(i) {
+  const brand = FEED_BRANDS[(i - 1) % FEED_BRANDS.length];
+  return {
+    type: "sdui.snippet.composite",
+    id: `campaign-${i}`,
+    on_click: { type: "navigate", payload: { target: "creator.campaigns.detail", op: "push", params: { id: String(i) } } },
+    data: {
+      layout: "cover",
+      surface: { bg_color: "sdui.color.neutral-inverse", border_color: "sdui.color.neutral-subtle" },
+      media: { type: "creator.snippet.banner_image", id: `m-${i}`, data: { image: { src: `https://picsum.photos/seed/camp${i}/640/320`, aspect_ratio: 2 } } },
+      float: { type: "creator.ui_component.image", id: `a-${i}`, data: { src: `https://picsum.photos/seed/brand${i}/120`, width: 64, height: 64, border_radius: "sdui.radius.full" } },
+      float_end: [
+        { type: "creator.ui_component.tag", id: `cash-${i}`, data: { label: { text: `₹${(1000 + i * 200).toLocaleString("en-IN")} Cash` }, bg_color: "sdui.color.positive-weak", text_color: "sdui.color.positive" } },
+      ],
+      body: [
+        { type: "creator.snippet.info_row", id: `n-${i}`, data: { title: { text: `${brand} · #${i}`, font_weight: "medium", font_size: "sdui.font-size.lg" } } },
+        { type: "creator.snippet.info_row", id: `meta-${i}`, data: { title: { text: "Beauty · Micro creators", font_size: "sdui.font-size.sm", color: "sdui.color.neutral-medium" } } },
+      ],
+      footer: { type: "creator.snippet.info_row", id: `f-${i}`, data: { title: { text: "Apply now — closes soon", font_size: "sdui.font-size.sm" } } },
+    },
+  };
+}
+
+// `nextCursor` null ⇒ last page (no on_view ⇒ stops). Otherwise the 2nd-last
+// card carries the load-more trigger pointing at the next cursor.
+function campaignBatch(startId, count, nextCursor) {
+  const cards = [];
+  for (let k = 0; k < count; k++) {
+    const card = campaignCard(startId + k);
+    if (nextCursor !== null && k === count - 2) {
+      card.viewability = {
+        on_view: [{
+          id: "load-more",
+          policy: "once",
+          action: { type: "bff_call", payload: { endpoint: "creator.campaigns.list", method: "GET", query_params: { cursor: String(nextCursor) } } },
+        }],
+      };
+    }
+    cards.push(card);
+  }
+  return cards;
+}
+
+function buildFeedPage() {
+  const nextCursor = FEED_TOTAL > FEED_PAGE_SIZE ? FEED_PAGE_SIZE : null;
+  return {
+    id: "demo.feed",
+    title: "Campaigns",
+    protocol_version: "1.0.0",
+    layout: "feed",
+    data: {
+      header: {
+        type: "creator.snippet.page_header",
+        id: "feed-hdr",
+        data: {
+          title: { text: "Campaigns", font_size: "sdui.font-size.xl", font_weight: "bold", color: "sdui.color.neutral-inverse" },
+          subtitle: { text: "infinite scroll · composite cards", color: "sdui.color.neutral-inverse" },
+          background: { gradient: { colors: ["#7C3AED", "#F2F2F2"], angle: 180 } },
+        },
+      },
+    },
+    items: campaignBatch(1, FEED_PAGE_SIZE, nextCursor),
+  };
+}
+
 const server = createServer(async (req, res) => {
   const url = req.url ?? "/";
   // eslint-disable-next-line no-console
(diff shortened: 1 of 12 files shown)


packages/sdui-runtime/src/action-engine/handlers/append-items.ts
```diff
@@ -1,20 +1,29 @@
...
+ * which would break the very triggers driving pagination. Keeping the raw nodes
+ * is also forward-compatible with any future node field.
  */
 export async function handleAppendItems(
   action: Action,
   _config: ActionEngineConfig,
   _engine: ActionEngine,
 ): Promise<void> {
   const payload = AppendItemsPayloadSchema.parse(action.payload);
+  const rawItems =
```
zenith-oi, line 22: **FLAG** — [input-validation-at-boundary] The raw-item extraction reads `(action.payload as { items?: Node[] } | undefined)?.items` and falls back to `payload.items` (the Zod-parsed result). The comment correctly explains why raw nodes must be preserved — Zod strips unknown fields like `viewability`. But the fallback to `payload.items` is the exact stripped value you're trying to avoid: if the cast somehow returns undefined (e.g. `action.payload` is a non-object), you silently use the schema-parsed items and lose `viewability`, breaking the very pagination triggers this PR ships. A safer pattern is to read the raw items first and only fall back to an empty array (or throw) rather than falling back to the stripped parse result:

```ts
const rawItems = (action.payload as Record<string, unknown>)?.items as Node[] | undefined;
if (!rawItems) {
  // payload.items from Zod is the stripped fallback — log a warning and use it,
  // or throw if items is required
}
```

Alternatively, skip the Zod parse of `items` entirely by using `AppendItemsPayloadSchema` with `items` typed as `z.unknown().array()` (passthrough), which is the cleanest fix and removes the dual-read entirely.

_Cited:_ `PR body — 'append_items appends the raw nodes (validated at render) so node-level fields like viewability aren't stripped'`

apps/sdui-playground/server/fixture-server.mjs
```diff
@@ -93,8 +171,32 @@ const server = createServer(async (req, res) => {
...
+  if (url.startsWith("/v1/creator/campaigns")) {
+    const cursor = Number(new URL(url, "http://x").searchParams.get("cursor") || "0");
+    const count = Math.min(FEED_PAGE_SIZE, FEED_TOTAL - cursor);
+    if (count <= 0) {
+      json(res, 200, { action: { type: "append_items", payload: { target: "demo.feed", items: [], has_more: false } } });
+      return;
+    }
+    const nextCursor = cursor + count < FEED_TOTAL ? cursor + count : null;
+    const items = campaignBatch(cursor + 1, count, nextCursor);
+    console.log(`[fixture-server] campaigns cursor=${cursor} → cards ${cursor + 1}..${cursor + count} (next=${nextCursor})`);
```
zenith-oi, line 187: **FLAG** — [pii-logging-discipline] `console.log` is used for server-side debug output in the fixture server. This is a playground/dev-only server so it's low severity, but the coding standard says never use `console.log` in production code. Since this file ships in the repo and could be run in CI or demo environments, it should use a proper logger or at minimum be gated behind a `DEBUG` env flag. The existing `// eslint-disable-next-line no-console` pattern used elsewhere in this file (line ~138) should be applied here too if the log is intentionally kept.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-scoped feature PR. The viewport lifecycle design is correct — real visibility via FlatList, stable refs, once-dedup keyed by (nodeId, phase, triggerId), and a context flag that prevents the onLayout proxy from double-firing inside a managed surface. One genuine issue worth flagging: the raw-item extraction in append-items.ts has a subtle redundancy that could silently fall back to the schema-stripped items in an edge case.**

### What this PR is doing
**Intent:** Add a real viewport-visibility lifecycle to the SDUI runtime's PageFeed (replacing the onLayout proxy) and use it to drive fully backend-controlled infinite scroll via on_view triggers on feed items.
**Delivered:** The diff fully delivers the stated intent. The viewport coordinator (ViewportManagedContext + fireViewability + FlatList onViewableItemsChanged wiring) is complete and correct. The SduiNode deferral path is clean. The top-level feed append in usePageStore is correct. The raw-node preservation in append-items handler is sound and well-documented. The playground fixture demonstrates the end-to-end flow. No acceptance items are missing.

**Positive:**
- The ViewportManagedContext design is exactly right — a boolean context flag that defaults false means all existing non-feed SduiNode behavior is completely unchanged with zero risk of regression.
- Stabilizing both viewabilityConfig and onViewableItemsChanged as refs (not inline objects/functions) is the correct React Native pattern — FlatList throws if these change after mount, and this PR gets it right from the start.
- The once-dedup key `${nodeId}::${phase}::${triggerIndex}` correctly handles the case where a trigger has no explicit id, falling back to positional index — this prevents double-fire on re-entry without requiring the BFF to assign IDs to every trigger.
- The top-level feed append in usePageStore (targeting page.id directly) is a clean, minimal change that doesn't disturb the existing tree-walk path for nested list nodes.
- The changeset entry is detailed and accurate — it describes the behavioral contract (50%/250ms, once-dedup, server-side cursor, clean termination) in a way that will be useful for consumers reading the changelog.

**Read against:**
- `PR body — viewport coordinator, backend-driven infinite scroll, raw-node preservation, SDK 3.1.0/3.2.0 contract note`
- `changeset/sdui-viewport-lifecycle-feed.md — feature summary and behavioral spec`

---

packages/sdui-runtime/src/action-engine/handlers/append-items.ts
zenith-oi, line 22: **FLAG** — [input-validation-at-boundary] The raw-item extraction `(action.payload as { items?: Node[] } | undefined)?.items ?? payload.items` bypasses the Zod-parsed result for the items array. The comment explains the rationale (avoid stripping unknown fields like `viewability`), but the pattern is fragile: if `action.payload` is not an object or `items` is not an array, the fallback silently passes `payload.items` (the Zod-stripped version), defeating the purpose. A safer approach is to use `AppendItemsPayloadSchema` with `.passthrough()` on the items array element schema, or to use `z.array(z.unknown())` for items in the schema and cast afterward — that way the Zod parse still validates the envelope shape while preserving unknown fields on each item node.

apps/sdui-playground/server/fixture-server.mjs
zenith-oi, line 187: **FLAG** — [pii-logging-discipline] console.log is used for debug output (`[fixture-server] campaigns cursor=...`). This is a dev-only fixture server so there is no PII risk, but the Zenith coding standard says never use console.log in production code. Since this is playground-only and the log contains no sensitive data, this is low priority — but worth a note for consistency.

zenith-oi (approved):
## Zenith Code Review

**Clean, well-scoped feature PR. The viewport coordinator design is sound — stable ref for viewabilityConfig, once-dedup keyed by (nodeId, phase, triggerId), correct separation between managed and self-managed surfaces. One carry-over FLAG from the previous round (raw-item extraction pattern) and one carry-over FLAG (console.log in fixture server) remain unaddressed; both are low-severity and playground/dev-only respectively.**

### What this PR is doing
**Intent:** Add a real viewport-visibility lifecycle to PageFeed (replacing the onLayout proxy) and use it to power fully backend-driven infinite scroll, where the BFF places an on_view trigger on the Nth-last card and returns an append_items action to load the next batch.
**Delivered:** The diff fully delivers the stated intent. ViewportManagedContext + fireViewability implement the coordinator; PageFeed wires onViewableItemsChanged with a stable config ref and wraps the list in ViewportManagedProvider; SduiNode defers its own view lifecycle when inside a managed surface; usePageStore gains top-level feed append; append-items handler preserves raw nodes to avoid stripping viewability fields; the playground fixture server demonstrates 15 cards over 2 backend-driven loads. All described behaviors are present in the diff.

**Positive:**
- viewabilityConfig is correctly held in a ref (viewabilityConfigRef) and handleViewableItemsChanged is stabilized via useRef(...).current — this is the exact pattern React Native requires to avoid the FlatList 'Changing viewabilityConfig on the fly is not supported' error. Good defensive engineering.
- The once-dedup key `${nodeId}::${phase}::${t.id ?? i}` is well-designed: it scopes dedup to (node, phase, trigger) so a node re-entering the viewport after exit correctly re-fires 'every' triggers while 'once' triggers stay suppressed.
- ViewportManagedContext defaults to false, so all existing non-feed pages continue using the self-managed Viewable/onLayout path with zero behavior change — clean opt-in design.
- The top-level feed append in usePageStore (targeting page.id) is a minimal, correct addition that doesn't disturb the existing appendItemsInTree path for nested list nodes.
- The changeset entry is detailed and accurate, covering the coordinator, infinite scroll contract, and playground demo — good for consumers tracking the minor bump.

---

packages/sdui-runtime/src/action-engine/handlers/append-items.ts
achin-oi: FLAG acknowledged — intentional. Appended nodes are validated at RENDER by SduiNode (each node's data is parsed against its snippet schema, with a fallback), and the envelope (target / cursor / has_more) IS validated here via AppendItemsPayloadSchema. We deliberately append the RAW nodes so node-level fields the installed SDK may not yet model (e.g. `viewability`) aren't silently stripped — that strip was the pagination bug this fixes. Documented in the handler comment. Resolving per approval.

apps/sdui-playground/server/fixture-server.mjs
achin-oi: FLAG acknowledged — fixture-server.mjs is the local playground dev fixture server; it is never built into the app or deployed. The logged values are request URLs + synthetic card indices/cursors (no PII). console.log is the right output for a dev tool. Resolving per approval.

PR #229, opened 2026-06-15
feat(sdui-runtime): region page model — scaffold, region reload, render-bindings, skeleton

achin-oi:
## What
The runtime for **shell-first, region-scoped SDUI pages** (tabbed / filtered feeds), on `@one-impression/sdk-native-sdui@^3.4.0`. Supersedes the earlier request-context primitives this PR opened with.

### Base scaffold
`usePageScaffold(page)` owns every cross-cutting page concern — lifecycle (`on_load` once / `on_dismount` / back / app-state), the live-page subscription, the `reload` partial-merge, per-region loading, bottom-sheet registration, refresh — and exposes `getRegion(name)` → content-or-skeleton. A layout becomes pure zone geometry; `PageFeed` is moved onto it. Every page gets the region capability uniformly (dormant unless it declares regions/skeletons).

### Region model
- **Region-scoped `reload`** + **partial-page merge** in `usePageStore` (`response.data` shallow-merges, `response.items` replaces; per-region loading flags drive skeletons). One verb: `["content"]` for a filter, `["header","content"]` for a tab switch; the footer shell never reloads.
- **Reactive render-bindings** — `{ ref: "$.local.<key>" }` (+ `contains`/`equals`) resolved before validation in `SduiNode`, so a chip's `selected` / a tab's `active` reflect local state instantly, no reload.
- **`set_local` `array_toggle`** (multi-select) + **backend-controlled debounce**.
- **`creator.snippet.skeleton`** renderer — composable shimmer (`rect`/`line`/`circle`, horizontal row groups, padding); per-region skeletons.

### Fixes
Chip wires `on_click` through its pressable + shows a remove **×** (trailing, icon-store glyph); `TabsFooter` bottom safe-area; nav host hides the native header when a header **region** is declared (`data.header_skeleton`). **ui-native `Chip`** gains a `trailingIcon` slot.

## Verification
Built + tested against published `sdk-native-sdui@3.4.0`; runtime suite green (the one red `node-registry` test is a pre-existing RN-Flow-parse env issue). Shell-first feed verified on-device: single header, filter chips, content cards, footer; `on_load → reload(["header","content"])` streams the tab in over the shell.

## Design
`packages/sdui-runtime/REGION-PAGE-MODEL.md`. Additive — existing pages/snippets unchanged.

### DIFF
--- a/apps/sdui-playground/server/fixture-server.mjs
+++ b/apps/sdui-playground/server/fixture-server.mjs
@@ -20,6 +20,10 @@ import { dirname, join } from "node:path";
  * `adb reverse tcp:3012 tcp:3012`.
  */
 const PORT = Number(process.env.PORT ?? 3012);
+// Optional simulated reload latency (ms) so the shimmer/skeleton is observable
+// while hand-testing the playground. Off by default (0) so it never slows
+// automated callers; set SDUI_FIXTURE_LATENCY_MS=3000 to watch the skeletons.
+const RELOAD_LATENCY_MS = Number(process.env.SDUI_FIXTURE_LATENCY_MS) || 0;
 const PAGE_PREFIX = "/sdui/page/";
 const SUBMIT_PREFIX = "/submit/";
 const HERE = dirname(fileURLToPath(import.meta.url));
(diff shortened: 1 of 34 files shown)


apps/sdui-playground/server/fixture-server.mjs
```diff
@@ -171,30 +322,30 @@ const server = createServer(async (req, res) => {
...
-      json(res, 200, { action: { type: "append_items", payload: { target: "demo.feed", items: [], has_more: false } } });
-      return;
-    }
-    const nextCursor = cursor + count < FEED_TOTAL ? cursor + count : null;
-    const items = campaignBatch(cursor + 1, count, nextCursor);
-    console.log(`[fixture-server] campaigns cursor=${cursor} → cards ${cursor + 1}..${cursor + count} (next=${nextCursor})`);
-    json(res, 200, { action: { type: "append_items", payload: { target: "demo.feed", items, has_more: nextCursor !== null } } });
+    // TEMP (debug): artificial 3s delay so the shimmer/skeleton stays on screen
+    // long enough to inspect. Remove before shipping — real reloads are fast.
+    await new Promise((r) => setTimeout(r, 3000));
```
zenith-oi: **BLOCK** — [debug-artifact-in-mergeable-code] A hardcoded `await new Promise((r) => setTimeout(r, 3000))` is present with an explicit comment 'Remove before shipping — real reloads are fast.' This 3-second artificial delay will cause any automated test or CI job that exercises the `/v1/creator/campaigns` endpoint to time out or hang. The comment acknowledges it must be removed; it should not be in a mergeable commit. Remove the delay (or gate it behind a `FIXTURE_DELAY_MS` env var defaulting to 0) before merging.

zenith-oi (dismissed):
## Zenith Code Review

**Solid, well-scoped engine PR. The new primitives (local refs, request binding, reload, debounce, scaffold hook, skeleton renderer) are cleanly separated, well-tested, and additive. One real issue: the fixture server leaves a hardcoded 3-second artificial delay that will break any CI/automated test that hits the fixture endpoint, and the comment says 'remove before shipping' — it should not merge as-is.**

### What this PR is doing
**Intent:** Add the runtime engine primitives for tabbed/filtered SDUI pages: `$.local.*` ref resolution, `bindRequestPayload` for binding local store state into BFF request fields, a `reload` action handler with partial-region merge, backend-controlled debounce in the action engine, a `Skeleton` snippet renderer, and a `usePageScaffold` hook that centralises page lifecycle — all demonstrated via a reworked fixture server.
**Delivered:** The diff delivers the stated intent fully. `$.local.*` ref resolution is in `resolve-ref.ts` with `contains`/`equals` predicates. `bindRequestPayload` + `normalizeQueryParams` are in `resolve-request-refs.ts`. The `reload` handler is in `handlers/reload.ts`. Debounce is in `action-engine.ts`. `usePageScaffold` centralises lifecycle. `SkeletonRenderer` is new. The fixture server is reworked to demonstrate the shell-first region model. Unit tests cover all new paths. The PR body explicitly defers on-device verification to a follow-up PR, which is appropriate. No acceptance items are missing relative to what the PR claims to deliver.

**1 BLOCK** — must fix before merge. See inline comments.

**Positive:**
- The `bindRequestPayload` / `resolveRequestRefs` / `normalizeQueryParams` split is clean: pure functions with no store imports, tested in isolation, and the CSV-coercion + empty-array-drop behaviour is explicitly tested.
- Debounce key design (verb + target + endpoint) is well-reasoned and the 'immediate dispatch cancels pending debounced run of the same key' invariant (tab switch cancels in-flight filter reload) is both documented and tested.
- Extracting `usePageScaffold` is the right call — it removes ~100 lines of duplicated lifecycle boilerplate from `PageFeedRenderer` and gives every future layout renderer the same capabilities for free.
- The `SkeletonRenderer` being fully BFF-composed (no hardcoded shapes) is the correct design; the shape vocabulary (rect/line/circle, row groups, justify) is minimal and covers the demonstrated use cases.
- Test coverage is thorough: debounce coalescing, key isolation, no-loop invariant, reload partial merge, local-ref binding into query string, and loading-state cleanup on error are all exercised.

**Read against:**
- `PR body — stated scope and explicit deferrals`
- `packages/sdui-runtime/REGION-PAGE-MODEL.md — design doc shipped in this PR`

---

apps/sdui-playground/server/fixture-server.mjs
achin-oi: Addressed in the latest push. The latency is now opt-in via `RELOAD_LATENCY_MS` (from `SDUI_FIXTURE_LATENCY_MS`, **default 0**) — no hardcoded delay, so automated callers are never slowed. The fixture is the private (`private: true`, unpublished) playground's mock BFF; the env-gated latency is intentional test tooling for observing the skeleton, set explicitly when hand-testing.

packages/sdui-runtime/src/action-engine/handlers/reload.ts
```diff
@@ -0,0 +1,61 @@
...
+    // Partial page: { data?: Record<string, unknown>, items?: Node[] }.
+    const body = (await res.json()) as {
+      data?: Record<string, unknown>;
+      items?: Node[];
+    };
+    usePageStore.getState().mergeRegions({ data: body.data, items: body.items });
+  } finally {
+    usePageStore.getState().setRegionsLoading(payload.regions, false);
+  }
+}
```
zenith-oi, line 61: **FLAG** — [missing-error-handling-on-external-call] The `handleReload` handler (not shown in the truncated diff but exercised by the reload.test.ts at L110-116) clears `loadingRegions` on error per the test — good. However, the test at L114 asserts the error message matches `/reload: BFF GET .* returned 500/`. If the `fetch` call itself throws a network error (timeout, DNS failure, connection refused) rather than returning a non-OK response, the `loadingRegions` flag for the affected regions may not be cleared, leaving the skeleton visible indefinitely. Confirm the `finally` block (or equivalent) in `reload.ts` clears loading flags on both non-OK responses AND thrown fetch errors.

zenith-oi (approved):
## Zenith Code Review

**Solid, well-structured PR that introduces the shell-first region model cleanly. The `usePageScaffold` extraction is the right abstraction, the debounce and ref-binding logic is well-tested, and the fixture server demo is a good end-to-end illustration. One carry-over BLOCK from the previous round remains unresolved: the simulated latency in the fixture server is now env-gated (not hardcoded at 3 s), which addresses the spirit of the complaint — see issue below for the delta assessment.**

### What this PR is doing
**Intent:** Introduce shell-first, region-scoped SDUI pages: a `usePageScaffold` base hook, a `reload` action handler with debounce and `$.local.*` ref-binding, a composable `Skeleton` renderer, reactive render-bindings (`contains`/`equals`), and a fixture-server demo wiring tabs + filter chips through the new primitives.
**Delivered:** The diff fully delivers the stated intent. `usePageScaffold` extracts all cross-cutting page concerns from `PageFeedRenderer`, which is now pure zone geometry. The `reload` handler, `bindRequestPayload`, `resolveRequestRefs`, `normalizeQueryParams`, and debounce logic are all present and tested. The `Skeleton` renderer, `set_local array_toggle`, `$.local.*` ref resolution with `contains`/`equals` modifiers, and the fixture-server demo are all implemented. The previous-round BLOCK about a hardcoded 3-second delay is the only outstanding item — see below.

**Positive:**
- `usePageScaffold` is a clean extraction — all cross-cutting page concerns (lifecycle, back-press, app-state, bottom-sheets, live-page subscription, refresh) live in one place; layout renderers are reduced to pure zone geometry. This is the right abstraction boundary.
- The debounce implementation is module-level (survives across dispatch calls in a burst) and the immediate-dispatch-cancels-pending-debounce path (tab switch cancels in-flight filter reload) is a subtle but important correctness detail that's well-commented and tested.
- `bindRequestPayload` is pure (takes a local snapshot, no store import), making it trivially testable — and the test suite covers the empty-array drop, mutation-safety, and CSV coercion cases thoroughly.
- The `Skeleton` renderer is genuinely composable: BFF-composed layout, no hardcoded shapes, `repeat` + `card` + `padding` give the BFF enough knobs to match any region's content without shipping new renderer code.
- The fixture server comment block at L62-71 is an unusually clear explanation of the shell-first contract — it will save the next engineer significant ramp-up time.

**Read against:**
- `packages/sdui-runtime/REGION-PAGE-MODEL.md`
- `PR body — shell-first region model, usePageScaffold, reload, debounce, reactive render-bindings, Skeleton renderer`

---

packages/sdui-runtime/src/action-engine/handlers/reload.ts
achin-oi: Confirmed handled. `setRegionsLoading(payload.regions, false)` is in a `finally` block, not a `catch` — so it clears the loading flags on **every** exit path: a non-OK response (thrown), a thrown `fetch` network error (timeout / DNS / connection refused), a JSON parse error, and success. The skeleton can't get stuck. The test asserts the non-OK message specifically, but the loading-cleanup is path-agnostic via `finally`.

PR #242, opened 2026-06-16
feat(sdui-runtime): alias sdui.* node-type namespace (registry)

achin-oi:
Step 1 of the `creator.* → sdui.*` migration (see `NAMESPACE-MIGRATION.md`). Adds `sdui.*` aliases for every `creator.*` registry key (same renderer); `creator.*` keys retained for back-compat. Resolver dispatch unchanged. Flips the playground fixtures to `sdui.*`.

**Ships first** — the runtime must render `sdui.*` before any producer emits it (the SDK rename PR follows).

### DIFF
--- a/packages/sdui-runtime/src/registries/snippets.ts
+++ b/packages/sdui-runtime/src/registries/snippets.ts
@@ -126,9 +126,12 @@ const baseSnippetRegistry: Record<string, (node: Node) => React.ReactElement> = 
 
 // Namespace migration: alias every `creator.*` entry to its `sdui.*` key so both
 // prefixes resolve to the same renderer while emitters move creator.* -> sdui.*.
+// Derived aliases are spread FIRST and the explicit base map LAST, so an
+// intentional explicit key (e.g. `sdui.snippet.composite`) always wins over a
+// derived alias and can never be silently overwritten.
 export const snippetRegistry: Record<string, (node: Node) => React.ReactElement> = {
-  ...baseSnippetRegistry,
   ...Object.fromEntries(
     Object.entries(baseSnippetRegistry).map(([k, v]) => [k.replace(/^creator\./, "sdui."), v]),
   ),
+  ...baseSnippetRegistry,
 };
--- a/packages/sdui-runtime/src/registries/snippets.ts
+++ b/packages/sdui-runtime/src/registries/snippets.ts
@@ -63,7 +63,7 @@ import { ChipRenderer } from "../snippets/Chip/index.js";
  * 43 snippets covering layout, headers, footers, cards, images,
  * info rows, inputs, and chips.
  */
-export const snippetRegistry: Record<string, (node: Node) => React.ReactElement> = {
+const baseSnippetRegistry: Record<string, (node: Node) => React.ReactElement> = {
   // Layout / Utility
   // First citizen of the domain-neutral `sdui.*` namespace (see migration plan).
   "sdui.snippet.composite": CompositeRenderer,
@@ -122,4 +122,13 @@ export const snippetRegistry: Record<string, (node: Node) => React.ReactElement>
 
   // Chip (1)
   "creator.snippet.chip": ChipRenderer,
+};
+
+// Namespace migration: alias every `creator.*` entry to its `sdui.*` key so both
+// prefixes resolve to the same renderer while emitters move creator.* -> sdui.*.
+export const snippetRegistry: Record<string, (node: Node) => React.ReactElement> = {
+  ...baseSnippetRegistry,
+  ...Object.fromEntries(
+    Object.entries(baseSnippetRegistry).map(([k, v]) => [k.replace(/^creator\./, "sdui."), v]),
+  ),
 };
--- a/packages/sdui-runtime/src/registries/snippets.ts
+++ b/packages/sdui-runtime/src/registries/snippets.ts
@@ -124,14 +124,15 @@ const baseSnippetRegistry: Record<string, (node: Node) => React.ReactElement> = 
   "creator.snippet.chip": ChipRenderer,
 };
 
-// Namespace migration: alias every `creator.*` entry to its `sdui.*` key so both
-// prefixes resolve to the same renderer while emitters move creator.* -> sdui.*.
-// Derived aliases are spread FIRST and the explicit base map LAST, so an
-// intentional explicit key (e.g. `sdui.snippet.composite`) always wins over a
-// derived alias and can never be silently overwritten.
+// Namespace migration: for each `creator.*` entry, add an `sdui.*` alias to the
+// same renderer (entries already in the `sdui.*` namespace are left untouched —
+// the loop only transforms `creator.*` keys). The base map is spread last, so
+// any explicitly declared key remains authoritative.
 export const snippetRegistry: Record<string, (node: Node) => React.ReactElement> = {
   ...Object.fromEntries(
-    Object.entries(baseSnippetRegistry).map(([k, v]) => [k.replace(/^creator\./, "sdui."), v]),
+    Object.entries(baseSnippetRegistry)
+      .filter(([k]) => k.startsWith("creator."))
+      .map(([k, v]) => [k.replace(/^creator\./, "sdui."), v]),
   ),
   ...baseSnippetRegistry,
 };
--- a/.changeset/namespace-sdui-registry.md
+++ b/.changeset/namespace-sdui-registry.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+Alias the `sdui.*` node-type namespace in the renderer registries. Every `creator.snippet.*` / `creator.ui_component.*` key now has an `sdui.*` alias resolving to the same renderer, so both prefixes render identically while emitters migrate `creator.* → sdui.*`. The resolver dispatch (on the `.snippet.`/`.ui_component.` segment) is unchanged; `creator.*` keys are retained for backward compatibility. See `NAMESPACE-MIGRATION.md`.
--- a/apps/sdui-playground/server/fixture-server.mjs
+++ b/apps/sdui-playground/server/fixture-server.mjs
@@ -120,16 +120,16 @@ function campaignCard(c) {
     data: {
       layout: "cover",
       surface: { bg_color: "sdui.color.neutral-inverse", border_color: "sdui.color.neutral-subtle" },
-      media: { type: "creator.snippet.banner_image", id: `m-${c.id}`, data: { image: { src: `https://picsum.photos/seed/camp${c.id}/640/320`, aspect_ratio: 2 } } },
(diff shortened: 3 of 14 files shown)


packages/sdui-runtime/src/registries/snippets.ts
```diff
@@ -123,3 +123,12 @@ export const snippetRegistry: Record<string, (node: Node) => React.ReactElement>
   // Chip (1)
   "creator.snippet.chip": ChipRenderer,
 };
+
+// Namespace migration: alias every `creator.*` entry to its `sdui.*` key so both
+// prefixes resolve to the same renderer while emitters move creator.* -> sdui.*.
+export const snippetRegistry: Record<string, (node: Node) => React.ReactElement> = {
```
zenith-oi, line 131: **FLAG** — [registry-alias-overwrites-native-sdui-key] `baseSnippetRegistry` already contains `"sdui.snippet.composite": CompositeRenderer` (line 69, the comment even calls it 'First citizen of the domain-neutral sdui.* namespace'). The alias spread at L129–L133 maps every `creator.*` key to its `sdui.*` equivalent. If a `creator.snippet.composite` key were ever added to `baseSnippetRegistry`, the spread would silently overwrite the intentional `sdui.snippet.composite` entry with whatever renderer `creator.snippet.composite` maps to. More immediately: the current spread produces a final registry where `sdui.snippet.composite` appears twice in the merge — once from `...baseSnippetRegistry` and once from the alias map (if a `creator.snippet.composite` ever exists). The alias regex `k.replace(/^creator\./, 'sdui.')` will not produce a collision today because there is no `creator.snippet.composite`, but the pattern is fragile. Recommend adding a guard or comment: either assert that no `sdui.*` key in the alias output collides with an existing `sdui.*` key in `baseSnippetRegistry`, or document explicitly that `sdui.snippet.composite` is the only pre-existing `sdui.*` key and the alias map will never produce it (since there is no `creator.snippet.composite`). A one-line comment or a runtime assertion at startup would prevent a future silent overwrite.

_Cited:_ `packages/sdui-runtime/src/registries/snippets.ts#L68 — sdui.snippet.composite is the only pre-existing sdui.* key; the alias spread must not clobber it`

zenith-oi (dismissed):
## Zenith Code Review

**Clean namespace-migration step: runtime registries now serve both `creator.*` and `sdui.*` keys via a single alias spread, and all playground fixtures have been mechanically updated to `sdui.*`. The approach is correct and backward-compatible. One genuine concern worth flagging: the alias map silently drops any `sdui.*`-native key that was already in `baseSnippetRegistry` (specifically `sdui.snippet.composite`) because the regex replace on `creator.*` entries won't touch it, but the spread order means a future `creator.snippet.composite` alias would overwrite it — worth a quick audit.**

### What this PR is doing
**Intent:** Migrate the SDUI playground fixtures and runtime registries from the `creator.*` namespace to `sdui.*` by adding aliased registry entries (both prefixes resolve to the same renderer) and updating all fixture JSON/JS files to use `sdui.*` type strings — Step 1 of the `creator.* → sdui.*` namespace migration.
**Delivered:** Fully delivered. The runtime registries (`snippets.ts`, `ui-components.ts`) now export both `creator.*` and `sdui.*` keys pointing to the same renderers via the spread+map alias pattern. All playground fixture files (JSON and the dynamic fixture server) have been updated to emit `sdui.*` type strings. The changeset file documents the change. The NAMESPACE-MIGRATION.md update is present. No acceptance items are cited in the PR body beyond the stated intent, and the diff satisfies that intent completely.

**Positive:**
- The alias pattern (spread base + Object.fromEntries map) is the right approach for a zero-downtime namespace migration — both prefixes resolve to the same renderer with no duplication of renderer imports.
- Keeping `creator.*` keys in `baseSnippetRegistry` intact means any producer still emitting the old namespace continues to work without a flag day.
- The changeset file and NAMESPACE-MIGRATION.md update are present, making the migration state discoverable for future engineers.

**Read against:**
- `packages/sdui-runtime/NAMESPACE-MIGRATION.md — migration plan referenced in PR body`
- `packages/sdui-runtime/src/registries/snippets.ts — alias pattern implementation`
- `packages/sdui-runtime/src/registries/ui-components.ts — alias pattern implementation`

---

packages/sdui-runtime/src/registries/snippets.ts
achin-oi: Fixed in 2985079 — the derived `sdui.*` aliases are now spread first and the explicit `baseSnippetRegistry` (and `baseUiComponentRegistry`) last, so an intentional explicit key like `sdui.snippet.composite` always wins and can never be overwritten by a `creator.*`-derived alias.

packages/sdui-runtime/src/registries/snippets.ts
```diff
@@ -123,3 +123,15 @@ export const snippetRegistry: Record<string, (node: Node) => React.ReactElement>
...
   "creator.snippet.chip": ChipRenderer,
 };
+
+// Namespace migration: alias every `creator.*` entry to its `sdui.*` key so both
+// prefixes resolve to the same renderer while emitters move creator.* -> sdui.*.
+// Derived aliases are spread FIRST and the explicit base map LAST, so an
+// intentional explicit key (e.g. `sdui.snippet.composite`) always wins over a
+// derived alias and can never be silently overwritten.
+export const snippetRegistry: Record<string, (node: Node) => React.ReactElement> = {
+  ...Object.fromEntries(
```
zenith-oi, line 132: **FLAG** — [registry-alias-overwrites-native-sdui-key] The alias-generation loop at L133-L135 iterates over `baseSnippetRegistry`, which already contains `"sdui.snippet.composite": CompositeRenderer` (L69). The `replace(/^creator\./, 'sdui.')` transform on that key is a no-op (the key doesn't start with `creator.`), so the loop emits `"sdui.snippet.composite" → CompositeRenderer` as an alias entry. The base-map spread at L136 then overwrites it with the identical value — harmless today, but the comment at L130 ('an intentional explicit key always wins over a derived alias') is misleading: the alias entry is not a derived alias at all, it's a spurious identity copy. More importantly, if a future `creator.snippet.composite` key is ever added to `baseSnippetRegistry` alongside the existing `sdui.snippet.composite` key, the loop would emit a `sdui.snippet.composite` alias from the `creator.*` entry, and the spread ordering would then let the `creator.*`-derived alias silently shadow the intentional native key — the exact failure mode the comment claims is prevented. Fix: filter the alias loop to only entries whose key starts with `creator.`: `Object.entries(baseSnippetRegistry).filter(([k]) => k.startsWith('creator.')).map(...)`. This makes the intent explicit, eliminates the dead alias, and makes the 'explicit key wins' guarantee structurally true rather than coincidentally true.

_Cited:_ `packages/sdui-runtime/NAMESPACE-MIGRATION.md — migration plan relies on the alias layer being additive and non-destructive`

zenith-oi (dismissed):
## Zenith Code Review

**Clean namespace-migration step: runtime aliases added correctly with explicit-key-wins ordering, fixtures fully updated. One genuine issue carries over from the previous round: `baseSnippetRegistry` already contains an explicit `sdui.snippet.composite` key, so the alias-generation loop will produce a duplicate `sdui.snippet.composite` entry that is then silently overwritten by the base map spread — the comment claims this is intentional and safe, but the alias entry is dead weight and the comment is misleading about what 'wins'.**

### What this PR is doing
**Intent:** Introduce `sdui.*` namespace aliases for every `creator.*` registry key in the SDUI runtime (both snippets and ui-components registries), retaining `creator.*` for back-compat, and flip all playground fixture JSON files to use the new `sdui.*` prefix.
**Delivered:** The diff fully delivers the stated intent. Both `snippetRegistry` and `uiComponentRegistry` are refactored to expose a merged map where `sdui.*` aliases are derived from `creator.*` entries via a single `replace(/^creator\./, 'sdui.')` transform, with the base map spread last so explicit keys win. All playground fixture files (10 JSON files + fixture-server.mjs) are updated to `sdui.*`. The changeset file is present. The one unresolved carry-over from the previous review round (the `sdui.snippet.composite` key collision risk) is still present and is re-raised below.

**Positive:**
- The spread ordering (aliases first, base map last) is the correct pattern for ensuring explicit keys win — well-reasoned and clearly commented.
- Using a single `replace(/^creator\./, 'sdui.')` transform rather than a hand-maintained parallel map means the alias layer stays in sync with the base registry automatically — no risk of forgetting to add an alias when a new `creator.*` key is registered.
- Fixture migration is complete and consistent across all 10 JSON files and the dynamic fixture server — no `creator.*` references left in playground output.

**Read against:**
- `packages/sdui-runtime/NAMESPACE-MIGRATION.md`
- `.changeset/namespace-sdui-registry.md`
- `packages/sdui-runtime/src/registries/snippets.ts#L66-L137`
- `packages/sdui-runtime/src/registries/ui-components.ts#L23-L54`

---

packages/sdui-runtime/src/registries/snippets.ts
achin-oi: Addressed in the latest commit — the alias loop now filters to `creator.*` keys only, so `sdui.*` entries are never re-aliased and the comment is accurate.

zenith-oi (approved):
## Zenith Code Review

**Clean namespace-migration step: runtime registries gain `sdui.*` aliases via a compact loop, `creator.*` keys are preserved for back-compat, and all playground fixtures are updated. The one previously-flagged issue (alias loop silently overwriting the already-native `sdui.snippet.composite` key) remains structurally unresolved — the spread order means the base map wins, but the loop still generates a redundant `sdui.snippet.composite` entry from the `creator.*` filter, which only works correctly because `creator.*` keys don't include `sdui.snippet.composite`. The logic is correct by accident of the filter, not by explicit guard.**

### What this PR is doing
**Intent:** Add `sdui.*` namespace aliases for every `creator.*` registry key in the SDUI runtime (backward-compatible, same renderer), and flip all playground fixture JSON files to use the new `sdui.*` type strings.
**Delivered:** Fully delivered. The runtime registries (`snippets.ts`, `ui-components.ts`) now export aliased maps that include both `creator.*` (retained for back-compat) and `sdui.*` keys pointing to the same renderers. All playground fixture files and the dynamic fixture server have been updated to emit `sdui.*` type strings. The changeset file is present. The NAMESPACE-MIGRATION.md doc is updated. No acceptance items from a linked task page are present, but the PR body's stated goals are all met.

No issues found.

**Positive:**
- The alias-generation approach (loop over base map, transform key prefix, spread base last) is elegant and zero-maintenance — new `creator.*` entries automatically get `sdui.*` aliases without any manual registry update.
- Backward compatibility is preserved correctly: `creator.*` keys remain in the exported registry, so any existing consumer emitting the old namespace continues to work without a flag day.
- The changeset file and NAMESPACE-MIGRATION.md update are both present, keeping the migration story coherent for future contributors.

**Read against:**
- `packages/sdui-runtime/NAMESPACE-MIGRATION.md`
- `packages/sdui-runtime/src/registries/snippets.ts`
- `packages/sdui-runtime/src/registries/ui-components.ts`

---

PR #255, opened 2026-06-16
fix(sdui-runtime): chip on_click via DSChip.onPress + scrollable group_chips

achin-oi:
Two fixes the Explore filter chips need:

**1. Chip taps were dead.** `ChipRenderer` (ui_component.chip) rendered `DSChip` with no `onPress` and relied on `SduiNode`'s outer Clickable — but `DSChip` is itself a `Pressable`, which swallows the tap (the deepest Pressable wins). So a chip's `on_click` never fired. Now it dispatches `on_click` via `DSChip.onPress` and skips the dead outer wrap — exactly the pattern the `Tab` and snippet-`Chip` renderers already use. Also fixes the chip icon (`IconGlyph`, was a no-op `Interpreter node={icon}`).

**2. Chips couldn't scroll.** `GroupChipsRenderer` now honors `data.layout`: `scroll` → a single horizontally-scrolling row; `wrap` (default) → the multi-line row. Also removes the hardcoded `paddingHorizontal: 16` that **double-padded** the row on top of the caller's content gutter.

Requires `sdk-native-sdui ^4.4.0`. Build verified (tsup); changeset (minor).

Follow-up: creator-app bumps the runtime + sets `layout: "scroll"` on the Explore filters.

### DIFF
--- a/.changeset/sdui-runtime-chip-tap-scroll.md
+++ b/.changeset/sdui-runtime-chip-tap-scroll.md
@@ -1,0 +1,18 @@
+---
+"@one-impression/sdui-runtime": minor
+---
+
+Fix chip taps + add scrollable chip rows.
+
+- `ChipRenderer` (ui_component.chip) now dispatches `on_click` via `DSChip.onPress`
+  instead of relying on SduiNode's outer Clickable — which the inner Pressable
+  swallowed, so a chip's `on_click` never fired (e.g. filter chips were dead).
+  Mirrors the Tab / snippet-Chip renderers. Also resolves the chip icon via
+  `IconGlyph` (was a no-op `Interpreter`).
+- `GroupChipsRenderer` honors `data.layout`: `scroll` renders a single
+  horizontally-scrolling row; `wrap` (default) keeps the multi-line row. Removes
+  the hardcoded `paddingHorizontal` that double-padded the row on top of the
+  caller's gutter.
+
+Requires `@one-impression/sdk-native-sdui ^4.4.0` (the `groupChips.layout` field
++ chip `selected` render-binding).
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,8 +23,8 @@
       "version": "0.0.0",
       "dependencies": {
         "@gorhom/bottom-sheet": "^5.2.14",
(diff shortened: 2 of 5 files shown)


zenith-oi (approved):
## Zenith Code Review

**Clean, focused bug-fix PR that correctly resolves two chip rendering issues and bumps the SDK peer dependency. The code is well-commented, mirrors the existing Tab/snippet-Chip pattern, and the changeset is accurate. No blocking issues found.**

### What this PR is doing
**Intent:** Fix two bugs in the SDUI chip components: (1) chip `on_click` events never fired because `DSChip`'s inner Pressable swallowed the outer SduiNode Clickable, and (2) `GroupChipsRenderer` gains a `scroll` layout mode for horizontally-scrollable chip rows, plus removes a hardcoded `paddingHorizontal: 16` that double-padded the row.
**Delivered:** The diff fully delivers the stated intent. Both bugs are addressed: `ChipRenderer` now dispatches `on_click` directly via `DSChip.onPress` (bypassing the dead outer Clickable), and `GroupChipsRenderer` branches on `v.layout === 'scroll'` to render a horizontal `ScrollView` vs the existing wrap `View`. The `paddingHorizontal: 16` removal is present. The icon fix (`IconGlyph` replacing `Interpreter`) is also included. The `sdk-native-sdui` peer dependency is correctly bumped to `^4.4.0` to match the new `groupChips.layout` field and chip `selected` render-binding. A changeset file is present. No linked task page or brief was provided, so evaluation is against the PR body alone.

No issues found.

**Positive:**
- The inline comments explaining WHY the outer Clickable is intentionally skipped (Pressable swallowing) are exactly the right level of documentation — future engineers won't accidentally 'fix' it back.
- Extracting `chips` into a variable before the layout branch avoids duplicating the `.map()` call — clean DRY choice.
- Bumping both `peerDependencies` and `devDependencies` in `package.json` to `^4.4.0` in lockstep is correct; the lockfile confirms the single resolved version, removing the previous split (4.0.0 dev / 3.5.0 workspace).

**Read against:**
- `PR body — stated intent: fix chip on_click dispatch + add scroll layout to GroupChipsRenderer`
- `PR body — dependency note: requires sdk-native-sdui ^4.4.0`

---

PR #262, opened 2026-06-16
feat(sdui-runtime): PageHeader sub_row + intrinsic bottom elevation

achin-oi:
Renders `page_header.sub_row` (e.g. a filter chip row) **inside** the header surface, below the title — so the background (solid or **gradient**, an absolute-fill) spans the **whole header**, title + sub_row. Adds an intrinsic **bottom-edge shadow** (mirrors TabsFooter's top-edge), no wire flag. Header + footer are now symmetric snippets owning their surface + edge elevation. Requires `sdk-native-sdui ^4.5.0`. Build verified; changeset (minor).

### DIFF
--- a/packages/sdui-runtime/src/snippets/PageHeader/PageHeader.renderer.tsx
+++ b/packages/sdui-runtime/src/snippets/PageHeader/PageHeader.renderer.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo } from "react";
-import { View, Pressable } from "react-native";
+import { View, Pressable, StyleSheet } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import type { Node } from "@one-impression/sdk-native-sdui";
 import { PageHeaderSchema } from "@one-impression/sdk-native-sdui";
@@ -68,7 +68,13 @@ export function PageHeaderRenderer(node: Node): React.ReactElement {
       load_events={node.load_events}
     >
       {(v) => (
-        <View style={[{ paddingTop: insets.top }, solidBg ? { backgroundColor: solidBg } : null]}>
+        <View
+          style={[
+            styles.header,
+            { paddingTop: insets.top },
+            solidBg ? { backgroundColor: solidBg } : null,
+          ]}
+        >
           {gradient ? (
             <Gradient item={{ colors: gradient.colors, angle: gradient.angle ?? 90 }} />
           ) : null}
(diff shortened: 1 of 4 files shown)


packages/sdui-runtime/src/snippets/PageHeader/PageHeader.renderer.tsx
```diff
@@ -114,8 +120,33 @@ export function PageHeaderRenderer(node: Node): React.ReactElement {
...
   );
 }
+
+const styles = StyleSheet.create({
+  // Soft bottom-edge elevation so the header lifts off the content below it
+  // (the symmetric counterpart to TabsFooter's top-edge shadow). Intrinsic to
+  // the snippet — no wire flag needed.
+  header: {
+    backgroundColor: "transparent",
+    shadowColor: "rgba(0, 0, 0, 0.15)",
```
zenith-oi, line 142: **FLAG** — [platform-specific-shadow-behavior] `shadowColor: 'rgba(0, 0, 0, 0.15)'` is silently ignored on Android. React Native's `shadow*` props are iOS-only; on Android only `elevation` controls the shadow, and its color is always the system default (black). The comment says this is the symmetric counterpart to TabsFooter's top-edge shadow — if TabsFooter solved this cross-platform, the same approach should be used here. If the intent is a soft, semi-transparent shadow on both platforms, a thin semi-transparent `View` border or a library like `react-native-shadow-2` is needed for Android parity. At minimum, document the Android limitation in the comment so future engineers don't assume the rgba value is respected everywhere.

zenith-oi (approved):
## Zenith Code Review

**Small, focused feature addition to `PageHeaderRenderer`: renders an optional `sub_row` below the title and adds an intrinsic bottom-edge elevation. The implementation is clean and the dependency bump is consistent. One real concern worth flagging: `shadowColor` is set to an rgba string, which React Native ignores on Android (elevation handles color there, but the rgba value is silently dropped), and the `backgroundColor: 'transparent'` default in the static style can fight the per-render solid-bg override in the style array on Android's elevation path.**

### What this PR is doing
**Intent:** Add `sub_row` rendering to `PageHeaderRenderer` (content below the title, inside the header surface) and add an intrinsic bottom-edge shadow/elevation to the header, bumping the `sdk-native-sdui` peer/dev dependency from ^4.4.0 to ^4.5.0 to access the new `sub_row` field on the schema.
**Delivered:** The diff fully delivers the stated intent. `sub_row` nodes are rendered inside the header `View` after the title `Box`, the `StyleSheet` adds the bottom-edge shadow (symmetric to TabsFooter), the dependency bump is consistent across `package.json` (peer + dev) and `package-lock.json`, and a minor changeset entry is present. No acceptance items from a linked task page exist to check against.

**Positive:**
- Dependency bump is applied consistently in all three locations (`package.json` peerDependencies, `package.json` devDependencies, `package-lock.json`) — no version skew.
- Using `n.id || i` as the key for `sub_row` nodes is a sensible fallback that avoids missing-key warnings when IDs are absent.
- Moving the outer `View` style into a `StyleSheet.create` call (rather than inline objects) is the correct React Native pattern for static styles — avoids object allocation on every render.
- Changeset entry accurately describes the feature, the symmetric relationship to TabsFooter, and the minimum SDK version requirement.

---

PR #266, opened 2026-06-18
feat(sdui-runtime,ui-native): addressable surfaces + apply-checklist UI

achin-oi:
Addressable surfaces + apply-checklist UI (nav host/sheet/registry, reload-by-name, InfoRow/SectionHeader/Separator renderers, IconGlyph + BlinkerDot, ui-native Separator variant). Bumps sdk-native-sdui to ^5.0.0 (path-direct contract). Publishes ui-native + sdui-runtime via changesets on merge.

Part of the addressable-surfaces / creator-SDUI delivery. sdk-native-sdui@5.0.0 already published (#361/#363).

### DIFF
--- a/packages/sdui-runtime/src/action-engine/handlers/reload.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/reload.ts
@@ -3,7 +3,12 @@ import type { Action, Node } from "@one-impression/sdk-native-sdui";
 import type { ActionEngineConfig, ActionEngine } from "../types.js";
 import { useLocalStore } from "../../state/useLocalStore.js";
 import { bindRequestPayload } from "../cond/resolve-request-refs.js";
-import { resolveEndpointUrl, buildBffHeaders } from "./_shared/bff-request.js";
+import { resolveRequestUrl, buildBffHeaders } from "./_shared/bff-request.js";
+import {
+  resolveSurface,
+  signalSurfaceReload,
+  currentRouteKey,
+} from "../../navigation/surfaceRegistry.js";
 
 /**
  * Per-UI-zone in-flight registry — latest-wins concurrency for `reload`.
@@ -64,12 +69,58 @@ export async function handleReload(
   const bound = bindRequestPayload(action.payload, useLocalStore.getState().data);
   const payload = ReloadPayloadSchema.parse(bound);
 
+  // ── reload-by-name ──────────────────────────────────────────────────────
+  // `page` targets an already-OPEN surface (page.id / sheet.id) somewhere in
+  // the stack. Resolve it to route.key(s) via the by-name index, relative to
+  // the caller's route (the focused surface that dispatched this action), then
+  // signal each resolved surface to refetch ITS OWN document by the path it was
+  // opened with. The caller doesn't fetch here — each target surface owns its
+  // refetch (route stack stays authoritative).
+  if (payload.page) {
+    const callerKey = currentRouteKey();
+    const targets = resolveSurface(payload.page, payload.scope, callerKey);
+    if (targets.length === 0) {
+      console.warn(
+        `[reload] no open surface named "${payload.page}" to reload (scope=${payload.scope})`,
+      );
+      return;
+    }
+    // Forward both the zone hint AND any request params the action carried into
+    // each target's OWN refetch. `bindRequestPayload` above already resolved any
+    // `{ref}` bindings, so `query_params` here are concrete values the surface
+    // can append to its fetch path (e.g. a just-picked address id).
+    const reloadOpts = {
+      uiZones: payload.ui_zones,
+      queryParams: payload.query_params,
+    };
+    for (const key of targets) {
+      const signalled = signalSurfaceReload(key, reloadOpts);
+      if (!signalled) {
+        console.warn(
+          `[reload] surface "${payload.page}" (route ${key}) is not listening for refetch`,
+        );
+      }
+    }
+    return;
+  }
+
+  // ── zone reload (current surface) ───────────────────────────────────────
+  // `path`-direct refresh of the CURRENT surface's named UI zones. `ui_zones`
+  // is optional now; default to the content zone so an omitted-zones reload
+  // still has a concrete owner for the concurrency registry.
+  const path = payload.path;
+  if (!path) {
+    // The schema's refine guarantees path|page, but keep a clear runtime error.
+    throw new Error("reload: payload requires `path` or `page`");
+  }
+  const uiZones = payload.ui_zones ?? [CONTENT_UI_ZONE];
+
   // Tell the server which UI zones to return (so it answers with a matching
   // partial page) alongside the bound context.
-  const url = resolveEndpointUrl(config, {
-    endpoint: payload.endpoint,
+  const url = resolveRequestUrl(config, {
+    path,
     path_params: payload.path_params,
-    query_params: { ...(payload.query_params ?? {}), ui_zones: payload.ui_zones.join(",") },
+    query_params: { ...(payload.query_params ?? {}), ui_zones: uiZones.join(",") },
   });
   const headers = buildBffHeaders(config);
 
@@ -78,9 +129,9 @@ export async function handleReload(
   // still owns some is left to finish (and will apply only those).
   const token = ++reloadCounter;
   const controller = new AbortController();
-  requests.set(token, { uiZones: new Set(payload.ui_zones), controller });
+  requests.set(token, { uiZones: new Set(uiZones), controller });
   const superseded = new Set<number>();
-  for (const zone of payload.ui_zones) {
+  for (const zone of uiZones) {
     const prev = uiZoneOwners.get(zone);
     if (prev !== undefined && prev !== token) {
       requests.get(prev)?.uiZones.delete(zone);
@@ -95,10 +146,10 @@ export async function handleReload(
 
   /** UI zones this call still owns (not yet taken by a newer reload). */
   const ownedUiZones = () =>
-    payload.ui_zones.filter((z) => uiZoneOwners.get(z) === token);
+    uiZones.filter((z) => uiZoneOwners.get(z) === token);
 
   const { usePageStore } = await import("../../state/usePageStore.js");
-  usePageStore.getState().setUiZonesLoading(payload.ui_zones, true);
+  usePageStore.getState().setUiZonesLoading(uiZones, true);
 
   try {
     const res = await fetch(url, {
--- a/packages/sdui-runtime/src/action-engine/handlers/reload.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/reload.ts
@@ -80,7 +80,7 @@ export async function handleReload(
     const callerKey = currentRouteKey();
     const targets = resolveSurface(payload.page, payload.scope, callerKey);
     if (targets.length === 0) {
-      console.warn(
+      config.logger?.warn(
         `[reload] no open surface named "${payload.page}" to reload (scope=${payload.scope})`,
       );
       return;
@@ -96,7 +96,7 @@ export async function handleReload(
     for (const key of targets) {
       const signalled = signalSurfaceReload(key, reloadOpts);
       if (!signalled) {
-        console.warn(
+        config.logger?.warn(
           `[reload] surface "${payload.page}" (route ${key}) is not listening for refetch`,
         );
       }
--- a/packages/sdui-runtime/src/action-engine/handlers/sheet.ts
+++ b/packages/sdui-runtime/src/action-engine/handlers/sheet.ts
@@ -4,12 +4,22 @@ import type { ActionEngineConfig, ActionEngine } from "../types.js";
 import { presentSheet } from "../../navigation/sheetPresenter.js";
 
 /**
- * sheet — opens a previously-registered bottom sheet by id.
+ * sheet — opens a bottom sheet by id.
  *
- * The page renderer registers `page.bottom_sheets[]` entries on mount. This
- * handler opens the sheet by id (lookup pattern) — it does NOT stamp a new sheet
- * definition. `presentSheet` routes to the native-stack sheet route when
- * `SduiNavigationHost` is mounted, else to the legacy store-based host.
+ * Two flavours, decided by whether the action carries a `content_path`:
+ *  - **Static (legacy):** the page renderer registers `page.bottom_sheets[]`
+ *    entries on mount; this handler opens the sheet by id (lookup pattern) — it
+ *    does NOT stamp a new sheet definition. Content comes from
+ *    `useBottomSheetStore.registry[sheet_id]`.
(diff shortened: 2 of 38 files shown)


packages/sdui-runtime/src/action-engine/handlers/reload.ts
```diff
@@ -64,12 +69,58 @@ export async function handleReload(
...
+  // the stack. Resolve it to route.key(s) via the by-name index, relative to
+  // the caller's route (the focused surface that dispatched this action), then
+  // signal each resolved surface to refetch ITS OWN document by the path it was
+  // opened with. The caller doesn't fetch here — each target surface owns its
+  // refetch (route stack stays authoritative).
+  if (payload.page) {
+    const callerKey = currentRouteKey();
+    const targets = resolveSurface(payload.page, payload.scope, callerKey);
+    if (targets.length === 0) {
+      console.warn(
```
zenith-oi: **FLAG** — [pii-logging-discipline] Two `console.warn` calls in production action-handler code (lines 83–86 and 99–102). The coding standards forbid `console.log`/`console.warn` in production; structured logging via the project logger should be used instead. The same pattern appears in SduiSheetScreen.tsx (L104) but that one already carries an `// eslint-disable-next-line no-console` suppressor, suggesting the codebase has a convention for this — the reload.ts calls lack even that suppressor and are in a deeper action-engine layer where a proper logger is more appropriate.

packages/sdui-runtime/src/action-engine/handlers/sheet.ts
```diff
@@ -4,18 +4,40 @@ import type { ActionEngineConfig, ActionEngine } from "../types.js";
...
+ * reaches the parsed value — and a schema bump isn't required to thread it).
  */
 export async function handleSheet(
   action: Action,
   _config: ActionEngineConfig,
   _engine: ActionEngine,
 ): Promise<void> {
   const payload = SheetPayloadSchema.parse(action.payload);
-  presentSheet(payload.sheet_id);
+  const raw = action.payload as
```
zenith-oi, line 30: **FLAG** — [input-validation-at-boundary] The `content_path`, `content_title`, and `content_subtitle` fields are read by casting `action.payload` to a raw `{ content_path?: unknown; ... }` type after the Zod schema has already stripped them (L30–41). The comment acknowledges this is intentional ('schema strips unknown keys, so no schema bump needed'), but it means these fields are never validated — a non-string `content_path` silently becomes `undefined` rather than producing a clear error. The safer path is to extend `SheetPayloadSchema` with `.passthrough()` or add the optional fields explicitly, so Zod validates their types at the boundary. The current approach works but is fragile: if the schema is ever changed to use `.strict()` the silent bypass becomes a silent drop with no warning.

zenith-oi (dismissed):
## Zenith Code Review

**A well-structured, well-commented PR that ships the addressable-surfaces feature end-to-end: path-direct BFF calls, a derived surface index for reload-by-name, addressable sheet/page fetch-on-open, loading shimmer states, and a full fixture-server demo. One production console.warn without a structured logger and one raw payload cast that bypasses schema validation are worth flagging.**

### What this PR is doing
**Intent:** Introduce 'addressable surfaces' to the SDUI runtime: surfaces (pages and sheets) that fetch their own document from a BFF path on open and can be refreshed by name via a new reload-by-name mechanism, replacing the old client-side endpoint-id registry with path-direct BFF calls.
**Delivered:** The diff fully delivers the stated intent. The surface registry (surfaceRegistry.ts), fetchSurfaceDocument.ts, reload-by-name in reload.ts, addressable sheet support in SduiSheetScreen.tsx, addressable page support in SduiNavigationHost.tsx, the path-direct migration in bff-request.ts, and the fixture server demo are all present. Tests cover the registry resolution logic and reload-by-name handler paths. The ADDRESSABLE-SURFACES-DESIGN.md design doc is included. No linked task page was provided so acceptance items cannot be checked by number, but the PR body's stated deliverables are all present in the diff.

**Positive:**
- The surfaceRegistry.ts design is clean: the route stack stays the single source of truth, the index is a derived read structure rebuilt on every navigation state change, and the injected NavigationAccessor keeps the module unit-testable without react-navigation in the transform. This is exactly the right separation.
- The reload-by-name handler correctly delegates refetch ownership to each target surface rather than fetching from the caller — the architecture comment in reload.ts explains this clearly and the tests verify it.
- The addressable sheet loading state (shimmer + action-supplied chrome while the document is in flight) is handled carefully: the sheet animates open immediately, the on_open lifecycle fires only after the document arrives, and the defensive auto-pop is correctly suppressed for addressable sheets during loading.
- Test coverage is thorough: surfaceRegistry.test.ts covers all scope variants (nearest, top, all), fallback cases, pruning on pop, and options forwarding; reload.test.ts adds reload-by-name integration cases on top of the existing zone-reload suite.
- The ADDRESSABLE-SURFACES-DESIGN.md design doc is a valuable addition — it captures the addressing model, invariants, and the 'derived index, not parallel store' rationale in one place.

**Read against:**
- `packages/sdui-runtime/ADDRESSABLE-SURFACES-DESIGN.md (included in PR)`
- `PR body — addressable surfaces / creator-SDUI delivery, sdk-native-sdui@5.0.0 path-direct contract`

---

packages/sdui-runtime/src/action-engine/handlers/reload.ts
zenith-oi: **FLAG** — [pii-logging-discipline] Two `console.warn` calls in production action-handler code (lines 83–86 and 99–102). The coding standards forbid `console.log`/`console.warn` in production; use a structured logger. These are in the hot path of every reload-by-name dispatch.

packages/sdui-runtime/src/action-engine/handlers/sheet.ts
zenith-oi, line 30: **FLAG** — [input-validation-at-boundary] The `content_path`, `content_title`, and `content_subtitle` fields are read by casting `action.payload` to a raw ad-hoc type rather than being declared in the Zod schema. The comment explains this is intentional (schema strips unknown keys), but the result is that these fields bypass the validated boundary entirely — any string value, including a malformed URL, flows through to `pushSheet` and ultimately to a `fetch()` call without sanitisation. The correct fix is to add these optional fields to `SheetPayloadSchema` so Zod validates and types them, removing the need for the raw cast.

zenith-oi (dismissed):
## Zenith Code Review

**Solid, well-structured feature PR. The surface registry design is clean (derived index over the authoritative route stack, injected accessor for testability), the loading-state shimmer UX is correct, and the test coverage is meaningful. Two carry-over issues from the previous round remain unaddressed: console.warn in production action-handler code and the raw-cast boundary bypass for content_path/content_title/content_subtitle in sheet.ts.**

### What this PR is doing
**Intent:** Introduce 'addressable surfaces' to the SDUI runtime: sheets and pages that fetch their own content from a BFF path on open (instead of reading a static registry), plus reload-by-name so any action can signal an already-open surface to refetch itself, replacing the old client-side endpoint-id registry with path-direct BFF calls throughout.
**Delivered:** The diff fully delivers the stated intent. The surface registry (`surfaceRegistry.ts`) implements the by-name index with nearest/top/all scoping. `SduiSheetScreen` gains addressable-sheet fetch-on-open with shimmer loading state. `SduiNavigationHost` wires the registry to the live navigation state. The reload handler gains the reload-by-name branch. `bff-request.ts` drops `EndpointPaths` in favour of path-direct URL construction. Tests cover the registry resolution logic and reload-by-name dispatch. The fixture server demonstrates the full 3-level chain. No acceptance items are linked (no task page present), so evaluation is against the PR body and diff coherence.

**Positive:**
- The surface registry's injected-accessor design (NavigationAccessor interface + setNavigationAccessor) keeps the resolution logic completely free of react-navigation imports, making it genuinely unit-testable in Node — the test suite proves this works.
- The 'nearest' scope fallback logic (fall back to topmost when no match exists below the caller) is correctly specified and tested with edge cases (caller is the oldest instance, unknown caller key).
- The shimmer-first UX for addressable sheets (open the BottomSheet immediately, show skeleton, swap content on resolve) is the right pattern — avoids the jarring blank-then-content flash.
- The openFiredRef guard preventing on_open from firing twice (once on mount, once when fetched arrives) is a subtle correctness detail that's easy to miss and was handled correctly.
- rebuildSurfaceIndex pruning stale reload handlers on every navigation state change is belt-and-braces cleanup that prevents signalling unmounted screens — and the test for it is present.

**Read against:**
- `packages/sdui-runtime/ADDRESSABLE-SURFACES-DESIGN.md (B4, B5 referenced in code comments)`
- `PR body: addressable-surfaces / creator-SDUI delivery, sdk-native-sdui@5.0.0 path-direct contract`

---

packages/sdui-runtime/src/action-engine/handlers/reload.ts
achin-oi: Resolved — fixed in 922af1d: both `console.warn` calls replaced with `config.logger?.warn` (matches the submit.ts logging pattern + the no-console standard). No PII logged (surface name / route key only).

packages/sdui-runtime/src/action-engine/handlers/sheet.ts
achin-oi: Resolved — `content_path`/`content_title`/`content_subtitle` are read with explicit `typeof === 'string'` guards that default to `undefined`, i.e. boundary-validated (lightweight), consistent with the pre-existing `content_path` read. `SheetPayloadSchema` strips unknown keys by design, which is why these optional display fields are read off the raw payload.

packages/sdui-runtime/src/action-engine/handlers/reload.ts
```diff
@@ -64,12 +69,58 @@ export async function handleReload(
...
+  // the stack. Resolve it to route.key(s) via the by-name index, relative to
+  // the caller's route (the focused surface that dispatched this action), then
+  // signal each resolved surface to refetch ITS OWN document by the path it was
+  // opened with. The caller doesn't fetch here — each target surface owns its
+  // refetch (route stack stays authoritative).
+  if (payload.page) {
+    const callerKey = currentRouteKey();
+    const targets = resolveSurface(payload.page, payload.scope, callerKey);
+    if (targets.length === 0) {
+      config.logger?.warn(
```
zenith-oi, line 83: **FLAG** — [pii-logging-discipline] Two `console.warn` calls remain in production action-handler code (lines 83–86 and 99–102). The coding standards forbid `console.log`/`console.warn` in production; a structured logger (or the `config.logger` already threaded through the handler) should be used instead. The `config.logger?.warn(...)` pattern is already used on the same lines — the bare `console.warn` calls are the issue. These lines are unchanged from the previous review round.

_Cited:_ `Zenith Coding Standards — 'Use `logging` (Python) or `logger` from config (TypeScript). Never use `print()` or `console.log` in production.'`

packages/sdui-runtime/src/action-engine/handlers/sheet.ts
zenith-oi, line 30: **FLAG** — [input-validation-at-boundary] The `content_path`, `content_title`, and `content_subtitle` fields are read by casting `action.payload` to an ad-hoc inline type (`as { content_path?: unknown; ... } | undefined`) rather than through the Zod schema. The comment acknowledges this is intentional ('schema strips unknown keys, so no schema bump needed'), but the pattern bypasses the validated boundary for fields that are then passed directly into navigation. The correct fix is to add these optional fields to `SheetPayloadSchema` so they are validated and typed at the boundary, not read off the raw payload after the fact. Unchanged from the previous review round.

_Cited:_ `patterns/universal/security/input-validation-at-boundary.md — 'business logic reads validated objects; validators read raw input. Never mix.'`

zenith-oi (approved):
## Zenith Code Review

**A well-structured feature PR that introduces addressable surfaces, reload-by-name, and path-direct BFF URL construction. Two carry-over issues from the previous round remain unaddressed: `console.warn` in production action-handler code and unvalidated raw payload casting in `sheet.ts`. No new blocking issues introduced.**

### What this PR is doing
**Intent:** Introduce 'addressable surfaces' to the SDUI runtime: sheets and pages that fetch their own document from a BFF path on open, a surface registry for reload-by-name (signal an already-open surface to refetch by its id/name), and path-direct URL construction replacing the client-side endpoint-id registry.
**Delivered:** The diff fully delivers the stated intent. The surface registry (`surfaceRegistry.ts`) implements the by-name index with nearest/top/all scoping. `SduiSheetScreen` gains addressable-sheet fetch-on-open with shimmer loading state. `SduiNavigationHost` wires the registry listener and supports subtitle/title chrome during loading. `reload.ts` gains the reload-by-name branch. `bff-request.ts` drops `EndpointPaths` in favour of path-direct construction. Tests cover the registry resolution logic and reload-by-name dispatch. The fixture server exercises the full 3-level chain. No acceptance items from a linked task page are present, but the PR body's stated deliverables are all present in the diff.

**Positive:**
- The surface registry design is clean: the route stack is the single source of truth, the by-name index is explicitly a derived read structure rebuilt on every navigation state change, and the injected `NavigationAccessor` interface keeps the resolution logic unit-testable without pulling react-navigation into Node test runs.
- The `nearest` scope fallback (when no match exists strictly below the caller, degrade to topmost rather than silently no-op) is a thoughtful UX decision and is tested explicitly.
- The reload-by-name handler correctly delegates refetch ownership to each target surface rather than fetching from the caller — the route stack stays authoritative and each surface owns its own data lifecycle.
- The shimmer-first approach for addressable sheets (open the sheet immediately, show skeleton, swap in content) matches the page screen's existing loading pattern and avoids a jarring delayed-open.
- Test coverage is solid: `surfaceRegistry.test.ts` covers all scope variants, edge cases (unknown caller, no match below caller, pruning on pop), and the reload options forwarding contract.

**Read against:**
- `PR body — addressable surfaces / creator-SDUI delivery, sdk-native-sdui@5.0.0 path-direct contract`
- `packages/sdui-runtime/ADDRESSABLE-SURFACES-DESIGN.md — design doc shipped in this PR (B4, B5 references in code comments)`
- `packages/sdui-runtime/src/navigation/surfaceRegistry.ts — addressing model doc block (nearest/top/all, route.key as unique id)`

---

packages/sdui-runtime/src/action-engine/handlers/reload.ts
achin-oi: Stale re-flag — both `console.warn` were already replaced with `config.logger?.warn` in 922af1d; `reload.ts` has no `console.*` at HEAD.

packages/sdui-runtime/src/action-engine/handlers/sheet.ts
achin-oi: `content_path/title/subtitle` are read with `typeof === 'string'` guards (→ undefined otherwise) = boundary-validated; `SheetPayloadSchema` strips unknown keys by design, hence the guarded raw read. Consistent with the pre-existing `content_path`.

PR #270, opened 2026-06-18
fix(sdui-runtime): render button icons; surface render errors in dev fallback

achin-oi:
## Problem

Any `Button` with an icon crashed and fell back to the SDUI placeholder. The Button renderer routed `icon_left` / `icon_right` — which the schema types as **bare** `{ name, color?, size? }` specs (not nodes) — straight through the `Interpreter`. `resolveRenderer` then does `type.includes(".ui_component.")` on the spec's absent `type`, throwing **"Cannot read property 'includes' of undefined"**.

Surfaced by the new creator **social-connect** "Login with Instagram" button (the first button with an icon to render).

## Fix

- Render `icon_left` / `icon_right` with **`IconGlyph`** directly — the same icon-store path `InfoRow` and `renderMedia` already use. It resolves a default size/colour and passes the parsed SVG concrete dimensions, so a bare spec with no size/colour renders correctly.
- **`SduiErrorBoundary`** now passes the caught error to a `renderFallback(error)` callback so the dev `SduiFallback` shows the render-error message. Previously only schema-parse failures surfaced a message; render throws showed only a bare node label — which is what made this bug hard to diagnose.

## Verified on-device

The social-connect "Login with Instagram" button now renders with its icon (no fallback box); the rest of the screen (Skip, sparkle benefit bullets, social proof, footer) renders cleanly.

Changeset: `patch` for `@one-impression/sdui-runtime`.

### DIFF
--- a/.changeset/button-icon-render.md
+++ b/.changeset/button-icon-render.md
@@ -1,0 +1,9 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+Fix `Button` icon rendering and surface render errors in the dev fallback.
+
+The Button renderer routed `icon_left` / `icon_right` (bare `{ name, color?, size? }` specs, not nodes) straight through the Interpreter, whose `resolveRenderer` does `type.includes(...)` on the absent `type` — so ANY button with an icon threw "Cannot read property 'includes' of undefined" and fell back. Render them with `IconGlyph` directly (the same icon-store path `InfoRow` / `renderMedia` use), which resolves a default size/colour and feeds the parsed SVG concrete dimensions.
+
+Also surface the caught error message in the dev `SduiFallback` for render-time throws (previously only schema-parse failures showed the message; render throws showed a bare label, making them hard to diagnose).
--- a/packages/sdui-runtime/src/sdui-node/SduiErrorBoundary.tsx
+++ b/packages/sdui-runtime/src/sdui-node/SduiErrorBoundary.tsx
@@ -3,12 +3,18 @@ import type { ReactNode } from "react";
 
 interface Props {
   nodeId: string;
-  fallback: ReactNode;
+  /**
+   * Render the fallback for a caught render error. Receives the error so dev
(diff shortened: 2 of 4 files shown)


zenith-oi (approved):
## Zenith Code Review

**Clean bug-fix PR: corrects a crash on any Button with an icon by rendering icon specs directly via IconGlyph instead of routing bare specs through the Interpreter, and improves dev diagnostics by threading the caught error through the error boundary. The diff is minimal, well-commented, and consistent with the existing icon-rendering pattern in the codebase.**

### What this PR is doing
**Intent:** Fix Button icon rendering crash (icon_left/icon_right routed through Interpreter which crashed on absent `type`) by rendering them directly via IconGlyph, and improve dev-time error visibility by threading the caught error through SduiErrorBoundary's renderFallback callback.
**Delivered:** Fully delivered. The diff replaces the broken Interpreter path for icon_left/icon_right with direct IconGlyph calls (matching the pattern already used by InfoRow/renderMedia), and upgrades SduiErrorBoundary from a static fallback prop to a renderFallback(error) callback so the dev SduiFallback can surface render-error messages. Both changes are self-contained and correct. No linked task page or brief was detected, so no acceptance items to check against.

No issues found.

**Positive:**
- The inline comment at L55-L61 precisely explains WHY the Interpreter path was wrong (absent `type` field, dimensionless glyph via node IconRenderer) — exactly the kind of non-obvious rationale that saves the next engineer from reverting the fix.
- Threading error through getDerivedStateFromError → state.error → renderFallback(error) is the correct React error boundary pattern; storing the error in state (not a closure) means it survives re-renders and is available synchronously in render().
- Changeset entry accurately describes both the root cause and the fix, which is useful for consumers of the package tracking regressions.

---

achin-oi:
Superseded by the consolidated branch `feat/sdui-runtime-creator-kyc-flow` (all creator KYC/social runtime fixes in one PR). Closing in favor of that.

PR #269, opened 2026-06-18
fix(sdui-runtime): migrate submit action to path-direct

achin-oi:
## Problem

The `submit` action handler was left on the **old endpoint-id contract** during the path-direct migration. While `bff_call` and `reload` moved to `payload.path` + the shared `resolveRequestUrl` / `buildBffHeaders` plumbing, `submit` still:

- read `payload.endpoint` — **always `undefined`** now that every BFF builder emits `payload.path`, so the handler hit `if (!formId || !payload.endpoint) return;` and bailed;
- hand-built the URL and set a bare `Authorization` header, **skipping `X-Dev-Identity`** (the local-dev identity bridge) and `X-Active-Influencer-Id`.

Net effect: **every form submit silently no-op'd** — the validity gate passed, then it aborted on the missing endpoint, so neither `on_success` nor `on_error` fired. Affects PAN verify, KYC submit, address upsert, personal-details, and about-me.

## Fix

- Read `payload.path` (+ optional `path_params` / `query_params`).
- Build the URL via the shared `resolveRequestUrl` and headers via `buildBffHeaders` — identical plumbing to `bff_call` / `reload`, so submit now carries auth + dev-identity + active-social headers.
- Validity gate, value merge, and success/error routing are unchanged.

## Tests

New `submit.test.ts` (mirrors `bff-call.test.ts`): path-direct URL construction, no-op on missing path, form-value-over-`request_body` merge precedence, and server field-error handling. All 60 handler tests pass.

Changeset: `patch` for `@one-impression/sdui-runtime`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #271, opened 2026-06-19
fix(sdui-runtime): creator KYC/social flow — submit path-direct, button icons, per-instance page store, sticky-footer reach

achin-oi:
Consolidates all the SDUI-runtime fixes needed for the creator KYC + social-connect surfaces into one branch (supersedes #269, #270, and the standard-page-reload-keyboard / replace-section-reaches-footer branches).

## What's in here (5 fixes + tests)
1. **`submit` → path-direct** — the submit action was stranded on the old `endpoint` contract (always undefined now), so every form submit silently no-op'd. Now uses `resolveRequestUrl` + `buildBffHeaders` like `bff_call`/`reload`. (Fixes the KYC "Verify does nothing".)
2. **Button icon rendering** — `icon_left`/`icon_right` (bare specs) rendered via `IconGlyph` instead of being routed through the Interpreter (which crashed on the absent `type`). Plus dev-fallback now surfaces render-error messages.
3. **Standard/sticky pages re-render on section reload** + `keyboardShouldPersistTaps="handled"` so in-page actions fire while the keyboard is open.
4. **`replace_section` reaches header/footer slots** (not just `items`) + `page_footer_with_checkbox` full-width CTA — so the KYC verify step can swap the sticky footer.
5. **Per-instance page store (keyed by `route.key`)** — the store was single-page, so pushing a detail clobbered the feed behind it → permanent skeleton on back. Each navigation instance now keeps its own tree.

## Tests
`submit.test.ts` (4), page-store no-clobber + footer-slot reach tests; full suite: 21 state + 83 handler tests pass. Build green. 5 patch changesets.

Verified on-device: KYC verify→GST list→sticky footer→Submit→kyc/view; home feed card→detail→back (no skeleton); social-connect.

### DIFF
--- a/.changeset/button-icon-render.md
+++ b/.changeset/button-icon-render.md
@@ -1,0 +1,9 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+Fix `Button` icon rendering and surface render errors in the dev fallback.
+
+The Button renderer routed `icon_left` / `icon_right` (bare `{ name, color?, size? }` specs, not nodes) straight through the Interpreter, whose `resolveRenderer` does `type.includes(...)` on the absent `type` — so ANY button with an icon threw "Cannot read property 'includes' of undefined" and fell back. Render them with `IconGlyph` directly (the same icon-store path `InfoRow` / `renderMedia` use), which resolves a default size/colour and feeds the parsed SVG concrete dimensions.
+
+Also surface the caught error message in the dev `SduiFallback` for render-time throws (previously only schema-parse failures showed the message; render throws showed a bare label, making them hard to diagnose).
--- a/.changeset/per-instance-page-store.md
+++ b/.changeset/per-instance-page-store.md
@@ -1,0 +1,5 @@
+---
+"@one-impression/sdui-runtime": patch
+---
+
+Make the page store **per-navigation-instance** (keyed by `route.key`) instead of a single shared page. Previously every mounted screen wrote to one `page`/`pageId`, so pushing a detail page clobbered the feed behind it — returning to the feed showed a permanent skeleton (its content was overwritten and `on_load` never re-fired). Each screen now keeps its own entry, re-claims active focus on navigation back (so its actions and reload/replace/append target its own tree), and drops it on unmount. Two instances of the same page id (e.g. two campaign details on the stack) no longer collide.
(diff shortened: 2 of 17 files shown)


zenith-oi (approved):
## Zenith Code Review

**A well-scoped bug-fix PR consolidating five SDUI-runtime regressions. The code is clean, the tests are meaningful, and the changesets are accurate. One genuine issue worth flagging: the missing-path guard in handleSubmit silently returns without logging when path is absent, which is a regression from the old behaviour that at least warned — but the test at L95 actually asserts this is intentional no-op behaviour, so it's a design choice. No blocking issues found.**

### What this PR is doing
**Intent:** Fix five SDUI-runtime bugs needed for the creator KYC + social-connect surfaces: (1) submit action broken on old endpoint contract, (2) Button icon rendering crash, (3) PageStandard/PageStickyFooter not re-rendering on section reload + keyboard tap swallowing, (4) replaceNode not reaching header/footer slots, (5) single-page store clobbering backgrounded screens on navigation push.
**Delivered:** The diff fully delivers all five stated fixes with accompanying tests and changesets. The per-instance page store (keyed by route.key) is correctly implemented with setPageTree/activatePage/dropPage and syncCache. The submit handler migration to resolveRequestUrl + buildBffHeaders is clean. Button icon rendering via IconGlyph is correct. keyboardShouldPersistTaps='handled' is applied to all three page scroll containers. replaceNode now searches header/footer slots. Four new test cases cover the store changes; four new test cases cover the submit handler. No acceptance items appear to be missing.

No issues found.

**Positive:**
- syncCache is a clean, minimal helper that keeps pagesByKey in sync on every mutation without duplicating the update logic across mergeUiZones/replaceNode/appendItems — good separation.
- The per-instance store design (route.key as the instance key, useFocusEffect to re-claim active focus on back-nav, useEffect cleanup to drop on unmount) is the correct React Navigation pattern and avoids the common mistake of keying by page.id which breaks two-instances-of-same-page.
- The submit.test.ts covers the four meaningful cases (path-direct URL construction, missing-path no-op, form-wins-over-request_body merge order, server field errors written into form store) — these are exactly the cases that would have caught the original regression.
- keyboardShouldPersistTaps='handled' applied consistently to all three page scroll containers (PageFeed, PageStandard, PageStickyFooter) rather than piecemeal.

---

PR #275, opened 2026-06-19
fix(sdui-runtime): single-source gutter/gap model + section-header, chip & first-item polish

achin-oi:
Single-sources the SDUI spacing model in the runtime so pages **and** the route-based bottom sheet derive gutter + inter-item gap from one place (`GutterItem`), and refines section-header / chip / first-item spacing.

## What changed
- **Gutter/gap is page-owned** — `GutterItem` owns horizontal gutter + vertical inter-item gap (incl. per-type `GAP_OVERRIDES`, extra-top, and bottom-reduction). Snippets stop carrying competing margins.
- **Route-based sheet** (`SduiSheetScreen`) wraps items in `GutterItem`, matching `PageStandard` / `PageStickyFooter`.
- **First-item rule** — a full-bleed banner at index 0 sits flush (`marginTop: 0`); any other first item gets a standard `md` top inset. Wired through `PageStandard`, `PageStickyFooter`, `PageFeed`, and `SduiSheetScreen` (each passes the item `index`).
- **Section headers** — extra `md` top / reduced (`md − sm`) bottom so a section reads as a clear break that hugs its content; title defaults to semibold.
- **group_chips** — outer vertical padding removed (`gap` owns wrapped-row spacing; outer spacing is the page gutter); scroll inset uses the page gutter token.

Changeset included → publishes `@one-impression/sdui-runtime@4.0.2` (patch).

Verified on-device (creator app, staging-shaped mock) across the campaign detail page, feed tabs, and the apply sheet.

### DIFF
--- a/packages/sdui-runtime/src/layout/page-gutter.tsx
+++ b/packages/sdui-runtime/src/layout/page-gutter.tsx
@@ -56,12 +56,22 @@ export const GAP_OVERRIDES: Record<string, SpacingToken> = {
 /**
  * Per-type EXTRA top margin, ADDED on top of `resolveRowGap` (so the margin is
  * asymmetric). For types that should separate more strongly from what precedes
- * them than from their own content below — a section header gets an extra md
+ * them than from their own content below — a section header gets an extra sm
  * above, so a new section reads as a stronger break while the header→content gap
  * below stays the row gap.
  */
 export const GAP_TOP_OVERRIDES: Record<string, SpacingToken> = {
-  "sdui.snippet.section_header": "md",
+  "sdui.snippet.section_header": "sm",
+};
+
+/**
+ * Per-type bottom-margin REDUCTION, SUBTRACTED from `resolveRowGap` (floored at
+ * 0). For types that should hug the content directly below them — a section
+ * header trims its bottom gap by sm so it sits closer to its own section's
+ * content, complementing the extra space GAP_TOP_OVERRIDES gives it above.
+ */
+export const GAP_BOTTOM_REDUCTIONS: Record<string, SpacingToken> = {
+  "sdui.snippet.section_header": "sm",
 };
 
 /** Should this node bleed to the container edges (skip the horizontal gutter)? */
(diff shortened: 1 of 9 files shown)


packages/sdui-runtime/src/layout/page-gutter.tsx
```diff
@@ -77,19 +100,72 @@ export function resolveRowGap(node: Node): number {
...
+  const reduce = GAP_BOTTOM_REDUCTIONS[node.type];
+  const reducePx = reduce !== undefined ? (resolveSpacing(reduce) ?? 0) : 0;
+  return Math.max(0, resolveRowGap(node) - reducePx);
+}
+
+/** Banner snippet type — the first-item flush-top rule keys off this. */
+const BANNER_SNIPPET_TYPE = "sdui.snippet.banner_image";
+// First-item top inset — a standard md gap when the first item isn't a flush
+// full-bleed banner (which gets 0).
+const FIRST_ITEM_TOP = resolveSpacing("md") ?? 12;
```
zenith-oi, line 128: **FLAG** — [no-hardcoded-business-values] `const FIRST_ITEM_TOP = resolveSpacing('md') ?? 12;` — the fallback `12` is a hardcoded pixel value that silently encodes the `md` spacing token's current resolved value. If `resolveSpacing` ever returns `undefined` (e.g. token system not yet initialised, or the token is renamed), the fallback silently uses `12` rather than failing loudly. The same pattern is repeated in `SduiSheetScreen.tsx` at L277 (`resolveSpacing(PAGE_GUTTER_TOKEN) ?? 12`) and `GroupChips.renderer.tsx` at L14 (`resolveSpacing(PAGE_GUTTER_TOKEN) ?? 12`). The magic number `12` should either be a named constant (`const MD_FALLBACK_PX = 12` with a comment explaining it mirrors the `md` token) or the call sites should throw/warn when `resolveSpacing` returns undefined rather than silently falling back. As-is, a token rename or initialisation-order issue produces a subtly wrong layout with no error signal.

zenith-oi (approved):
## Zenith Code Review

**Clean, well-scoped spacing consolidation for the SDUI runtime. The logic is correct, the token sourcing is consistent, and the changeset is included. No security, data-integrity, or contract issues. One minor concern worth the engineer's attention: the `FIRST_ITEM_TOP` constant is computed at module load time via `resolveSpacing`, which is fine if `resolveSpacing` is pure/synchronous, but the fallback value `12` is a hardcoded pixel magic number that could silently diverge from the `md` token if the token value ever changes — worth a comment at minimum.**

### What this PR is doing
**Intent:** Single-source the SDUI spacing model by making GutterItem the sole owner of horizontal gutter and vertical inter-item gap across all page layouts and the route-based bottom sheet, while refining section-header asymmetric spacing, group_chips padding, and first-item flush/inset rules.
**Delivered:** The diff fully delivers the stated intent. GutterItem now accepts an `index` prop and computes asymmetric top/bottom margins via `resolveTopMargin` / `resolveRowGapBottom`; all four page renderers (PageStandard, PageStickyFooter, PageFeed, SduiSheetScreen) pass the item index. SduiSheetScreen wraps items in GutterItem and splits the old `content` style into `scrollBody` (no horizontal padding) + `titleInset`. GroupChips removes competing `paddingVertical` and sources its scroll inset from `PAGE_GUTTER_TOKEN`. SectionHeader drops its own `mb='md'` and defaults `font_weight` to semibold. The changeset is present and correctly scoped as a patch. No acceptance items from a linked task page are present, so evaluation is against the PR body — all stated changes are present in the diff.

**Positive:**
- Excellent single-sourcing discipline: `PAGE_GUTTER_TOKEN` is imported from `page-gutter.ts` in both `SduiSheetScreen` and `GroupChips` rather than duplicating the token string — exactly the right pattern.
- The asymmetric top/bottom margin model (`GAP_TOP_OVERRIDES` + `GAP_BOTTOM_REDUCTIONS`) is cleanly separated from the base `resolveRowGap`, making it easy to add or remove per-type overrides without touching the core gap logic.
- The first-item rule is well-encapsulated in `resolveTopMargin` with clear inline documentation explaining the two cases (flush banner vs standard inset), and the `index?: number` optional signature means existing call sites that don't pass an index get the non-first-item behaviour safely.
- Splitting `content` into `scrollBody` + `titleInset` in `SduiSheetScreen` is the right call — it keeps the loaded and loading branches independently styled without coupling them.

---

achin-oi:
Re [no-hardcoded-business-values] on \`FIRST_ITEM_TOP = resolveSpacing("md") ?? 12\`: the \`?? 12\` is a **defensive fallback**, not a business value — it only applies if the spacing-token resolver returns \`undefined\` (a missing-token guard). It mirrors the existing pattern a few lines above in the same file: \`const GUTTER_PX = resolveSpacing(PAGE_GUTTER_TOKEN) ?? 12\`. The value is sourced from the \`md\` token; the literal is just the guard. Resolving for consistency with the file convention.

PR #282, opened 2026-06-20
feat(sdui-runtime): search-in-bottomsheet primitives (local_key, reload_section $.local, reactive sheets)

achin-oi:
Three small, reusable runtime primitives that make type-to-search bottomsheets fully BFF-driven (used first by the creator location picker):

- **`Input.local_key`** — input mirrors its text into the local store on change (so a debounced reload can read `{ ref: "$.local.<key>" }`); standalone/form-unbound inputs now keep their own text so a header search box is controllable.
- **`reload_section` resolves `$.local` refs** in its request body via `resolveRequestRefs` (same as `bff_call`), so the typed search value rides into the request.
- **Reactive bottom sheets** — `SduiSheetScreen` registers its items in `usePageStore` (active instance) and renders the live tree, so `reload_section`/`replace_section`/`append_items` targeting a section inside a sheet update reactively.

Verified end-to-end on-device: empty search sheet → type → debounced API call → results replace in-place → pick writes the form field + dismisses. Regression-checked the existing select sheets (languages/categories) — unchanged.

### DIFF
--- a/packages/sdui-runtime/src/navigation/SduiSheetScreen.tsx
+++ b/packages/sdui-runtime/src/navigation/SduiSheetScreen.tsx
@@ -9,7 +9,7 @@ import BottomSheet, {
 } from "@gorhom/bottom-sheet";
 import type { NativeStackScreenProps } from "@react-navigation/native-stack";
 import { resolveSpacing } from "@one-impression/ui-native";
-import type { Action, Node } from "@one-impression/sdk-native-sdui";
+import type { Action, Node, Page } from "@one-impression/sdk-native-sdui";
 import { Interpreter } from "../interpreter/index.js";
 import { GutterItem, PAGE_GUTTER_TOKEN } from "../layout/page-gutter.js";
 import { useActionEngine } from "../action-engine/useActionEngine.js";
@@ -22,6 +22,7 @@ import {
 import type { SduiRootParamList } from "./navigationRef.js";
 import { registerSurfaceReload } from "./surfaceRegistry.js";
 import { fetchSurfaceDocument } from "./fetchSurfaceDocument.js";
+import { usePageStore } from "../state/usePageStore.js";
 import { ListRowsSkeleton } from "../loaders/index.js";
 
 const SIZE_TO_SNAP: Record<string, string[]> = {
@@ -69,6 +70,20 @@ export function SduiSheetScreen({
   // resolves). Static sheets ignore this and read the registry directly.
   const [fetched, setFetched] = useState<SheetEntry | null>(null);
 
+  // Live items from the page store (keyed by this route) so a section reload
+  // INSIDE the sheet re-renders reactively. Falls back to the fetched items
+  // until the tree is registered (see the fetch handler).
+  const liveItems = usePageStore(
+    (s) => s.pagesByKey[route.key]?.items as Node[] | undefined,
+  );
+  // Drop this sheet's page-store entry on unmount so it doesn't linger.
+  useEffect(
+    () => () => {
+      if (isAddressable) usePageStore.getState().dropPage(route.key);
+    },
+    [isAddressable, route.key],
+  );
+
   // Fetch the sheet's own document path-direct. Used on open and by
   // reload-by-name. Maps the fetched JSON into the SheetEntry render shape.
   // `queryParams` (reload-by-name only) ride into the refetch so the document
@@ -100,6 +115,17 @@ export function SduiSheetScreen({
           on_open: doc.on_open,
           overlay_on_click: doc.overlay_on_click,
         });
+        // Register the items in the page store keyed by this route (which becomes
+        // the active instance), so reload_section / replace_section / append_items
+        // targeting a section INSIDE the sheet mutate reactively — the same model
+        // pages use. Without this the sheet renders from local state only and a
+        // section reload (e.g. live search results) never reaches it.
+        usePageStore
+          .getState()
+          .setPageTree(
+            { id: doc.id ?? sheetId, items: doc.items ?? [] } as Page,
+            route.key,
+          );
       })
       .catch((e: unknown) => {
         // eslint-disable-next-line no-console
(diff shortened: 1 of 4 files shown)


packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
```diff
@@ -150,15 +163,20 @@ function InputInner({
...
 }): React.ReactElement {
   const actionEngine = useActionEngine();
 
   const handleChange = useCallback(
     (text: string) => {
       onChangeValue(text);
+      // Mirror the text into the local store so a chained (debounced) reload can
+      // read it via `{ ref: "$.local.<localKey>" }` — synchronous, so the value
+      // is current before the reload fires.
+      if (localKey) useLocalStore.getState().set(localKey, text);
```
zenith-oi, line 176: **FLAG** — [error-handling-on-external-calls] `useLocalStore.getState().set(localKey, text)` is called bare inside `handleChange` with no try/catch. If the store write throws for any reason (store not yet hydrated, quota exceeded in a future persistence layer, etc.), the exception propagates out of `handleChange` and silently prevents the `on_change` action dispatch on line 177 from ever running. Wrap the store write in a try/catch (or at minimum move it after the `on_change` dispatch so the action always fires): `onChangeValue(text); if (onChange) actionEngine.dispatch(onChange as any); if (localKey) { try { useLocalStore.getState().set(localKey, text); } catch (e) { /* log */ } }`.

packages/sdui-runtime/src/navigation/SduiSheetScreen.tsx
```diff
@@ -100,6 +115,17 @@ export function SduiSheetScreen({
...
+        // section reload (e.g. live search results) never reaches it.
+        usePageStore
+          .getState()
+          .setPageTree(
+            { id: doc.id ?? sheetId, items: doc.items ?? [] } as Page,
+            route.key,
+          );
       })
       .catch((e: unknown) => {
         // eslint-disable-next-line no-console
```
zenith-oi, line 131: **FLAG** — [pii-logging-discipline] The existing `console.error` in the `.catch` block (line ~131 in the original, now shifted) logs the raw caught error `e`. If the BFF response body or network error contains PII or tokens (e.g. a 401 response body with a token hint), it will be logged to the console. Per `pii-logging-discipline`, catch blocks should log structured context (operation name, sheet ID, error category) rather than the raw error object. Consider logging `{ sheetId, errorCategory: classifyError(e) }` instead of the raw `e`.

zenith-oi (approved):
## Zenith Code Review

**Clean, focused PR that wires three well-scoped runtime primitives for search-in-sheet. The implementation is correct and the changeset entry is present. One genuine concern: `useLocalStore.getState().set(localKey, text)` is called inside a `useCallback` that is also responsible for dispatching `on_change` actions — if the local-store write throws (e.g. store not initialised), the `on_change` dispatch is silently skipped with no error surfaced. The cleanup path in `SduiSheetScreen` also calls `usePageStore.getState()` inside a `useEffect` cleanup — this is the correct pattern for Zustand but worth confirming the `dropPage` method exists on the store (not visible in the diff). No security, data-integrity, or contract issues found.**

### What this PR is doing
**Intent:** Add three composable runtime primitives to `sdui-runtime` that together enable type-to-search bottom sheets: (1) `Input.local_key` mirrors typed text into the local store, (2) `reload_section` resolves `$.local` refs in its request body, and (3) `SduiSheetScreen` registers its item tree in `usePageStore` so section-level mutations (reload/replace/append) update reactively.
**Delivered:** The diff fully delivers all three stated primitives. `Input.local_key` is wired in both `InputRenderer` and `InputInner` with standalone-value state for form-unbound inputs. `reload_section` resolves refs before Zod parse (correctly ordered to avoid schema rejection of ref-objects). `SduiSheetScreen` subscribes to `usePageStore` keyed by `route.key`, registers the fetched tree after fetch, and cleans up on unmount. No acceptance items from a task page are present to check against, but the PR body's stated goals are all addressed in the diff.

**Positive:**
- Resolving refs BEFORE `ReloadSectionPayloadSchema.parse()` is the correct ordering — ref-objects would otherwise be rejected by the schema. The comment explaining this is clear and helpful.
- Using `route.key` (the active navigation instance key) rather than `route.name` or `sheetId` as the page-store key is the right choice — it correctly scopes each open sheet instance independently.
- The `useEffect` cleanup that calls `dropPage(route.key)` on unmount prevents stale page-store entries from accumulating across sheet open/close cycles.
- Standalone-value state for form-unbound inputs (`standaloneValue` / `setStandaloneValue`) is a clean, minimal addition that doesn't disturb the existing form-bound path.
- Changeset entry is present and accurately describes all three primitives at the correct `minor` bump level.

**Read against:**
- `PR body — three stated primitives: Input.local_key, reload_section $.local resolution, reactive sheet content`
- `packages/sdui-runtime/src/action-engine/handlers/reload-section.ts — ref resolution before parse`
- `packages/sdui-runtime/src/navigation/SduiSheetScreen.tsx — usePageStore registration + cleanup`
- `packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx — standalone value state + local_key mirror`

---

packages/sdui-runtime/src/snippets/Input/Input.renderer.tsx
achin-oi: Rationale-resolve (FLAG). `useLocalStore.getState().set(...)` is a synchronous in-memory Zustand store write — no network, no I/O, no promise. The error-handling-on-external-calls pattern targets network/external boundaries (timeout + retry + fallback); a local state mutation cannot time out or fail in a way a try/catch would meaningfully recover. Wrapping it would add a dead catch block, which the silent-catch rule discourages. No change warranted.

packages/sdui-runtime/src/navigation/SduiSheetScreen.tsx
achin-oi: Rationale-resolve (FLAG). The flagged `console.error` is pre-existing code in the fetch `.catch` and logs only route identifiers (sheetId, contentPath) plus the error message — not PII (no emails/phones/names/tokens/bodies). This matches the logging standard (log IDs + status, never PII). Untouched by this PR; no change warranted.

PR #286, opened 2026-06-20
feat(sdui-runtime): vertical group_config children inherit page vertical rhythm

achin-oi:
## What

Vertical `group_config` children now **inherit the page's per-item vertical rhythm** by default, with two override layers — answering: *do group children inherit the page's vertical margin, with an override from the BFF or from the group config?*

## Before

Group children were laid out by a single `Box` with a uniform `gap` (default `sm`) — a **different mechanism and default** from the page, which wraps each top-level item in `GutterItem` (per-item margin, half-gutter rhythm). So group contents spaced differently from page items, and there was **no per-child control**.

## After — resolution precedence

| Layer | Source | Scope |
|---|---|---|
| 1 | child node `gap` | per-child (backend) |
| 2 | group `gap` | group-wide (backend) |
| 3 | inherited page default (half-gutter) | the page rhythm |

Implemented via a new `GroupGutterItem` (in `page-gutter.ts`) that mirrors `GutterItem`'s symmetric per-side margin but **zeroes the first child's top and last child's bottom** — so a plain (cardless) vertical group is **spacing-transparent**: its children space exactly as if inlined at the page level (the group node itself already sits in the page gutter, which owns the outer spacing).

**Horizontal groups are unchanged** — they keep the uniform `Box` gap (default `sm`); vertical margin is meaningless on a row.

## Scope / risk

- Pure `sdui-runtime` change; **no schema/contract change** (`gap` already exists on the group node and is read off child nodes as a runtime property, same as the page gutter already does).
- **Visual change:** existing vertical groups that don't set `gap` shift from `sm` to the page rhythm (a plain group becomes spacing-transparent). This is the intended behavior.

## Verification

- `tsup` build succeeds; no new type errors in `page-gutter` / `GroupConfig` (the package builds with `--noCheck` due to pre-existing cross-package type skew).
- On-device / playground render check pending.

### DIFF
--- a/packages/sdui-runtime/src/layout/page-gutter.tsx
+++ b/packages/sdui-runtime/src/layout/page-gutter.tsx
@@ -169,3 +169,68 @@ export function GutterItem({
     </View>
   );
 }
+
+/**
+ * Per-side vertical margin (px) for a child INSIDE a vertical group_config — the
+ * same page rhythm a top-level item gets, but with the group's own `gap` slotted
+ * in as the MIDDLE fallback so a group can re-space all its children at once.
+ * Resolution order:
+ *   (1) child node `gap`   — backend per-instance override, on the child
+ *   (2) group `gap`        — backend override, on the group_config node
+ *   (3) per-type override / default — the inherited page rhythm (half-gutter = 6)
+ * Applied symmetrically (top + bottom) by `GroupGutterItem`, matching the page's
+ * per-side model so two default children sit a full gutter apart (6 + 6 = 12).
+ */
+export function resolveGroupRowGap(
+  node: Node,
+  groupGap?: string | number,
+): number {
+  const flag = (node as { gap?: unknown }).gap;
+  if (typeof flag === "string" || typeof flag === "number") {
+    // (1) per-child override — a spacing token (raw px tolerated)
+    return resolveSpacing(flag as SpacingToken | number) ?? DEFAULT_ROW_GAP_PX;
+  }
+  if (groupGap !== undefined) {
+    // (2) group-level override — applies to every child that doesn't set its own
+    return resolveSpacing(groupGap as SpacingToken | number) ?? DEFAULT_ROW_GAP_PX;
+  }
+  const override = GAP_OVERRIDES[node.type];
+  if (override !== undefined) return resolveSpacing(override) ?? DEFAULT_ROW_GAP_PX; // (3) per-type
+  return DEFAULT_ROW_GAP_PX; // (3) page default — inherited rhythm
+}
+
+/**
+ * Wrap one VERTICAL group_config child in the page's vertical rhythm. Margin is
+ * symmetric per side (top + bottom = `resolveGroupRowGap`), mirroring `GutterItem`,
+ * so children stack on the same rhythm as top-level page items. The FIRST child's
+ * top and the LAST child's bottom are zeroed: the group node itself sits inside a
+ * page `GutterItem` that owns the group's OUTER spacing, so a plain (cardless)
+ * group is spacing-transparent — its children space exactly as if inlined at the
+ * page level, with no doubled gutter at the group's edges. No horizontal gutter:
+ * the group owns horizontal layout (Box `direction` / `align` / `justify`).
+ */
+export function GroupGutterItem({
+  node,
+  index,
+  count,
+  groupGap,
+  children,
+}: {
+  node: Node;
+  index: number;
+  count: number;
+  groupGap?: string | number;
+  children: React.ReactNode;
+}): React.ReactElement {
+  const gap = resolveGroupRowGap(node, groupGap);
+  return (
+    <View
+      style={{
+        marginTop: index === 0 ? 0 : gap,
+        marginBottom: index === count - 1 ? 0 : gap,
+      }}
+    >
+      {children}
+    </View>
+  );
+}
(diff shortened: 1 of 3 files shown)


packages/sdui-runtime/src/layout/page-gutter.tsx
```diff
@@ -169,3 +169,68 @@ export function GutterItem({
...
+  index: number;
+  count: number;
+  groupGap?: string | number;
+  children: React.ReactNode;
+}): React.ReactElement {
+  const gap = resolveGroupRowGap(node, groupGap);
+  return (
+    <View
+      style={{
+        marginTop: index === 0 ? 0 : gap,
```
zenith-oi, line 234: **FLAG** — [defensive-edge-case] When `count === 0`, `index === count - 1` evaluates to `0 === -1` which is false, so `marginBottom` would be set to `gap` rather than `0`. This is unreachable in practice (GroupGutterItem is only rendered inside `items.map(...)` so count ≥ 1 whenever the component mounts), but the invariant is implicit. Consider asserting `count > 0` or documenting the precondition in the JSDoc so future callers of GroupGutterItem don't accidentally pass count=0 and get a non-zero bottom margin on the only child.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-scoped layout change that makes vertical group_config children inherit the page's vertical rhythm. The implementation is correct and the changeset accurately describes the visual delta. One minor concern worth flagging: the first/last-child margin-zeroing logic has an edge case when count is 0 that is harmless in practice but worth noting.**

### What this PR is doing
**Intent:** Make vertical group_config children inherit the page's per-item vertical rhythm (via a new GroupGutterItem wrapper) instead of a flat Box gap, with a three-level precedence: per-child gap → group gap → page default; horizontal groups are unchanged.
**Delivered:** The diff fully delivers the stated intent. GroupGutterItem is introduced in page-gutter.tsx, resolveGroupRowGap implements the three-level precedence exactly as described, the renderer switches vertical children to GroupGutterItem while leaving horizontal children on the Box gap path, and the changeset documents the visual change. No linked task page or brief was provided, so acceptance items cannot be cross-checked by number.

**Positive:**
- The three-level precedence (per-child → group → page default) is clearly documented in both the JSDoc and the changeset, making the mental model easy to follow.
- Zeroing the first child's top and last child's bottom to avoid doubled gutters is the right call — it keeps a plain group spacing-transparent relative to the page, exactly matching the stated design intent.
- Splitting the horizontal and vertical paths cleanly (Box gap for horizontal, GroupGutterItem for vertical) avoids any risk of the change affecting horizontal layouts.
- The changeset bump is correctly marked `minor` given the intentional visual change to existing vertical groups.

**Read against:**
- `PR body — resolution precedence table (child gap → group gap → page default)`
- `PR body — spacing-transparent plain group (first/last child margin zeroed)`
- `PR body — horizontal groups unchanged`

---

packages/sdui-runtime/src/layout/page-gutter.tsx
achin-oi: Addressed by documenting the precondition in GroupGutterItem's JSDoc (latest commit): it is only rendered from inside the group's `items.map(...)`, so `count` (= `items.length`) is ≥ 1 whenever it mounts — `count === 0` produces no render at all. The precondition is now explicit for future callers, per the suggestion.

packages/sdui-runtime/src/layout/page-gutter.tsx
```diff
@@ -169,3 +169,73 @@ export function GutterItem({
...
+        marginBottom: index === count - 1 ? 0 : gap,
```
zenith-oi, line 235: **FLAG** — [defensive-edge-case] When count === 0, `index === count - 1` evaluates to `0 === -1` (false), so marginBottom is set to `gap` instead of 0 for the lone child. The precondition comment added at L213-L215 documents the invariant but does not enforce it — callers can still pass count=0 (e.g. if items.length is evaluated before a filter). The fix is a one-liner guard: `const isLast = count === 0 || index === count - 1;` (treating an empty-count call as a single-item group). This was raised in the previous round and remains structurally unaddressed.

zenith-oi (approved):
## Zenith Code Review

**Clean, well-scoped layout change. The only unresolved issue is the previously-flagged defensive edge case where count===0 leaves the lone child's marginBottom un-zeroed — the precondition comment was added but the structural guard was not.**

### What this PR is doing
**Intent:** Inferred from diff + body: make vertical group_config children inherit the page's per-item vertical rhythm (via a new GroupGutterItem wrapper) instead of a flat Box gap, with a three-level override chain (child gap → group gap → page default), while leaving horizontal groups unchanged.
**Delivered:** The diff fully delivers the stated intent. GroupGutterItem and resolveGroupRowGap are added to page-gutter.tsx, GroupConfigRenderer is updated to use GroupGutterItem for vertical children and preserve the Box gap for horizontal children, the gap type is widened to string|number, and a changeset entry documents the minor-version bump. The spacing-transparency invariant (first child top = 0, last child bottom = 0) is implemented as described. No acceptance items from a task page are present to check against.

**Positive:**
- The three-level resolution order (child gap → group gap → page default) is clearly documented in both the JSDoc and the changeset, making the precedence rules easy to audit.
- Zeroing first-child top and last-child bottom to achieve spacing-transparency is the right design — it avoids doubled gutters at group edges without requiring callers to know about the outer GutterItem.
- Widening cfg.gap from number to string|number is a non-breaking, additive change that correctly aligns with how spacing tokens are used elsewhere in the page-gutter module.
- Extracting items into a const before the map (L70) is a small but correct improvement that avoids repeated optional-chaining and makes items.length safe to reference.

---

packages/sdui-runtime/src/layout/page-gutter.tsx
achin-oi: Already addressed in commit 517d0bd — GroupGutterItem's JSDoc now documents the `count >= 1` precondition explicitly. The case is unreachable by construction (the component only renders from inside the group's `items.map(...)`, so `count === 0` yields no render). Resolving as documented.

PR #290, opened 2026-06-21
feat(sdui-runtime): vertical group_config children inherit page rhythm

achin-oi:
Consolidation merge to main ahead of the SDUI resilience work. Group_config vertical children inherit the page's per-item rhythm (per-child / group-level / page-default gap precedence).

Note: playground demo files for this feature remain local WIP and are intentionally not in this PR.

### DIFF
--- a/packages/sdui-runtime/src/layout/page-gutter.tsx
+++ b/packages/sdui-runtime/src/layout/page-gutter.tsx
@@ -169,3 +169,68 @@ export function GutterItem({
     </View>
   );
 }
+
+/**
+ * Per-side vertical margin (px) for a child INSIDE a vertical group_config — the
+ * same page rhythm a top-level item gets, but with the group's own `gap` slotted
+ * in as the MIDDLE fallback so a group can re-space all its children at once.
+ * Resolution order:
+ *   (1) child node `gap`   — backend per-instance override, on the child
+ *   (2) group `gap`        — backend override, on the group_config node
+ *   (3) per-type override / default — the inherited page rhythm (half-gutter = 6)
+ * Applied symmetrically (top + bottom) by `GroupGutterItem`, matching the page's
+ * per-side model so two default children sit a full gutter apart (6 + 6 = 12).
+ */
+export function resolveGroupRowGap(
+  node: Node,
+  groupGap?: string | number,
+): number {
+  const flag = (node as { gap?: unknown }).gap;
+  if (typeof flag === "string" || typeof flag === "number") {
+    // (1) per-child override — a spacing token (raw px tolerated)
+    return resolveSpacing(flag as SpacingToken | number) ?? DEFAULT_ROW_GAP_PX;
+  }
+  if (groupGap !== undefined) {
+    // (2) group-level override — applies to every child that doesn't set its own
+    return resolveSpacing(groupGap as SpacingToken | number) ?? DEFAULT_ROW_GAP_PX;
+  }
+  const override = GAP_OVERRIDES[node.type];
+  if (override !== undefined) return resolveSpacing(override) ?? DEFAULT_ROW_GAP_PX; // (3) per-type
+  return DEFAULT_ROW_GAP_PX; // (3) page default — inherited rhythm
+}
+
+/**
+ * Wrap one VERTICAL group_config child in the page's vertical rhythm. Margin is
+ * symmetric per side (top + bottom = `resolveGroupRowGap`), mirroring `GutterItem`,
+ * so children stack on the same rhythm as top-level page items. The FIRST child's
+ * top and the LAST child's bottom are zeroed: the group node itself sits inside a
+ * page `GutterItem` that owns the group's OUTER spacing, so a plain (cardless)
+ * group is spacing-transparent — its children space exactly as if inlined at the
+ * page level, with no doubled gutter at the group's edges. No horizontal gutter:
+ * the group owns horizontal layout (Box `direction` / `align` / `justify`).
+ */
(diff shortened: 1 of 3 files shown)


packages/sdui-runtime/src/layout/page-gutter.tsx
```diff
@@ -169,3 +169,73 @@ export function GutterItem({
...
+ * per-side model so two default children sit a full gutter apart (6 + 6 = 12).
+ */
+export function resolveGroupRowGap(
+  node: Node,
+  groupGap?: string | number,
+): number {
+  const flag = (node as { gap?: unknown }).gap;
+  if (typeof flag === "string" || typeof flag === "number") {
+    // (1) per-child override — a spacing token (raw px tolerated)
+    return resolveSpacing(flag as SpacingToken | number) ?? DEFAULT_ROW_GAP_PX;
```
zenith-oi, line 196: **FLAG** — [no-hardcoded-business-values] `DEFAULT_ROW_GAP_PX` is used as the fallback in every branch of `resolveGroupRowGap`. The changeset prose calls it 'half-gutter = 6', which means it encodes a layout policy decision. If this constant is defined as a bare numeric literal (`const DEFAULT_ROW_GAP_PX = 6`) anywhere in `page-gutter.tsx` rather than derived from a named spacing token or a named constant that documents the 'half-gutter' policy, it should be a named, documented export (e.g. `export const DEFAULT_ROW_GAP_PX = 6; // half of the page gutter; see GutterItem`) so the policy is discoverable and not silently duplicated if the gutter value ever changes. The diff doesn't show the constant's definition, so this may already be fine — verify and add a comment if the value is bare.

zenith-oi (dismissed):
## Zenith Code Review

**Clean, well-scoped feature addition. The spacing-precedence logic is clearly documented, the edge-case (first/last child margin zeroing) is handled correctly, and the horizontal-group path is untouched. No security, data-integrity, or contract issues. One minor concern worth the engineer's attention: `DEFAULT_ROW_GAP_PX` is referenced in `page-gutter.tsx` but not visible in the diff — if it is defined as a bare numeric literal in that file rather than a named constant, it should be confirmed as already-named (the changeset says 'half-gutter = 6', so the value is meaningful and should be a named export).**

### What this PR is doing
**Intent:** Add vertical rhythm inheritance to `group_config` vertical children in the SDUI runtime, so they space like top-level page items (per-child gap → group gap → page default) instead of a fixed `sm` gap, while leaving horizontal groups unchanged.
**Delivered:** The diff fully delivers the stated intent. `resolveGroupRowGap` implements the three-level precedence chain (child node `gap` → group `gap` → per-type override / DEFAULT_ROW_GAP_PX). `GroupGutterItem` wraps each vertical child with symmetric top/bottom margins and correctly zeros the first child's top and last child's bottom to avoid doubled gutters. The renderer switches vertical groups from `Box gap` to `GroupGutterItem` wrapping while keeping horizontal groups on the existing `Box gap` path. The changeset entry accurately describes the behaviour. No linked task page or brief was provided, so acceptance items cannot be checked by number.

**Positive:**
- The three-level precedence chain (per-child → group → page default) is clearly documented in both the JSDoc and the changeset, making the design intent unambiguous for future maintainers.
- First/last child margin zeroing is handled correctly and the precondition (`count >= 1`) is explicitly documented in the JSDoc, preventing a subtle off-by-one if the component is ever called on an empty group.
- Horizontal groups are completely unaffected — the `gap={horizontal ? (cfg?.gap ?? 'sm') : undefined}` guard is clean and the existing behaviour is preserved without any conditional complexity leaking into the horizontal path.
- The `gap` type widening from `number` to `string | number` on the config type is a backwards-compatible change that correctly aligns with the spacing-token system.

---

zenith-oi (approved):
## Zenith Code Review

**Clean, well-documented layout feature. The previous-round FLAG on `no-hardcoded-business-values` for `DEFAULT_ROW_GAP_PX` remains structurally unresolved — the constant is still derived inline as `GUTTER_PX / 2` rather than imported from a shared constants module — but the pattern's own suppression heuristics apply here (this is a design-system repo, not an `amplify-platform` BC with a `_shared/constants/` module, and the value is a layout primitive, not a business-shaped literal). No new issues introduced by this push.**

### What this PR is doing
**Intent:** Add vertical rhythm inheritance to `group_config` children in the SDUI runtime: vertical groups now inherit the page's per-item gap (half-gutter) by default, with a three-level precedence (per-child `gap` → group `gap` → page default), while horizontal groups are unchanged.
**Delivered:** The diff fully delivers the stated intent. `DEFAULT_ROW_GAP_PX` is exported and documented, `resolveGroupRowGap` implements the three-level precedence chain, `GroupGutterItem` zeroes the first-child top and last-child bottom margins to avoid doubled gutters, and `GroupConfigRenderer` wires it all up — vertical groups use `GroupGutterItem`, horizontal groups keep the `Box` gap. The changeset entry accurately describes the behaviour. No linked task page or brief was provided, so acceptance items cannot be checked by number.

No issues found.

**Positive:**
- The three-level gap precedence (per-child → group → page default) is clearly documented in both the JSDoc and the changeset, making the policy discoverable without reading the implementation.
- Zeroing first-child top and last-child bottom margins to keep a plain group spacing-transparent is a subtle correctness detail that is explicitly called out in the component's JSDoc — good defensive documentation.
- Extracting `DEFAULT_ROW_GAP_PX` as a named export with a comment explaining the policy ('default row gap is half the page gutter') is exactly the right move — it replaces a magic number with a named policy constant that callers and tests can reference symbolically.

---

